Issue #1, opened 2018-10-19
Set up basic project structure


igordavydsson:
Basic django REST API

PR #4, opened 2018-10-19
Add basic project structure

igordavydsson:
Refs #1


reservations/models.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.postgres.fields import JSONField
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+
+class Reservation(models.Model):
+    created_at = models.DateTimeField(verbose_name=_('created at'), auto_now_add=True, editable=False)
```
tuomas777: `editable=False` is not needed when `auto_now_add=True`

igordavydsson: alright! thanks!

PR #8, opened 2018-10-26
Send notifications when new reservations created

igordavydsson:


notifications/tests/test_notifications.py
```diff
@@ -0,0 +1,86 @@
...
+from notifications.models import NotificationTemplate, NotificationTemplateException
+from notifications.utils import render_notification_template
+
+
+@pytest.fixture
+def notification_template(settings):
+    settings.LANGUAGES = (('fi', 'Finnish'), ('en', 'English'))
+
+    template = NotificationTemplate.objects.language('en').create(
+        type=NotificationType.SIGNUP_CREATED,
```
tuomas777: Hmm, notifications in these tests seem to have a weird type... :)

reservations/signals.py
```diff
@@ -0,0 +1,20 @@
...
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from notifications.enums import NotificationType
+from notifications.utils import send_notification
+from .models import Reservation
+
+
+@receiver(post_save, sender=Reservation)
+def reservation_notification_handler(sender, **kwargs):
```
tuomas777: IMO would be cleaner to do like `def reservation_notification_handler(sender, instance, created, **kwargs):`

notifications/tests/test_notifications.py
igordavydsson: 😄 oopsie-poopsie 😄

igordavydsson:
@tuomas777 ^ 🙂

Issue #7, opened 2018-10-25
Import ports data into our API


igordavydsson:
We have the data for all needed ports in a CSV file.
Figure out how to serve it from the API.

PR #10, opened 2018-11-08
Add commands for creating / updating Harbors

igordavydsson:
Refs #7

### DIFF
--- a/harbors/management/commands/create_harbors_from_json.py
+++ b/harbors/management/commands/create_harbors_from_json.py
@@ -1,0 +1,56 @@
+"""
+This command creates or updates Harbor objects based on a JSON
+that has the following structure for its items:
+
+"AIRORANTA": {
+    "servicemap_id": "40393",
+    "berth_count": 11,
+    "max_length": 500,
+    "max_width": 200
+}, ...
+
+"""
+import json
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from django.utils.text import slugify
+
+from harbors.models import Harbor
+
+
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument('--file', action='store', dest='file',
+                            help="Path to JSON file with harbors' data")
+
+    def handle(self, **options):
+        json_filepath = options['file']
+        if not json_filepath:
+            raise CommandError('No path to JSON file provided')
+
+        number_of_created_harbors = 0
+        number_of_modified_harbors = 0
+
+        with open(json_filepath, 'r') as json_file:
+            harbors_dict = json.load(json_file)
+            for harbor_name, harbor_data in harbors_dict.items():
+                with transaction.atomic():
+                    defaults = {
+                        "servicemap_id": harbor_data['servicemap_id'],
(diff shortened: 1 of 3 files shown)


harbors/management/commands/create_harbors_from_json.py
```diff
@@ -0,0 +1,52 @@
...
+class Command(BaseCommand):
+
+    def add_arguments(self, parser):
+        parser.add_argument('--file', action='store', dest='file',
+                            help="Path to JSON file with harbors' data")
+
+    def handle(self, **options):
+        json_filepath = options['file']
+        if not json_filepath:
+            self.stderr.write('No path to JSON file provided')
```
tuomas777: Probably better to raise [CommandError](https://docs.djangoproject.com/en/2.1/howto/custom-management-commands/#django.core.management.CommandError) here?

harbors/management/commands/create_harbors_from_json.py
```diff
@@ -0,0 +1,52 @@
...
+
+        number_of_created_harbors = 0
+        number_of_modified_harbors = 0
+
+        with open(json_filepath, 'r') as json_file:
+            harbors_dict = json.load(json_file)
+            for harbor_name, harbor_data in harbors_dict.items():
+                harbor, created = Harbor.objects.get_or_create(identifier=slugify(harbor_name))
```
tuomas777: Could use [update_or_create()](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#update-or-create) here.

harbors/management/commands/update_harbors_from_servicemap.py
```diff
@@ -0,0 +1,54 @@
...
+
+            harbor.zip_code = sm_info['address_zip']
+            harbor.www_url = sm_info['www']['fi']  # only FI for now
+            harbor.email = sm_info['email']
+            harbor.phone = sm_info['phone']
+
+            harbor.municipality = Municipality.objects.get(id=sm_info['municipality'])
+
+            latitude = sm_info['location']['coordinates'][1]
+            longitude = sm_info['location']['coordinates'][0]
```
tuomas777: Fancy people would do `longitude, latitude = sm_info['location']['coordinates']`.

harbors/management/commands/update_harbors_from_servicemap.py
```diff
@@ -0,0 +1,54 @@
...
+            harbor.location = Point(longitude, latitude, srid=settings.DEFAULT_SRID)
+
+            harbor.save()
```
tuomas777: Is there some reason why there cannot be just one `harbor.save()` at the end of the loop?

igordavydsson: Not sure, that you can set properly translated fields without saving each language... This `harbor.save()` line might be deleted though? if we save the object pretty soon when handling translated field?

tuomas777: Ok, sounds a bit stupid, but might be, I don't know Parler that well :) If that is the case, I would keep this save as well to make sure the object is saved even when there are no translations in the upcoming translated fields (probably not possible in practice but still).

PR #11, opened 2018-11-08
Add Helsinki harbors' data as fixtures

igordavydsson:
Refs #7


harbors/fixtures/helsinki-harbors.json
```diff
@@ -0,0 +1,2283 @@
...
+        "name": "Jollavene",
+        "master": 1
+    }
+},
+{
+    "model": "harbors.boattypetranslation",
+    "pk": 2,
+    "fields": {
+        "language_code": "fi",
+        "name": "Sotuvene",
```
tuomas777: I guess this should be `soutuvene` unless it actually is"social security number boat"... :D

igordavydsson: :D good catch! 👍

PR #15, opened 2018-11-12
Add helsinki harbors images

igordavydsson:
Refs #7


harbors/management/commands/add_helsinki_harbors_images.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class Command(BaseCommand):
+
+    def handle(self, **options):
+        updated_harbors = 0
+        missing_images = []
+
+        for harbor in Harbor.objects.all():
+            with transaction.atomic():
```
tuomas777: There seems to be only one db affecting place (`harbor.image_file.save(...)`), so not sure if this atomic is actually doing anything in practice.

harbors/management/commands/add_helsinki_harbors_images.py
```diff
@@ -0,0 +1,44 @@
...
+                self.stdout.write(
+                    'Assigning image for harbor with servicemap ID {}'.format(harbor.servicemap_id)
+                )
+                image_filename = '{}.jpg'.format(harbor.servicemap_id)
```
tuomas777: Should probably first check if `harbor.servicemap_id` is actually set? Or maybe better would be to loop only harbors that have `servicemap_id` set?

harbors/management/commands/add_helsinki_harbors_images.py
igordavydsson: good point! will get rid of it

harbors/management/commands/add_helsinki_harbors_images.py
igordavydsson: also good point! will fix this up

harbors/management/commands/add_helsinki_harbors_images.py
```diff
@@ -0,0 +1,42 @@
...
+from harbors.models import Harbor
+
+
+class Command(BaseCommand):
+
+    def handle(self, **options):
+        updated_harbors = 0
+        missing_images = []
+
+        for harbor in Harbor.objects.filter(servicemap_id__isnull=False):
```
tuomas777: I still have a slight problem with this, as `servicemap_id` can be both `null` and `""`, I think more robust would be to check both of those.

igordavydsson: ok, had the same thought...

igordavydsson: fixed :)

PR #18, opened 2018-12-18
Update Helsinki harbors' data fixture

igordavydsson:
New harbors added and old harbors edited per new specs.


igordavydsson:
This is just edited data. I'll just merge this right away.

PR #19, opened 2018-12-19
Add availability levels to harbors

igordavydsson:
Availability level is supposed to indicate how busy a harbor is
and how easy it is to get a place there.
Current idea: three levels (green, yellow and red) from least to
most busy. Displayed later on the frontned it would help users
to choose harbors.


harbors/models.py
```diff
@@ -87,6 +104,11 @@ class Harbor(TranslatableModel):
         BoatType, verbose_name=_('Suitable boat types'), related_name='harbors', blank=True
     )
 
+    availability_level = models.ForeignKey(
+        AvailabilityLevel, null=True, blank=True, verbose_name=_('Availability level'),
+        related_name='harbors', on_delete=models.CASCADE
```
tuomas777: Are you sure you want to delete harbors when deleting an availability level? `SET_NULL` or maybe `PROTECT` seem more viable options, of course I don't know anything about the service so I might be wrong :)

igordavydsson: oops 😅 copypaste let me down again

igordavydsson: very good catch!

igordavydsson:
@tuomas777 updated the PR. also changed the `on_delete`, form which I copied the `CASCADE` one 😬

tuomas777:
The migration doesn't seem to be updated?

igordavydsson:
@tuomas777 🙈 updated the migration too this time

tuomas777:
😄 👍

PR #25, opened 2019-01-10
Require authentication to POST reservations

igordavydsson:
`CORS_ORIGIN_WHITELIST` does not ban requests from non-front urls.
This will ensure that only UI will POST reservations.


igordavydsson:
OK, not needed.

PR #28, opened 2019-01-14
Add creation timestamp to reservations CSV

igordavydsson:
Add column with reservations' creation timestamps
to CSV export file.


igordavydsson:
This is for the folks over at boat management department (? or whatever it's called). They want it to be as human readable as possible.

tuomas777:
ok! 👍

PR #33, opened 2019-01-15
Add an option to notify admins about new reservations

igordavydsson:
If any admin is chosen to receive a notification and if admin
notifications have body and subject configured, django will
send email to admins about new reservations.


notifications/admin.py
```diff
@@ -17,9 +20,35 @@ def __init__(self, *args, **kwargs):
         choices = [x for x in self.fields['type'].choices if x[0] not in existing_types]
         self.fields['type'].choices = choices
 
+        admins_qs = get_user_model().objects.filter(email__isnull=False).\
```
tuomas777: I don't think `User.email` can be `null`, so this should be `.exclude(email='')`.

notifications/utils.py
```diff
@@ -24,8 +24,10 @@ def send_notification(email, notification_type, context=None, language=DEFAULT_L
     if context is None:
         context = {}
 
+    template = NotificationTemplate.objects.get(type=notification_type)
```
tuomas777: This will throw if there no `NotificationTemplate` object at all, which wasn't the case before, and IMO is not optimal.

igordavydsson: ah, right! missed that when refactoring

igordavydsson:
@tuomas777 thanks for the comments! updated 🙂

PR #37, opened 2019-01-17
Add main fields to reservation model

igordavydsson:
All the data about reservations should be now stored in separate
fields, instead of being pushed into 'data' field as simple json.


igordavydsson:
Thanks for the tip @tuomas777 ! 🙂 
Got rid of those `null=True` on the `CharField`s. Will add other commits to this PR.

reservations/migrations/0003_from_data_to_main_fields.py
```diff
@@ -0,0 +1,127 @@
...
+from decimal import Decimal, InvalidOperation
+from functools import reduce
+
+from django.db import IntegrityError, migrations, transaction
+
+CHAR_FIELDS_TO_DATA_KEY_MAP = {
+    'phone_number': "applicant.contact.mobile_phone",
+    'address': "applicant.postal.street_address",
+    'zip_code': "applicant.postal.postal_code",
+    'municipality': "applicant.postal.munacipality",
```
tuomas777: There seems to be a typo in "applicant.postal.mun**A**cipality".

reservations/migrations/0003_from_data_to_main_fields.py
```diff
@@ -0,0 +1,127 @@
...
+                    value = deep_get(r.data, data_address)
+                    if value:
+                        setattr(r, attr_name, value)
+
+                boat_weight = deep_get(r.data, "boat.weight")
+                if boat_weight:
+                    try:
+                        r.boat_weight = int(boat_weight)
+                    except (TypeError, ValueError):
+                        pass
```
tuomas777: Are you sure it is safe to just silently ignore all these conversion errors, not even some warning logging needed?

reservations/api.py
```diff
@@ -1,13 +1,45 @@
...
+
     class Meta:
         model = Reservation
         fields = '__all__'
 
+    def create(self, validated_data):
+        # Django REST requires to explicitly write creation logic
+        # for ManyToMany relations with a Through model.
+        chosen_harbors = validated_data.pop('harborchoice_set')
+        reservation = super(ReservationSerializer, self).create(validated_data)
```
tuomas777: Just `super().create(...)` is enough in Python 3.

reservations/migrations/0003_from_data_to_main_fields.py
igordavydsson: yeah, frontend had a typo there, so all the reservations now have this typo too 🙂

reservations/api.py
igordavydsson: ah, right! forgot that those could be ommited. 👍

reservations/migrations/0003_from_data_to_main_fields.py
igordavydsson: good point! best to add at least logger 👍

PR #96, opened 2019-10-08
VEN-145 | Fix notifications

igordavydsson:
Fix signal for sending reservation notifications.

Refs VEN-145

### DIFF
--- a/reservations/notifications.py
+++ b/reservations/notifications.py
@@ -1,6 +1,12 @@
+from unittest.mock import MagicMock
+
+from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
+from django_ilmoitin.dummy_context import dummy_context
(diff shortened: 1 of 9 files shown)


reservations/notifications.py
```diff
@@ -20,3 +26,11 @@ class Labels:
     NotificationType.WINTER_STORAGE_RESERVATION_CREATED.value,
     NotificationType.WINTER_STORAGE_RESERVATION_CREATED.label,
 )
+
+dummy_context.context.update(
+    {
+        "created_at": localize_datetime(timezone.now()),
+        # TODO: pass more appropriate in-memory objects
+        "reservation": Mock(),
```
igordavydsson: especially this part... how does this look?

igordavydsson: discussed in Slack: seems like a viable solution for the time being

PR #103, opened 2019-11-01
VEN-138 | Add initial data models for customer profiles

igordavydsson:
Berth reservations backend is going to store some of the data
about the customers in CustomerProfile, which is an extension
to the Profile model in the open-city-profile backend. They
should share the same UUID for proper matching. Logic for this
will come later.

`Users` app will ideally contain info that has to do with auth
logic and Tunnistamo connection. It might get a heavy refactor,
if / when Tunnistamo will get deprecated / refactored.

`Customers` app should be a layer on top of the `User` and store
more persistent info, not dependant that much on Tunnistamo.


igordavydsson:
OK, thanks @tuomas777 ! This will be a pilot project with the federation stuff, as the fastest way for us to see how production ready it is is to go with it for now, I think 😅 Maybe later on, we'll switch it for something else, or maybe it will turn out not that bad after all.

Infra team would like to keep node.js code in a separate repo, so I will move everything to [berth-federation-gateway](https://github.com/City-of-Helsinki/berth-federation-gateway).

Will merge this PR now.

PR #108, opened 2019-12-23
Expose Berth mooring type enum members as a query

igordavydsson:
We need them to be available through a regular GQL query and not
just an introspection query ("__type" query). This way when the
forntend fetches them from the apollo gateway, the request will
come through the gateway to our GQL API.

Refs VEN-364

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -25,6 +25,23 @@ BerthMooringTypeEnum = graphene.Enum.from_enum(
 )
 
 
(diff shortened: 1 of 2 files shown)


igordavydsson:
OK, after some discussions and some delibiration, the solution presented in this PR is deemed "hacky" and the PR will be closed. 🙂 The translations will be implemented directly in the frontend's codebase.

PR #110, opened 2020-01-08
VEN-344 | Add free text field to berth model

a-rmz:
Add a free text field to the Berth model for comments.


resources/models.py
```diff
@@ -406,6 +406,7 @@ class Berth(AbstractBoatPlace):
         related_name="berths",
         on_delete=models.PROTECT,
     )
+    comments = models.TextField(verbose_name=_("comments"), blank=True,)
```
igordavydsson: the name should be `comment`, as field names in plural are used for reverse RoeignKeys and M2M relations 🙂

igordavydsson:
please, avoid back-n-forth changes between commits. you could change the original migration file as well as `models.py` and then just amend your original commit. 
amending commits is the preferred way of reacting to PR feedback, unless you have a huge and complicated PR, where you have merged `develop` into you feature branch a few times.

a-rmz:
> please, avoid back-n-forth changes between commits. you could change the original migration file as well as `models.py` and then just amend your original commit and `push -f`.
> amending commits is the preferred way of reacting to PR feedback, unless you have a huge and complicated PR, where you have merged `develop` into you feature branch a few times.

Okay! Good to know, wasn't sure if it was better to amend and `push -f` or add more commits, but will do from now.

a-rmz:
Wohoo! 😄

PR #111, opened 2020-01-10
Update data model for boats

igordavydsson:
In order to unify the measurements across different models,
we will ensure that width/length/draught are measured in meters
and stored as decimals, while weight is measured in kilos and 
stored as integers.
Changes to other data models will come later in separate PRs.


customers/models.py
```diff
@@ -50,22 +53,27 @@ class Boat(TimeStampedModel, UUIDModel):
 
     # Dimensions
     length = models.DecimalField(
-        verbose_name=_("length"), decimal_places=2, max_digits=5
+        verbose_name=_("length (m)"),
+        decimal_places=2,
+        max_digits=5,
+        validators=[MinValueValidator(Decimal("0.01"))],
```
charn: That is a very tiny ship. :joy:

customers/migrations/0004_change_boat_weight_to_int.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ("customers", "0003_add_validators_to_boat_dimensions"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="boat",
+            name="weight",
+            field=models.PositiveIntegerField(
```
charn: Will this automatically get rid possible decimals i.e. is it safe to do without a data migration or do we even need to care about the data at this point?

customers/models.py
igordavydsson: for paper boats 🙃 ⛵️

customers/migrations/0004_change_boat_weight_to_int.py
igordavydsson: we do not have any data at this point, so I just skipped data migration.

PR #112, opened 2020-01-10
VEN-134 | Add initial data models for leases

igordavydsson:
Apart rom the leases itself, we want to store the status changes
in a sensible format, hence the separate data models for them.


leases/models.py
```diff
@@ -0,0 +1,126 @@
...
+        blank=True,
+        null=True,
+        related_name="leases",
+    )
+
+    def __str__(self):
+        return " {} > {} - {} ({})".format(
+            self.place, self.start_date, self.end_date, self.status
+        )
+
```
tuomas777: I think you could use [`default_related_name`](https://docs.djangoproject.com/en/3.0/ref/models/options/#default-related-name) for this model and `BerthLease` if you want to have proper related names for customer and boat.

leases/tests/test_lease_models.py
```diff
@@ -0,0 +1,56 @@
...
+    BerthLease,
+    BerthLeaseChange,
+    WinterStorageLease,
+    WinterStorageLeaseChange,
+)
+from .factories import BerthLeaseFactory
+
+
+def test_berth_lease_model(berth_lease):
+    assert BerthLease.objects.all().count() == 1
```
tuomas777: Unnecessary `.all()`.

leases/tests/test_lease_models.py
```diff
@@ -0,0 +1,56 @@
...
+    assert BerthLeaseChange.objects.count() == 0
+
+    berth_lease.end_date = berth_lease.end_date + timedelta(days=1)
+    berth_lease.save()
+
+    assert BerthLeaseChange.objects.count() == 0
+
+
+def test_winter_storage_lease_model(winter_storage_lease):
+    assert WinterStorageLease.objects.all().count() == 1
```
tuomas777: Unnecessary `.all()`.

leases/tests/test_lease_models.py
```diff
@@ -0,0 +1,56 @@
...
+def test_winter_storage_lease_status_changes_are_tracked(winter_storage_lease):
+    assert WinterStorageLeaseChange.objects.count() == 0
+
+    winter_storage_lease.status = LeaseStatus.EXPIRED
+    winter_storage_lease.save()
+
+    assert WinterStorageLeaseChange.objects.count() == 1
+
+
+def test_lease_should_have_boat_owner_as_customer(berth_lease):
```
igordavydsson: note to self: no need for the fixture in this test

leases/models.py
igordavydsson: oh, good idea, looks pretty neat! thanks 🙂

leases/tests/test_lease_models.py
igordavydsson: good catch 👀

leases/tests/test_lease_models.py
igordavydsson: good catch 👀

PR #113, opened 2020-01-13
VEN-143 | Berth mutations

a-rmz:
## Description ✨ 
* Add create, update, and delete mutations for `Berth`s.
* Add a test suite for the mutations
* Override `NOTIFICATIONS_ENABLED` on the test settings to avoid tests failing if the `.env` is missing.

## Issues 🐛 
Partially closes [VEN-143](https://helsinkisolutionoffice.atlassian.net/browse/VEN-143)

### DIFF
--- a/berth_reservations/schema.py
+++ b/berth_reservations/schema.py
@@ -30,9 +30,13 @@ class Query(customers.schema.Query, resources.schema.Query, graphene.ObjectType)
     pass
 
 
+class Mutation(resources.schema.Mutation, graphene.ObjectType):
+    pass
+
+
 # We need to list all the extended types separately,
 # otherwise graphene will not generate their schemas.
 extended_types = [customers.schema.ProfileNode]
 
 
-new_schema = build_schema(Query, types=extended_types)
+new_schema = build_schema(Query, Mutation, types=extended_types)
--- a/berth_reservations/tests/conftest.py
+++ b/berth_reservations/tests/conftest.py
@@ -1,5 +1,7 @@
 import pytest
 
+from .factories import UserFactory
+
 
 @pytest.fixture(autouse=True)
 def autouse_django_db(db):
@@ -10,3 +12,26 @@ def autouse_django_db(db):
 def force_settings(settings):
     settings.MAILER_EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"
     settings.DEFAULT_FROM_EMAIL = "noreply@foo.bar"
+
+
+@pytest.fixture
+def user(request):
+    permissions = request.param if hasattr(request, "param") else None
+
+    if permissions == "superuser":
+        user = UserFactory(is_superuser=True)
+    elif permissions == "staff":
+        user = UserFactory(is_staff=True)
+    elif permissions == "none":
+        user = None
+    # When the fixture is called without parameters, return the base user
+    elif permissions == "base" or permissions is None:
+        user = UserFactory()
+
+    return user
+
+
+@pytest.fixture
+def superuser():
+    user = UserFactory(is_superuser=True)
+    return user
--- a/berth_reservations/tests/utils.py
+++ b/berth_reservations/tests/utils.py
@@ -69,6 +69,10 @@ def assert_not_enough_permissions(executed):
     )
 
 
+def assert_doesnt_exist(model, executed):
+    assert executed["errors"][0]["message"] == f"{model} matching query does not exist."
+
+
 def assert_field_missing(field_name, executed):
     errors = str(executed["errors"])
     assert field_name in errors
--- a/berth_reservations/tests/utils.py
+++ b/berth_reservations/tests/utils.py
@@ -10,7 +10,13 @@ class GraphQLTestClient(Client):
     """
 
     def execute(
-        self, query="", op_name=None, variables=None, graphql_url="/graphql/", lang="en"
+        self,
+        query="",
+        op_name=None,
+        variables=None,
+        graphql_url="/graphql/",
+        lang="en",
+        user=None,
     ):
         """
         Function that posts the passed query to our project's GraphQL endpoint.
@@ -26,6 +32,10 @@ class GraphQLTestClient(Client):
         :type variables: None | dict
         :param lang: Language to be set in the "Accept-Language" header.
         :type lang: str
+        :param user: In case the query requires authentication, a "user" with the
+                     desired permissions can be passed. It will only attempt to
+                     login if a user is passed.
+        :type user: users.models.User
 
         :return: response from graphql endpoint.  The response has the "data" key.
                  It will have the "error" key if any error happened.
@@ -38,11 +48,28 @@ class GraphQLTestClient(Client):
         if variables:
             body["variables"] = {"input": variables}
 
+        if user:
+            self.force_login(user)
+
         resp = self.post(
             graphql_url,
             json.dumps(body),
             content_type="application/json",
             HTTP_ACCEPT_LANGUAGE=lang,
         )
+
         response_json = json.loads(resp.content.decode())
         return response_json
+
+
+def assert_not_enough_permissions(executed):
+    assert (
+        executed["errors"][0]["message"]
+        == "You do not have permission to perform this action"
+    )
+
+
+def assert_field_missing(field_name, executed):
+    errors = str(executed["errors"])
+    assert field_name in errors
+    assert "found null" in errors
--- a/resources/tests/conftest.py
+++ b/resources/tests/conftest.py
@@ -8,7 +8,6 @@ from .factories import (
     BoatTypeFactory,
     HarborFactory,
     PierFactory,
-    UserFactory,
     WinterStorageAreaFactory,
     WinterStoragePlaceFactory,
     WinterStorageSectionFactory,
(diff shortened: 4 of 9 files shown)


berth_reservations/tests/conftest.py
```diff
@@ -10,3 +10,4 @@ def autouse_django_db(db):
 def force_settings(settings):
     settings.MAILER_EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"
     settings.DEFAULT_FROM_EMAIL = "noreply@foo.bar"
+    settings.NOTIFICATIONS_ENABLED = 1
```
igordavydsson: this is Python syntax now, instead of the dotenv syntax, so `True` wouldbe more appropriate 🙂

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,284 @@
...
+    assert_not_enough_permissions(executed)
+
+
+def test_create_berth_no_number(pier, berth_type, staff_user):
+    variables = {
+        "pierId": to_global_id("PierNode", str(pier.id)),
+        "berthTypeId": to_global_id("BerthTypeNode", str(berth_type.id)),
+    }
+
+    assert len(Berth.objects.all()) == 0
```
charn: Just FYI you can also do this with `Berth.objects.count()` or `Berth.objects.all().count()`

resources/tests/factories.py
```diff
@@ -110,3 +113,15 @@ class WinterStoragePlaceFactory(AbstractPlaceFactory):
...
         model = WinterStoragePlace
+
+
+class UserFactory(factory.django.DjangoModelFactory):
+    first_name = factory.Faker("first_name")
+    last_name = factory.Faker("last_name")
+    username = factory.Faker("user_name")
+    email = factory.Faker("email")
+    is_staff = False
+    is_superuser = False
```
igordavydsson: these are `False` by default, so no need to specify that 🙂

resources/tests/conftest.py
```diff
@@ -53,3 +61,21 @@ def winter_storage_section():
...
+
+@pytest.fixture
+def user():
+    user = UserFactory()
+    return user
+
+
+@pytest.fixture
+def staff_user():
+    user = UserFactory.build(is_staff=True)
```
igordavydsson: you don't have to use the `build()` method for testing, `UserFactory(is_staff=True)` is enough

resources/tests/conftest.py
```diff
@@ -53,3 +61,21 @@ def winter_storage_section():
...
+    return user
+
+
+@pytest.fixture
+def superuser():
+    user = UserFactory.build(is_superuser=True)
```
igordavydsson: same as above 🙂

berth_reservations/tests/utils.py
```diff
@@ -38,11 +44,16 @@ def execute(
         if variables:
             body["variables"] = {"input": variables}
 
+        if user:
+            user.save()
```
igordavydsson: is saving really necessary? 🙂

a-rmz:
I had to undo the changes related to the `NOTIFICATIONS_ENABLED` because it's used on start time. So once the server is already up, it won't change anything if the value changes. That's why the value has to be set on the env beforehand.

[berth-reservations/blob/develop/reservations/signals.py#L27](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/reservations/signals.py#L27)
```python
if settings.NOTIFICATIONS_ENABLED:
    reservation_saved.connect(
        reservation_notification_handler, dispatch_uid="reservation_saved"
    )
```

igordavydsson:
@a-rmz ah, true 🤔 and I think I tried adding that settings override there and faced the same issue before, but successfully forgot about it 😅 
well, let's use the env then

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,284 @@
...
+    mutation DeleteBerth($input: DeleteBerthMutationInput!) {
+      deleteBerth(input: $input) {
+        __typename
+      }
+    }
+"""
+
+
+def test_delete_berth_superuser(superuser, berth):
+    berth.save()
```
igordavydsson: do you need to save it? 🤔

berth_reservations/tests/utils.py
```diff
@@ -38,11 +48,33 @@ def execute(
...
+
+
+def assert_doesnt_exist(model, executed):
+    assert executed["errors"][0]["message"] == f"{model} matching query does not exist."
+
+
+def assert_field_missing(field_name, executed):
+    errors = str(executed["errors"])
+    assert field_name in errors
+    assert "found null" in errors
```
igordavydsson: pretty neat helper functions 👍 GJ

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,284 @@
...
+        variables=variables,
+        graphql_url=GRAPHQL_URL,
+        user=superuser,
+    )
+
+    assert len(Berth.objects.all()) == 0
+
+
+def test_delete_berth_staff_user(staff_user, berth):
+    berth.save()
```
igordavydsson: do you need to save it? 🤔

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,284 @@
...
+        graphql_url=GRAPHQL_URL,
+        user=staff_user,
+    )
+
+    assert len(Berth.objects.all()) == 1
+    assert_not_enough_permissions(executed)
+
+
+def test_delete_berth_no_user(berth):
+    berth.save()
```
igordavydsson: do you need to save it? 🤔

berth_reservations/schema.py
```diff
@@ -30,9 +30,14 @@ class Query(customers.schema.Query, resources.schema.Query, graphene.ObjectType)
     pass
 
 
+class Mutation(resources.schema.Mutation, graphene.ObjectType):
+    pass
+
+
 # We need to list all the extended types separately,
+
```
igordavydsson: 👀

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,262 @@
...
+        "berthTypeId": to_global_id("BerthTypeNode", str(berth_type.id)),
+    }
+
+    assert Berth.objects.count() == 0
+
+    executed = client.execute(
+        query=CREATE_BERTH_MUTATION,
+        variables=variables,
+        graphql_url=GRAPHQL_URL,
+        user=None,
```
igordavydsson: you probably would want to pass `staff_user` here 🤔

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,262 @@
...
+    )
+
+    assert Berth.objects.count() == 0
+    assert_not_enough_permissions(executed)
+
+
+def test_create_berth_no_number(pier, berth_type, staff_user):
```
igordavydsson: to see that the problem is actually in the missing number field, you would need to have otherwise working situation, i.e. superuser submits the mutation

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,262 @@
...
+    assert Berth.objects.count() == 1
+
+    executed = client.execute(
+        query=UPDATE_BERTH_MUTATION,
+        variables=variables,
+        graphql_url="/graphql_v2/",
+        user=superuser,
+    )
+
+    assert Berth.objects.count() == 1
```
igordavydsson: probably no need to assert this, as the count should not have changed anyway

resources/tests/test_resources_mutations.py
```diff
@@ -0,0 +1,262 @@
...
+    assert executed["data"]["updateBerth"]["berth"]["berthType"]["id"] == berth_type_id
+
+
+def test_update_berth_no_id(berth, pier, berth_type, superuser):
+    pier_id = to_global_id("PierNode", str(pier.id))
+    berth_type_id = to_global_id("BerthTypeNode", str(berth_type.id))
+
+    variables = {
+        "number": "666",
+        "comment": "foobar",
```
igordavydsson: also you could skip this "comment" field in most test cases (maybe keep it in one or two to check that it exists at all), as it is not a required field for the model, would have made your life 0.1% easier 😅  no need to delete now, of course, since you already have written all these lines.

resources/tests/test_resources_mutations.py
a-rmz: That's exactly what I was testing.

igordavydsson: hm, yeah, ok, let's keep it 👍

resources/tests/conftest.py
```diff
@@ -64,14 +64,17 @@ def winter_storage_place():
 
 
 @pytest.fixture
-def user():
-    user = UserFactory()
-    return user
-
+def user(request):
+    permissions = request.param
```
igordavydsson: what does pytest pass here, when the fixture is called in a regular way, i.e. not from a parametrize call? I mean, what is the default for this `request.param`? do we need to have a default `else` for when it is `None`? 🤔

a-rmz: When you just call the fixture, `request.param` is not defined.

I updated the fixture to check if it has the attribute, otherwise, sets it to none and returns.

resources/tests/conftest.py
```diff
@@ -64,14 +64,18 @@ def winter_storage_place():
 
 
 @pytest.fixture
-def user():
-    user = UserFactory()
-    return user
-
+def user(request):
```
igordavydsson: could you please move this fixture to `berth_reservations/tests/conftest.py` as it is probably gonna be used by multiple apps within the project 🙂 this is also why I was asking about default `request.param`, since other tests might use it directly

a-rmz: Sure! I can also default no param to the base user (I think it would make more sense)

PR #115, opened 2020-01-16
VEN-143 | Harbor mutations

a-rmz:
## Description :sparkles:
* Add mutation and tests for the `Harbor` resource.

## Issues :bug:
Partially closes VEN-143

## Testing :alembic:
To manually test the mutations, you can run the following queries on GraphiQL with the inputs provided. Just add valid IDs and you're ready to test. :rocket:

```graphql
query FetchHarbor($fetchId: ID!) {
  harbor(id: $fetchId) {
    id
    type
    geometry {
      type
      coordinates
    }
    bbox
    properties {
      name
      servicemapId
      streetAddress
      zipCode
      phone
      email
      wwwUrl
      imageLink
      municipality
      imageFile
      availabilityLevel {
        title
      }
      numberOfPlaces
      maximumWidth
      maximumLength
      maximumDepth
    }
  }
}

mutation CreateHarbor($createHarbor:CreateHarborMutationInput!) {
  createHarbor(input:$createHarbor) {
    harbor {
      id
      type
      geometry {
        type
        coordinates
      }
      bbox
      properties {
        name
        servicemapId
        streetAddress
        zipCode
        phone
        email
        wwwUrl
        imageLink
        municipality
        imageFile
        availabilityLevel {
          id
          title
        }
        numberOfPlaces
        maximumWidth
        maximumLength
        maximumDepth
      }
    }
  }
}

mutation DeleteHarbor($deleteId: DeleteHarborMutationInput!) {
	deleteHarbor(input: $deleteId) {
    __typename
  }  
}

mutation UpdateHarbor($updateHarbor: UpdateHarborMutationInput!) {
  updateHarbor(input: $updateHarbor) {
    harbor {
      id
      type
      geometry {
        type
        coordinates
      }
      bbox
      properties {
        name
        servicemapId
        streetAddress
        zipCode
        phone
        email
        wwwUrl
        imageLink
        municipality
        imageFile
        availabilityLevel {
          id
          title
        }
        numberOfPlaces
        maximumWidth
        maximumLength
        maximumDepth
      }
    }
  }
}
```

```json
{
  "fetchId": "",
  "deleteId": {
    "id": ""
  },
  "createHarbor": {
    "municipality": "helsinki",
    "imageFile": "https://foobar.com/image.png",
    "availabilityLevel": "1",
    "numberOfPlaces": 150,
    "maximumWidth": 350,
    "maximumLength": 400,
    "maximumDepth": 100,
    "name": "Foobarsatama",
    "streetAddress": "Foobarsatamantie 1234",
    "servicemapId": "1",
    "location": {
      "type": "Point",
      "coordinates": [125.6, 10.1]
    },
    "zipCode": "00119",
    "phone": "123123123",
    "email": "foo@bar.com",
    "wwwUrl": "https://foobar.com",
    "imageLink": "https://foobar.com/image.png"
  },
  "updateHarbor": {
    "id": "",
    "municipality": "espoo",
    "zipCode": "25251"
  }
}
```

### DIFF
--- a/berth_reservations/tests/utils.py
+++ b/berth_reservations/tests/utils.py
@@ -70,7 +70,14 @@ def assert_not_enough_permissions(executed):
 
 
 def assert_doesnt_exist(model, executed):
-    assert executed["errors"][0]["message"] == f"{model} matching query does not exist."
+    errors = str(executed["errors"])
+    assert f"{model} matching query does not exist" in errors
+
+
+def assert_field_duplicated(field_name, executed):
+    errors = str(executed["errors"])
+    assert field_name in errors
+    assert "already exists" in errors
 
 
 def assert_field_missing(field_name, executed):
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -2,12 +2,14 @@ import graphene
 import graphql_geojson
 from django.db import transaction
 from django.db.models import Prefetch
+from django.utils.translation import get_language
 from graphene import relay
 from graphene_django.fields import DjangoConnectionField, DjangoListField
 from graphene_django.filter import DjangoFilterConnectionField
 from graphene_django.types import DjangoObjectType
 from graphql_jwt.decorators import login_required, superuser_required
 from graphql_relay import from_global_id
+from munigeo.models import Municipality
 
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 
@@ -170,6 +172,16 @@ class WinterStorageAreaNode(graphql_geojson.GeoJSONType):
             return None
 
 
+class AbstractAreaInput:
+    servicemap_id = graphene.String()
+    zip_code = graphene.String()
+    phone = graphene.String()
+    email = graphene.String()
+    www_url = graphene.String()
+    location = graphql_geojson.Geometry()
+    image_link = graphene.String()
+
+
 class CreateBerthMutation(graphene.ClientIDMutation):
     class Input:
         number = graphene.String(required=True)
@@ -328,6 +340,54 @@ class DeleteBerthTypeMutation(graphene.ClientIDMutation):
         return DeleteBerthTypeMutation()
 
 
+class HarborInput(AbstractAreaInput):
+    municipality_id = graphene.String()
+    image_file = graphene.String()
+    availability_level_id = graphene.ID()
+    number_of_places = graphene.Int()
+    maximum_width = graphene.Int()
+    maximum_length = graphene.Int()
+    maximum_depth = graphene.Int()
+    name = graphene.String()
+    street_address = graphene.String()
+
+
+class CreateHarborMutation(graphene.ClientIDMutation):
+    class Input(HarborInput):
+        pass
+
+    harbor = graphene.Field(HarborNode)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        lang = get_language()
+
+        availability_level_id = kwargs.pop("availability_level_id", None)
+        if availability_level_id:
+            try:
+                kwargs["availability_level"] = AvailabilityLevel.objects.get(
+                    pk=availability_level_id
+                )
+            except AvailabilityLevel.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        municipality_id = kwargs.pop("municipality_id", None)
+        if municipality_id:
+            try:
+                kwargs["municipality"] = Municipality.objects.get(id=municipality_id)
+            except Municipality.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        harbor = Harbor.objects.language(lang).create(**kwargs)
+
+        return CreateHarborMutation(harbor=harbor)
+
+
 class Query:
     availability_levels = DjangoListField(AvailabilityLevelType)
     boat_types = DjangoListField(BoatTypeType)
@@ -452,3 +512,6 @@ class Mutation:
     create_berth_type = CreateBerthTypeMutation.Field()
     delete_berth_type = DeleteBerthTypeMutation.Field()
     update_berth_type = UpdateBerthTypeMutation.Field()
+
+    # Harbors
+    create_harbor = CreateHarborMutation.Field()
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -388,6 +388,47 @@ class CreateHarborMutation(graphene.ClientIDMutation):
         return CreateHarborMutation(harbor=harbor)
 
 
+class UpdateHarborMutation(graphene.ClientIDMutation):
+    class Input(HarborInput):
+        id = graphene.ID(required=True)
+
+    harbor = graphene.Field(HarborNode)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        id = from_global_id(kwargs.pop("id"))[1]
+
+        lang = get_language()
+
+        try:
+            harbor = Harbor.objects.language(lang).get(pk=id)
+
+            availability_level_id = kwargs.pop("availability_level_id", None)
+            if availability_level_id:
+                kwargs["availability_level"] = AvailabilityLevel.objects.language(
+                    lang
+                ).get(pk=availability_level_id)
+
+            municipality_id = kwargs.pop("municipality_id", None)
+            if municipality_id:
+                kwargs["municipality"] = Municipality.objects.get(id=municipality_id)
+        except (
+            Harbor.DoesNotExist,
+            AvailabilityLevel.DoesNotExist,
+            Municipality.DoesNotExist,
+        ) as e:
+            raise VenepaikkaGraphQLError(e)
+
+        update_object(harbor, kwargs)
+
+        return UpdateHarborMutation(harbor=harbor)
+
+
 class DeleteHarborMutation(graphene.ClientIDMutation):
     class Input:
         id = graphene.ID(required=True)
@@ -539,3 +580,4 @@ class Mutation:
     # Harbors
     create_harbor = CreateHarborMutation.Field()
     delete_harbor = DeleteHarborMutation.Field()
+    update_harbor = UpdateHarborMutation.Field()
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -388,6 +388,29 @@ class CreateHarborMutation(graphene.ClientIDMutation):
         return CreateHarborMutation(harbor=harbor)
 
 
+class DeleteHarborMutation(graphene.ClientIDMutation):
+    class Input:
+        id = graphene.ID(required=True)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        id = from_global_id(kwargs.get("id"))[1]
+
+        try:
+            harbor = Harbor.objects.get(pk=id)
+        except Harbor.DoesNotExist as e:
+            raise VenepaikkaGraphQLError(e)
+
+        harbor.delete()
+
+        return DeleteHarborMutation()
+
+
 class Query:
     availability_levels = DjangoListField(AvailabilityLevelType)
     boat_types = DjangoListField(BoatTypeType)
@@ -515,3 +538,4 @@ class Mutation:
 
     # Harbors
     create_harbor = CreateHarborMutation.Field()
+    delete_harbor = DeleteHarborMutation.Field()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,18 @@
+default_language_version:
+    python: python3    
+repos:
+  - repo: https://github.com/psf/black
+    rev: 19.10b0
+    hooks:
+      - id: black
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.7.9
+    hooks:
+      - id: flake8
+        additional_dependencies: [flake8-bugbear]
+        exclude: "migrations"
+  - repo: https://github.com/timothycrosley/isort
+    rev: 4.3.21
+    hooks:
+      - id: isort
+        exclude: "migrations"
--- a/berth_reservations/tests/conftest.py
+++ b/berth_reservations/tests/conftest.py
@@ -1,6 +1,6 @@
 import pytest
 
-from .factories import UserFactory
+from .factories import MunicipalityFactory, UserFactory
 
 
 @pytest.fixture(autouse=True)
@@ -12,6 +12,7 @@ def autouse_django_db(db):
 def force_settings(settings):
     settings.MAILER_EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"
     settings.DEFAULT_FROM_EMAIL = "noreply@foo.bar"
+    settings.LANGUAGE_CODE = "en"
 
 
 @pytest.fixture
@@ -35,3 +36,9 @@ def user(request):
 def superuser():
     user = UserFactory(is_superuser=True)
     return user
+
+
+@pytest.fixture
+def municipality():
+    municipality = MunicipalityFactory()
+    return municipality
--- a/berth_reservations/tests/factories.py
+++ b/berth_reservations/tests/factories.py
@@ -1,5 +1,6 @@
 import factory
 from django.contrib.auth import get_user_model
+from munigeo.models import Municipality
 
 
 class UserFactory(factory.django.DjangoModelFactory):
@@ -10,3 +11,11 @@ class UserFactory(factory.django.DjangoModelFactory):
 
     class Meta:
         model = get_user_model()
+
+
+class MunicipalityFactory(factory.django.DjangoModelFactory):
+    id = factory.Faker("word")
+    name = factory.Faker("city")
+
+    class Meta:
+        model = Municipality
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -4,6 +4,7 @@ flake8
 freezegun
 ipython
 isort
+pre-commit
 pytest
 pytest-cov
 pytest-django
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,30 +5,37 @@
 #    pip-compile requirements-dev.in
 #
 appdirs==1.4.3            # via black
+appnope==0.1.0            # via ipython
+aspy.yaml==1.3.0          # via pre-commit
 atomicwrites==1.3.0       # via pytest
 attrs==19.3.0             # via black, pytest
 autopep8==1.4.4
 backcall==0.1.0           # via ipython
 black==19.10b0
+cfgv==2.0.1               # via pre-commit
 click==7.0                # via black
 coverage==4.5.4           # via pytest-cov
 decorator==4.4.1          # via ipython, traitlets
 entrypoints==0.3          # via flake8
 flake8==3.7.9
 freezegun==0.3.12
-importlib-metadata==0.23  # via pluggy, pytest
+identify==1.4.10          # via pre-commit
+importlib-metadata==0.23  # via pluggy, pre-commit, pytest
+importlib-resources==1.0.2  # via pre-commit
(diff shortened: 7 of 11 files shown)


berth_reservations/tests/utils.py
```diff
@@ -84,3 +84,8 @@ def assert_invalid_enum(field_name, expected_value, executed):
     assert expected_value in errors
     assert field_name in errors
     assert f'Expected type "{expected_value}"' in errors
+
+
+def assert_model_doesnt_exist(model, executed):
```
igordavydsson: you already have the same util a few lines above 🤔

a-rmz: 🤦‍♀ 
TIL: naming is important.

resources/schema.py
```diff
@@ -170,6 +171,28 @@ def resolve_image_file(self, info, **kwargs):
...
+    servicemap_id = graphene.String()
+    zip_code = graphene.String()
+    phone = graphene.String()
+    email = graphene.String()
+    www_url = graphene.String()
+    location = graphql_geojson.Geometry()
+    image_link = graphene.String()
+
+    class Meta:
+        abstract = True
```
igordavydsson: I don't think you need tp specify that. Graphene does not check for this.

resources/schema.py
```diff
@@ -328,6 +351,138 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+class CreateHarborMutation(graphene.ClientIDMutation):
+    class Input:
+        area = HarborAreaInput(required=True)
+        municipality = graphene.String()
+        image_file = graphene.String()
+        availability_level = graphene.ID()
+        number_of_places = graphene.Int()
+        maximum_width = graphene.Int()
+        maximum_length = graphene.Int()
+        maximum_depth = graphene.Int()
```
igordavydsson: you can skip these, as they will be calculated automatically later, ref VEN-362 🙂

resources/schema.py
```diff
@@ -452,3 +607,8 @@ class Mutation:
     create_berth_type = CreateBerthTypeMutation.Field()
     delete_berth_type = DeleteBerthTypeMutation.Field()
     update_berth_type = UpdateBerthTypeMutation.Field()
+
+    # Berths
```
igordavydsson: `# Harbors`

resources/schema.py
a-rmz: Should we remove them right away or wait until 362 is developed? To ensure we have support them.

igordavydsson: well, if they work now in these mutations, then we can keep 'em I guess

a-rmz: They do work. Let's keep them then for now and I'll add a comment to the ticket to remind me to remove them.

resources/schema.py
```diff
@@ -170,6 +171,25 @@ def resolve_image_file(self, info, **kwargs):
...
+    servicemap_id = graphene.String()
+    zip_code = graphene.String()
+    phone = graphene.String()
+    email = graphene.String()
+    www_url = graphene.String()
+    location = graphql_geojson.Geometry()
+    image_link = graphene.String()
+
+
+class HarborAreaInput(AbstractAreaInput):
```
igordavydsson: you could probably rename it to `HarborInput` and have all the other fields added right here. so that eventually the mutation looks like this:
```
"createHarbor": {
    "municipality": "helsinki",
    "imageFile": "https://foobar.com/image.png",
    "availabilityLevel": "1",
    "numberOfPlaces": 150,
    "maximumWidth": 350,
    "maximumLength": 400,
    "maximumDepth": 100,
    "name": "Foobarsatama",
    "streetAddress": "Foobarsatamantie 1234",
    "servicemapId": "1",
    "location": {
      "type": "Point",
      "coordinates": [125.6, 10.1]
    },
    "zipCode": "00119",
    "phone": "123123123",
    "email": "foo@bar.com",
    "wwwUrl": "https://foobar.com",
    "imageLink": "https://foobar.com/image.png"
}
```

resources/schema.py
```diff
@@ -328,6 +348,62 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         return DeleteBerthTypeMutation()
 
 
+class CreateHarborMutation(graphene.ClientIDMutation):
+    class Input:
+        area = HarborAreaInput(required=True)
+        municipality = graphene.String()
+        image_file = graphene.String()
```
igordavydsson: uploading files through graphene is not that straightforward, we would need to use [this package](https://github.com/lmcgartland/graphene-file-upload) or some other that implements [this spec](https://github.com/jaydenseric/graphql-multipart-request-spec)

igordavydsson: if you'd like, you could skip it in this PR, and imeplement it later, we can create a ticket. would avoid extra complixity in this PR 🙂

resources/schema.py
```diff
@@ -328,6 +348,62 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        lang = info.context.META.get("HTTP_ACCEPT_LANG", settings.LANGUAGE_CODE).split(
+            "-"
+        )[0]
+
+        area = kwargs.pop("area", {})
+
+        try:
+            availability_level = AvailabilityLevel.objects.language(lang).get(
```
igordavydsson: you should accommodate the case, when `kwargs.get("availability_level")` is `None`, since it's non-required. it should only fail if the mutations contains an ID for it, but an object cannot be found

resources/schema.py
```diff
@@ -328,6 +348,62 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+    harbor = graphene.Field(HarborNode)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        lang = info.context.META.get("HTTP_ACCEPT_LANG", settings.LANGUAGE_CODE).split(
```
igordavydsson: hm, trying to track down which of our package checks it where and how, maybe using just django's [`get_language()`](https://docs.djangoproject.com/en/3.0/_modules/django/utils/translation/#get_language) would work here? could use test it? I would like to somehow avoid `get`ting from `META` 😅

resources/schema.py
a-rmz: I tried about doing something like that, either composing or extending the `AbstractAreaInput` type, but couldn't find a way to do it (not as straightforward as it sounds). That's why I decided to keep `area` as a separate input that could be reusable.

I can dedicate some more time to see if there could be another way to achieve that behavior.

resources/schema.py
a-rmz: By doing it from `META`, we're getting the language coming from the request. If I get it right, `get_language` would only tell you the default language of the running Django. But I'll take a look as well.

resources/schema.py
a-rmz: Opened [VEN-405](https://helsinkisolutionoffice.atlassian.net/browse/VEN-405)

resources/schema.py
igordavydsson: yeah, I was thinking that since graphene can handle that request header, some package somewhere must be checking for it, and I was trying to figure out where it happens and how it does it. if they also do `META.get` then I guess there is no prettier way to do it, but maybe there is. 😅  I could not find where that check happens in graphene however, so I thought maybe it's just at the Django level and maybe `get_language` would work for this. maybe it doesn't, so this needs slightly more tinkering 🙂

resources/schema.py
igordavydsson: would this work? :
```python
class AbstractAreaInput(graphene.InputObjectType):
    servicemap_id = graphene.String()
    zip_code = graphene.String()
    phone = graphene.String()
    email = graphene.String()
    www_url = graphene.String()
    location = graphql_geojson.Geometry()
    image_link = graphene.String()


class HarborInput(AbstractAreaInput):
    availability_level = graphene.ID()
    number_of_places = graphene.Int()
    maximum_width = graphene.Int()
    maximum_length = graphene.Int()
    maximum_depth = graphene.Int()
    name = graphene.String()
    street_address = graphene.String()


class CreateHarborMutation(graphene.ClientIDMutation):
    class Input:
        harbor = HarborInput(required=True)
```

not exactly the way I initially planned, where all the fields are directly declared on the `Input`, but I just don't see any ontological reason to have some of the fields within the `area` field, and some directly on the `Harbor` object.

I used `AbstractArea` only to minimize the lines of code, as there were common fields between `WinterStorageArea` and `Harbor`, not because they have some `area` entity within them.

resources/schema.py
```diff
@@ -328,6 +348,140 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            "-"
+        )[0]
+
+        area = kwargs.pop("area", {})
+
+        try:
+            availability_level_id = kwargs.pop("availability_level", None)
```
igordavydsson: ideally, this would be outside of the `try` block:
```python
availability_level_id = kwargs.pop("availability_level", None)
if availability_level_id:
    try:
        availability_level = AvailabilityLevel.objects.language(lang).get(
                    pk=availability_level_id
                )
        except AvailabilityLevel.DoesNotExist as e:
```
this way you might save avoid starting the routine, when it's not needed 🙂

resources/schema.py
charn: Django also has `get_language_from_request`

igordavydsson: 😍 probably that is what some package uses somewhere, which I could not find 👍

charn: That said, I made a view:

```python
def language_check(request, *args, **kwargs):
    print(get_language())
    print(get_language_from_request(request))
    return HttpResponse(status=200)
```

And both `get_language` and `get_language_from_request` seemed to be following the language information my browser was sending i.e. `get_language` should probably work in this case. Of course you also need to have `django.middleware.locale.LocaleMiddleware` middleware configured, which you do.

igordavydsson: but probably [`get_language`](https://github.com/django/django/blob/stable/2.2.x/django/utils/translation/trans_real.py#L231) would not catch the `HTTP_ACCEPT_LANG` header, while `get_language_from_request` seems to [do that](https://github.com/django/django/blob/stable/2.2.x/django/utils/translation/trans_real.py#L474)

charn: That is true. Though looks like `get_language` will return the active translation, which is changed by [LocaleMiddleware](https://github.com/django/django/blob/stable/2.2.x/django/middleware/locale.py#L18-L26). `LocaleMiddleware` itself is using `get_language_from_request` for extracting the language from the request.

igordavydsson: ah, true! so, `get_language` should work for our purposes 👍

resources/schema.py
a-rmz: I first thought about extending the `Input`:
```python
class CreateHarborMutation(graphene.ClientIDMutation):
    class Input(AbstractAreaInput):
        additional_fields = graphene.String()
```

But when you try to extend another `InputObjectType` from the `Input` class, it throws this error:

```
berth       | TypeError: Cannot create a consistent method resolution
berth       | order (MRO) for bases InputObjectType, Input
```

I didn't dig much into it tho. It might be doable overriding a couple of methods, but who knows. 🤷‍♂ 

--- 

Also, but doing something like
```python
class CreateHarborMutation(graphene.ClientIDMutation):
    class Input:
        harbor = HarborInput(required=True)
```
the input would have the shape:
```json
{
  "input": {
    "harbor": {
      "foo": "bar",
    }
}
```
so there would be an unnecessary additional `harbor` field, BUT it's doable.

a-rmz: You can ignore all of my previous blabbering. I found out why was it failing and how to properly do it. I'll update this on Monday with tons of changes.

berth_reservations/tests/utils.py
igordavydsson: somehow, this sneaked back in 😅

a-rmz: WHAT?

resources/schema.py
```diff
@@ -328,6 +340,52 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        lang = get_language()
+
+        availability_level_id = kwargs.pop("availability_level", None)
```
igordavydsson: actually, would it be okay to just pass `availability_level_id` to `Harbor.objects.language(lang).create(**params)`? would it not assign proper availability level or `null`, when it was not given? 🤔 would save us a few lines here.

igordavydsson: kinda like what you do for `municipality`

resources/schema.py
```diff
@@ -328,6 +340,52 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+
+        if availability_level_id:
+            try:
+                availability_level = AvailabilityLevel.objects.language(lang).get(
+                    pk=availability_level_id
+                )
+            except AvailabilityLevel.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        params = {**kwargs}
```
igordavydsson: why do you need to init a new dict? `kwargs` is dict already and if you `pop`ed the right things it should work ok 🙂

resources/schema.py
igordavydsson: Alright, looks very good the way it is now! 👍 🏅

resources/schema.py
a-rmz: The reason why I did it like that is to check if the AL exists when it's passed.

https://github.com/City-of-Helsinki/berth-reservations/pull/115/files#diff-09f1721fc193a5a442fee1cf24f5bd49R370
```python
            try:
                availability_level = AvailabilityLevel.objects.language(lang).get(
                    pk=availability_level_id
                )
            except AvailabilityLevel.DoesNotExist as e:
```

For `municipality` I'm assuming it will work bc the model comes from an external dependency (`munigeo`), but I could also check like that.

resources/schema.py
a-rmz: True 🤔 I was thinking to do it like that to keep the `kwargs` immutable, but if I'm popping then it doesn't matter that much.

resources/schema.py
igordavydsson: it does come from `munigeo`, but nothing guarantees that the database has populated data or that the user passed appropriate municipality identifier here 🤔 

I think `Harbor.objects.create(..., availability_level_id=<non-existend-id>, ...)` would raise an error (and no harbor will be created either), which would be an okay result for this case. Could you check this behaviour locally?

a-rmz: Sure, I'll confirm if it works. 👍

a-rmz: It returns the following error, so it works, maybe I'll wrap it on a `VenepaikkaGraphQLError`

```json
{
  "errors": [
    {
      "message": "insert or update on table \"resources_harbor\" violates foreign key constraint \"resources_harbor_availability_level_i_5d49f83c_fk_resources\"\nDETAIL:  Key (availability_level_id)=(110) is not present in table \"resources_availabilitylevel\".\n",
      "locations": [
        {
          "line": 59,
          "column": 3
        }
      ],
      "path": [
        "createHarbor"
      ]
    }
  ],
  "data": {
    "createHarbor": null
  }
}
```

igordavydsson: yeah, that is what we expected pretty much. hmmm, 🤔 but it would be `IntegrityError` and we would not be able to tell the frontend, which FK caused it. 🤔

a-rmz: I'm trying to figure out what actually comes out of it, and `.create()` doesn't raise an error, the error is raised until you try to access the `availability_level`

This:
```python
try:
  print("foo")
  print(availability_level_id)
  harbor = Harbor.objects.language(lang).create(**kwargs)
  print(harbor.avaliability_level)
  print("bar")
except Exception as e:
  print("bam")
  print("baz")
```

raises 
``` "'Harbor' object has no attribute 'avaliability_level'"```

So, if we want to have more control, we should check it in advance.

Comment (igordavydsson) [reply resources/schema.py]: the time has come to `git fetch && git checkout harbor-mutations` and get dirty with it 😸

Comment (a-rmz) [reply resources/schema.py]: GS

Comment (igordavydsson) [reply resources/schema.py]: ok, after some investigation, it turned out that `transaction.atomic` would eat any `DatabaseError` that could get raised during the `INSERT` (from [here and onwards](https://github.com/django/django/blob/master/django/db/transaction.py#L217)) and that is why this code would never end up in the `except` clause:
```python
try:
    harbor = Harbor.objects.language(lang).create(**kwargs)
except Exception as e:
```

so, we end up with two options:
1. try to `.get()` `AvailabilityLevel` and `Municipality` based on the IDs from the user input, but I think that is not really that necessary (at least not for the current business logic) and costs some extra hits to the DB.
2. just pass the entire input to `.create()` and let `atomic` catch and handle any exceptions. This way we do the entire business logic in just one DB hit and with fewer lines of code. We could even rename some of the input fields and it will be super succinct:
```python
class CreateHarborMutation(graphene.ClientIDMutation):
    class Input(AbstractAreaInput):
        municipality_id = graphene.String()
        availability_level_id = graphene.ID()
        number_of_places = graphene.Int()
        maximum_width = graphene.Int()
        maximum_length = graphene.Int()
        maximum_depth = graphene.Int()
        name = graphene.String()
        street_address = graphene.String()

    harbor = graphene.Field(HarborNode)

    @classmethod
    @login_required
    @superuser_required
    @transaction.atomic
    def mutate_and_get_payload(cls, root, info, **kwargs):
        # TODO: Should check if the user has permissions to
        # delete the specific resource
        lang = get_language()

        harbor = Harbor.objects.language(lang).create(**kwargs)
        return CreateHarborMutation(harbor=harbor)
```
the error message that the frontend would get from our backend is not that pretty, as it will be simply the `DatabaseError` casted to a string, but I guess they will be more or less ok with it 😅 
`open-city-profile` kinda does similar thing, when it calls [this helper function](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/utils.py#L15) from [this mutation](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/schema.py#L331)

what do you think of the simplistic option number 2?

a-rmz: Option 3 would be to do something like [this](https://docs.djangoproject.com/en/3.0/topics/db/transactions/#controlling-transactions-explicitly)

```python
@classmethod
@login_required
@superuser_required
@transaction.atomic
    def mutate_and_get_payload(cls, root, info, **kwargs):
        # TODO: Should check if the user has permissions to
        # delete the specific resource
        lang = get_language()

        truy @transaction.atomic
        harbor = Harbor.objects.language(lang).create(**kwargs)
        return CreateHarborMutation(harbor=harbor)
```

It would take a bit more effort adding that on every mutation, but would give us finer control to handle errors.

a-rmz: Option 3 would be to do something like [this](https://docs.djangoproject.com/en/3.0/topics/db/transactions/#controlling-transactions-explicitly)

```python
@classmethod
@login_required
@superuser_required
@transaction.atomic
def mutate_and_get_payload(cls, root, info, **kwargs):
  lang = get_language()

  try:
    with transaction.atomic():
      harbor = Harbor.objects.language(lang).create(**kwargs)
  except IntegrityError as e:
  return CreateHarborMutation(harbor=harbor)
```

It would take a bit more effort adding that on every mutation, but would give us finer control to handle errors.

resources/schema.py
```diff
@@ -328,6 +340,121 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+        municipality_id = kwargs.pop("municipality_id", None)
+        if municipality_id:
+            try:
+                kwargs["municipality"] = Municipality.objects.get(id=municipality_id)
+            except Municipality.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        try:
+            harbor = Harbor.objects.language(lang).create(**kwargs)
+        except Exception as e:
```
igordavydsson: can we maybe try to catch some more specific errors? IntegrityError, I guess is one.

a-rmz: Yeah, can also do that, I was aiming to catch anything unexpected, but we could either (1) catch specific errors, like Integrity, or just let it bubble in case it would happen.

a-rmz: We're not really catching such errors anywhere else, I was just thinking it could be useful, _just in case_.

igordavydsson: the reason being, `VenePaikkaError` would not be sent to sentry, other errors will, once we implement a new `GraphQLSentryView`

a-rmz: Let's remove it then, otherwise we'll be in the same situation where the `IntegrityError` is absorbed by the `transaction.atomic`.

PR #116, opened 2020-01-24
VEN-143 | Pier mutations

a-rmz:
## Description :sparkles:
* Add mutation and tests for the `Pier` resource.

## Issues :bug:
Partially closes VEN-143

## Testing :alembic:
To manually test the mutations, you can run the following queries on GraphiQL with the inputs provided. Just add valid IDs and you're ready to test. :rocket:

```graphql
query FetchPier($fetchId: ID!) {
  pier(id: $fetchId) {
    id
    properties {
      harbor {
      	id
        properties {
          name
        }
      }
      identifier
      electricity
      gate
      lighting
      mooring
      water
      wasteCollection
      suitableBoatTypes {
        id
        name
      }
    }
  }
}

mutation CreatePier($createInput: CreatePierMutationInput!) {
  createPier(input: $createInput) {
    pier {
      id
    	properties {
        harbor {
          id
        }
        identifier
        electricity
        gate
        lighting
        mooring
        water
        wasteCollection
        suitableBoatTypes {
          id
        }
      }
    }
  }
}

mutation UpdatePier($updateInput: UpdatePierMutationInput!) {
  updatePier(input: $updateInput) {
    pier {
      id
    	properties {
        harbor {
          id
        }
        identifier
        electricity
        gate
        lighting
        mooring
        water
        wasteCollection
        suitableBoatTypes {
          id
        }
      }
    }
  }
}

mutation DeletePier($deleteInput: DeletePierMutationInput!) {
  deletePier(input: $deleteInput) {
    __typename
  }
}
```

```json
{
  "fetchId": "",
  "createInput": {
    "harborId": "",
    "lighting": true,
    "wasteCollection": true,
    "mooring": true,
    "suitableBoatTypes": [ "1" ],
    "identifier": "foobar4",
    "location": {
      "type": "Point",
      "coordinates": [66.6, 99.9]
    },
    "water": true,
    "electricity": true,
    "gate": false
  },
  "updateInput": {
    "id": "",
    "suitableBoatTypes": []
  },
  "deleteInput": {
    "id": ""
  }
}
```

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -2,6 +2,7 @@ import graphene
 import graphql_geojson
 from django.db import transaction
 from django.db.models import Prefetch
+from django.db.utils import IntegrityError
 from django.utils.translation import get_language
 from graphene import relay
 from graphene_django.fields import DjangoConnectionField, DjangoListField
@@ -180,6 +181,14 @@ class AbstractAreaInput:
     www_url = graphene.String()
     location = graphql_geojson.Geometry()
     image_link = graphene.String()
+
+
+class AbstractAreaSectionInput:
+    identifier = graphene.String()
+    location = graphql_geojson.Geometry()
+    electricity = graphene.Boolean()
+    water = graphene.Boolean()
+    gate = graphene.Boolean()
 
 
 class CreateBerthMutation(graphene.ClientIDMutation):
@@ -452,6 +461,54 @@ class DeleteHarborMutation(graphene.ClientIDMutation):
         return DeleteHarborMutation()
 
 
+class PierInput(AbstractAreaSectionInput):
+    harbor_id = graphene.ID()
+    suitable_boat_types = graphene.List(graphene.ID)
+    mooring = graphene.Boolean()
+    waste_collection = graphene.Boolean()
+    lighting = graphene.Boolean()
+
+
+class CreatePierMutation(graphene.ClientIDMutation):
+    class Input(PierInput):
+        harbor_id = graphene.ID(required=True)
+
+    pier = graphene.Field(PierNode)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        suitable_boat_types = kwargs.pop("suitable_boat_types", [])
+
+        harbor_global_id = kwargs.pop("harbor_id", None)
+        if harbor_global_id:
+            harbor_id = from_global_id(harbor_global_id)[1]
+            try:
+                kwargs["harbor"] = Harbor.objects.get(pk=harbor_id)
+            except Harbor.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        boat_types = set()
+        for boat_type_id in suitable_boat_types:
+            try:
+                boat_type = BoatType.objects.get(pk=boat_type_id)
+            except BoatType.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+            boat_types.add(boat_type)
+
+        try:
+            pier = Pier.objects.create(**kwargs)
+            pier.suitable_boat_types.set(boat_types)
+        except IntegrityError as e:
+            raise VenepaikkaGraphQLError(e)
+
+        return CreatePierMutation(pier=pier)
+
+
 class Query:
     availability_levels = DjangoListField(AvailabilityLevelType)
     boat_types = DjangoListField(BoatTypeType)
@@ -581,3 +638,6 @@ class Mutation:
     create_harbor = CreateHarborMutation.Field()
     delete_harbor = DeleteHarborMutation.Field()
     update_harbor = UpdateHarborMutation.Field()
+
+    # Piers
+    create_pier = CreatePierMutation.Field()
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -509,6 +509,29 @@ class CreatePierMutation(graphene.ClientIDMutation):
         return CreatePierMutation(pier=pier)
 
 
+class DeletePierMutation(graphene.ClientIDMutation):
+    class Input:
+        id = graphene.ID(required=True)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        id = from_global_id(kwargs.get("id"))[1]
+
+        try:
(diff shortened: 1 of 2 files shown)


resources/schema.py
```diff
@@ -452,6 +461,55 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
         return DeleteHarborMutation()
 
 
+class PierInput(AbstractAreaSectionInput):
+    harbor_id = graphene.ID()
+    suitable_boat_types = graphene.List(graphene.String)
```
igordavydsson: list of IDs?
```python
suitable_boat_types = graphene.List(graphene.ID)
```

resources/schema.py
```diff
@@ -452,6 +461,55 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            harbor_global_id = from_global_id(harbor_id)[1]
+            try:
+                kwargs["harbor"] = Harbor.objects.get(pk=harbor_global_id)
+            except Harbor.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        boat_types = set([])
+        for boat_type_id in suitable_boat_types:
+            try:
+                boat_type = BoatType.objects.get(pk=boat_type_id)
```
igordavydsson: would this work:
```python
suitable_boat_types = kwargs.pop("suitable_boat_types", [])

boat_types = []
try:
    boat_types = BoatType.objects.get(pk__in=suitable_boat_types)
except BoatType.DoesNotExist as e:

...
```

resources/tests/test_resources_mutations.py
```diff
@@ -837,3 +837,119 @@ def test_update_harbor_municipality_doesnt_exist(harbor, superuser):
...
+    }
+    assert Pier.objects.count() == 0
+
+    executed = client.execute(
+        query=CREATE_PIER_MUTATION,
+        variables=variables,
+        graphql_url=GRAPHQL_URL,
+        user=superuser,
+    )
+    print(executed)
```
igordavydsson: 👀

resources/schema.py
a-rmz: The `BoatType` query uses plain IDs, not GQL Global IDs

```graphql
query BT {
  boatTypes {
    id
  }
}
```

```json
{
  "data": {
    "boatTypes": [
      {
        "id": "1"
      },
      {
        "id": "2"
      },
      {
        "id": "3"
      },
      {
        "id": "4"
      },
      {
        "id": "5"
      },
      {
        "id": "6"
      },
      {
        "id": "7"
      },
      {
        "id": "8"
      }
    ]
  }
}
```

resources/schema.py
a-rmz: Yeah, would work if there was a way to figure out if there was an invalid one.

resources/schema.py
igordavydsson: yes, plain ID can be a string or an integer in graphene

igordavydsson: I think this might be valid list `["1", 2, "3"]` 🤔

resources/schema.py
```diff
@@ -452,6 +461,55 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+
+        harbor_id = kwargs.pop("harbor_id", None)
+        if harbor_id:
+            harbor_global_id = from_global_id(harbor_id)[1]
+            try:
+                kwargs["harbor"] = Harbor.objects.get(pk=harbor_global_id)
+            except Harbor.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+
+        boat_types = set([])
```
igordavydsson: I think if you just want to initiate an empty set, you could just do `set()`

resources/schema.py
igordavydsson: yeah, ok, let's keep it like this

resources/tests/test_resources_mutations.py
```diff
@@ -837,3 +837,118 @@ def test_update_harbor_municipality_doesnt_exist(harbor, superuser):
...
     assert Harbor.objects.count() == 1
     assert_doesnt_exist("Municipality", executed)
+
+
+CREATE_PIER_MUTATION = """
+mutation CreatePier($input: CreatePierMutationInput!) {
+  createPier(input: $input) {
+    pier {
+      id
+        properties {
```
igordavydsson: indentation? 👀

resources/schema.py
```diff
@@ -510,6 +510,49 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+            try:
+                boat_type = BoatType.objects.get(pk=boat_type_id)
+            except BoatType.DoesNotExist as e:
+                raise VenepaikkaGraphQLError(e)
+            boat_types.add(boat_type)
+
+        try:
+            update_object(pier, kwargs)
+            pier.suitable_boat_types.set(boat_types)
+            pier.save()
```
igordavydsson: I think `save()` is not needed after changing m2m relations, as they are saved immediately.

resources/schema.py
```diff
@@ -510,6 +510,49 @@ def mutate_and_get_payload(cls, root, info, **kwargs):
...
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
+        # TODO: Should check if the user has permissions to
+        # delete the specific resource
+        id = from_global_id(kwargs.pop("id"))[1]
+
+        try:
+            harbor_id = kwargs.pop("harbor_id", None)
+            if harbor_id:
+                global_harbor_id = from_global_id(harbor_id)[1]
```
igordavydsson: well, this is actually `harbor_id`, but the var above could be called `global_harbor_id` 🙂

resources/schema.py
a-rmz: Another option would be to compare something like this:
```python
set(suitable_boat_types).difference(
    set(Model.objects.filter(id__in=suitable_boat_types).values_list('id'),
    flat=True)
)
```
And if there's a difference, we can raise the error saying that values X, Y, &Z  are missing.

igordavydsson: oh, pretty neat idea! probably a bit of an overengineering for this case, and this is still a hit to the db, but slightly less readable that your current solution. let's keep things as they are 🙂

resources/schema.py
a-rmz: Just FTR, it is a [valid id](https://docs.graphene-python.org/en/latest/types/scalars/).
```
graphene.ID represents a unique identifier, often used to refetch an
object or as key for a cache. The ID type appears in a JSON
response as a String; however, it is not intended to be
human-readable. When expected as an input type, any string
(such as “4”) or integer (such as 4) input value will be accepted as an ID.
```

igordavydsson: yeah, that's what I meant. 🙂

PR #118, opened 2020-01-27
VEN-152 | Rename Reservations to Applications

a-rmz:
## Description ✨ 
Rename all the `Reservation` related models to `Application` to support the new flows coming to the app.

## Issues 🐛 
Partially closes [VEN-152](https://helsinkisolutionoffice.atlassian.net/browse/VEN-152)

## Testing ⚗️ 
Running with `docker-compose`:
```shell
$ docker exec berth pytest
```

Running locally:
```shell
$ pytest
```

## Notes 🗒 
There are _a lot_ of changes.

The app is still called `reservations`, it will be renamed to `applications` but that requires a bit more tweaking, that's why we split it in different PRs, to avoid cluttering.


reservations/admin.py
```diff
@@ -12,28 +12,28 @@
...
+    export_winter_storage_applications_as_xlsx,
 )
 
 logger = logging.getLogger(__name__)
 
 
-class ReservationTypeFilter(admin.SimpleListFilter):
-    title = _("Reservation type")
+class ApplicationTypeFilter(admin.SimpleListFilter):
+    title = _("application type")
```
igordavydsson: nitpick: this is visible in the django admin and looks more official when starts with a capital letter

reservations/admin.py
```diff
@@ -12,28 +12,28 @@
...
 
     parameter_name = "berth_switch"
 
     def lookups(self, request, model_admin):
-        return (("0", _("Reservation")), ("1", _("Switch application")))
+        return (("0", _("application")), ("1", _("Switch application")))
```
igordavydsson: nitpick: same as above

reservations/admin.py
```diff
@@ -138,12 +138,12 @@ class BerthReservationAdmin(admin.ModelAdmin):
...
-    list_filter = (ReservationTypeFilter,)
-    actions = ["export_reservations", "resend_reservation_confirmation"]
+    list_display = ("created_at", "first_name", "last_name", "application_type")
+    list_filter = (ApplicationTypeFilter,)
+    actions = ["export_applications", "resend_application_confirmation"]
 
-    def reservation_type(self, obj):
-        return _("Reservation") if obj.berth_switch is None else _("Switch application")
+    def application_type(self, obj):
+        return _("application") if obj.berth_switch is None else _("Switch application")
```
igordavydsson: nitpick: same as above

reservations/admin.py
```diff
@@ -165,56 +165,56 @@ def get_berth_switch_reason(self, obj):
...
-        for reservation in queryset:
+        for application in queryset:
             try:
                 send_notification(
-                    reservation.email,
-                    NotificationType.BERTH_RESERVATION_CREATED,
-                    reservation.get_notification_context(),
-                    reservation.language,
+                    application.email,
+                    NotificationType.BERTH_application_CREATED,
```
igordavydsson: enum members should be fully capitalized per convention

a-rmz: Forgot to add `Preserve case` on Pycharm :derp:

reservations/admin.py
```diff
@@ -286,68 +286,68 @@ class WinterStorageReservationAdmin(admin.ModelAdmin):
...
-        for reservation in queryset:
+        for application in queryset:
             try:
                 send_notification(
-                    reservation.email,
-                    NotificationType.WINTER_STORAGE_RESERVATION_CREATED,
-                    reservation.get_notification_context(),
-                    reservation.language,
+                    application.email,
+                    NotificationType.WINTER_STORAGE_application_CREATED,
```
igordavydsson: needs full capitalization

reservations/schema.py
```diff
@@ -141,67 +141,67 @@ def mutate(self, info, **kwargs):
...
-            winter_storage_reservation=reservation, ok=ok
+        return CreateWinterStorageApplication(
+            winter_storage_application=application, ok=ok
         )
 
 
 class Mutation(graphene.ObjectType):
-    create_berth_reservation = CreateBerthReservation.Field()
-    create_winter_storage_reservation = CreateWinterStorageReservation.Field()
+    create_berth_application = CreateBerthApplication.Field()
```
igordavydsson: well, these should keep the old names, only internal names can be changed to `Application`

a-rmz: Are those the ones being used already?

igordavydsson: also, the types' names should remain the same from the GQL POV.

igordavydsson: otherwise frontend will complain about outdated types in their generated typings

igordavydsson: reservations/tests/test_reservations_graphql.py should pass without any changes 🙂

igordavydsson: Yes, it's the ones in use.

a-rmz: Ok ok. 👍 Reverted to old names.

reservations/schema.py
```diff
@@ -116,18 +116,18 @@ class WinterStorageReservationInput(BaseReservationInput):
 
 class CreateBerthReservation(graphene.Mutation):
     class Arguments:
-        berth_reservation = BerthReservationInput(required=True)
+        berth_application = BerthApplicationInput(required=True)
```
igordavydsson: these changes still cause a change of the API. `test_reservations_graphql.py` should pass without any changes to the tests. you can add some TODOs to remind us to rename the objects later on though

a-rmz: Would it make sense to presereve only the field names (`berth_reservation`) but already rename what we can (i.e. `BerthReservationInput -> BerthApplicationInput`)?

Or would it be better to wait and rename all the related fields at the same time?

igordavydsson: if you rename the input, it will change the API, the frontend will ask for regeneration of types. API should stay 100 % the same for now

PR #126, opened 2020-02-03
VEN-152 | Expose applications via GQL

a-rmz:
## Description ✨ 
* Add queries for single `Application` and edge list of `Applications`
* Replace `ForeignKey` relationship on `Lease -> Application` to `OneOnOne`

## Issues 🐛 
Partially closes [VEN-152](https://helsinkisolutionoffice.atlassian.net/browse/VEN-152)

## Testing ⚗️ 
To manually test the query, first, create an application from the Django Admin.

The following query returns a list with all the existing `Applications` with the whole Application object.
```graphql
query BerthApplications {
  berthApplications {
    edges {
      node {
        id
        createdAt
        isProcessed
        language
        firstName
        lastName
        email
        phoneNumber
        address
        phoneNumber
        address
        zipCode
        municipality
        companyName
        businessId
        boatType
        boatRegistrationNumber
        boatName
        boatModel
        boatLength
        boatWidth
        acceptBoatingNewsletter
        acceptFitnessNews
        acceptLibraryNews
        acceptOtherCultureNews
        informationAccuracyConfirmed
        data
        applicationCode
        harborchoiceSet {
          harbor
          priority
        }
      }
    }
  }
}
```

To query for an individual `Application`, you can use the following query:
```graphql
query SingleApplication {
  berthApplication(id: "") {
    id
  }
}
```


applications/new_schema.py
```diff
@@ -0,0 +1,64 @@
...
+        if self.boat_type:
+            return self.boat_type.id
+        return None
+
+    def resolve_harborchoice_set(self, info, **kwargs):
+        if self.harborchoice_set.count():
+            return self.harborchoice_set.all()
+
+
+class Query:
```
charn: I see that there's some inconsistency where sometimes a `Query` inherits `graphene.ObjectType` and somethimes it doesn't. I don't see any reason why it shouldn't inherit the common graphene base class. Is there?

applications/schema.py
```diff
@@ -200,7 +200,7 @@ def mutate(self, info, **kwargs):
         )
 
 
-class Mutation(graphene.ObjectType):
+class Mutation:
```
charn: Is the inheritance unnecessary? At least it serves as clarification `graphene.ObjectType` vs `graphene.Mutation`

a-rmz: I removed it on this because [this](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/resources/schema.py#L691) one didn't have it, and that one was taken from Kukkuu.

I was talking about it with @igordavydsson and came to the conclusion that, since it's already extended on the main schema, there's no need to extend it twice.
```python
class Mutation(resources.schema.Mutation, graphene.ObjectType):
    pass

new_schema = build_schema(query=Query, mutation=Mutation, types=extended_types)
```

applications/new_schema.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    def filter_berth_switch(self, queryset, name, value):
+        lookup = "__".join([name, "isnull"])
+        return queryset.filter(**{lookup: not value})
+
+
+class BerthApplicationTypeNode(DjangoObjectType):
+    boat_type = graphene.String()
+    harborchoice_set = graphene.List(HarborChoiceType)
```
igordavydsson: got this idea: could you please rename this field, since we resolve it manually anyway, to `harbor_choices`? more human-friendly 🙂

applications/schema.py
igordavydsson: yeah, it seems that this aggregators for mutations and queries are kinda like mixin objects, and do not have to inherit from `graphene.ObjectType`. would be nice to make it consistent across all apps in a separate commit.

applications/new_schema.py
```diff
@@ -0,0 +1,75 @@
...
+        model = BerthSwitch
+
+    def resolve_harbor(self, info, **kwargs):
+        if self.harbor:
+            return self.harbor.servicemap_id
+        return None
+
+
+class BerthApplicationFilter(django_filters.FilterSet):
+    has_berth_switch = django_filters.BooleanFilter(
```
igordavydsson: this is a bit of a technical way to name it, could you make it just `switch_application`? 🙂

igordavydsson: and even better: `switch_applications` 🙂

applications/new_schema.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def resolve_harborchoice_set(self, info, **kwargs):
+        if self.harborchoice_set.count():
+            return self.harborchoice_set.all()
+        return None
+
+
+class Query:
+    application = graphene.relay.Node.Field(BerthApplicationTypeNode)
+    applications = DjangoFilterConnectionField(
```
igordavydsson: also, keeping in mind that there will also be `winter_storage_applications`, could these be `berth_applications`? 🙂

leases/models.py
```diff
@@ -84,7 +84,7 @@ class WinterStorageLease(AbstractLease):
         on_delete=models.PROTECT,
         related_name="leases",
     )
-    application = models.ForeignKey(
+    application = models.OneToOneField(
```
igordavydsson: I think you also would need to rename `related_field` to maybe `lease`?

applications/new_schema.py
```diff
@@ -0,0 +1,75 @@
...
+        BerthApplicationTypeNode, filterset_class=BerthApplicationFilter
+    )
+
+    @login_required
+    @superuser_required
+    def resolve_applications(self, info, **kwargs):
+        return BerthApplication.objects.all()
```
igordavydsson: you could do some `select_related()` and `prefetch_related()` to optimize the query somewhat. check [from here onwards](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/resources/schema.py#L612) for some references

applications/new_schema.py
igordavydsson: you would then need to add `harborchoice_set` to `exclude` 🙂

applications/new_schema.py
a-rmz: What should be prefetched? I'm thinking `chosen_harbors` and `boat_type` at least (since they're the only ones with foreign keys), but what else should be included?

igordavydsson: yes, `boat_type`, `berth_switch` and `chosen_harbors` (which would in practice mean `harborchoice_set` and `harborchoice_set__harbor` or something) 🤔

igordavydsson: and `berth_switch` recursively, meaning `berth_switch__harbor` and `berth_switch__reason`

igordavydsson: and `berth_switch__reason__translations` 😅

leases/models.py
a-rmz: So renaming it from `leases` to `lease`?

igordavydsson: yes, to describe the updated logic

applications/new_schema.py
a-rmz: 👍

a-rmz: So,

```python
return BerthApplication.objects.prefetch_related(
    "boat_type",
    "berth_switch",
    "berth_switch__harbor",
    "berth_switch__reason",
    "berth_switch__reason__translations",
    "harborchoice_set__harbor",
)
```

igordavydsson: so, `prefetch_related` is for m2m relations. `select_related` is for FKs:
```python
return BerthApplication.objects.select_related(
    "boat_type",
    "berth_switch",
    "berth_switch__harbor",
    "berth_switch__reason",
    "harborchoice_set__harbor",
).prefetch_related("berth_switch__reason__translations")
```

a-rmz: I was just reading about that. 😅

applications/new_schema.py
```diff
@@ -0,0 +1,81 @@
...
+    switch_applications = django_filters.BooleanFilter(
+        field_name="berth_switch", method="filter_berth_switch"
+    )
+
+    def filter_berth_switch(self, queryset, name, value):
+        lookup = "__".join([name, "isnull"])
+        return queryset.filter(**{lookup: not value})
+
+
+class BerthApplicationTypeNode(DjangoObjectType):
```
igordavydsson: this can be called just `BerthApplicationNode` 🙂

igordavydsson: the convention we "kinda" adopted:
- regular GQL type = `<ModelName>Type`
- type with node interface = `<ModelName>Node`
any class names that do not follow this convention can be renamed at any conveneint time 🙂

applications/new_schema.py
```diff
@@ -0,0 +1,81 @@
...
+        if self.harbor:
+            return self.harbor.servicemap_id
+        return None
+
+
+class BerthSwitchType(OldBerthSwitchType):
+    harbor = graphene.String()
+
+    class Meta:
+        model = BerthSwitch
```
igordavydsson: please, `exclude = ("berthapplication_set",)` 🙂

PR #127, opened 2020-02-05
[VEN-442] Export applications to Excel

a-rmz:
## Description ✨ 
Add a `Switch reason` column.
* If the application is a `switch application`, the column should have the finnish text for the reason or `"---"` if no reason.
* If it's a regular application, the column is just empty

Add columns `Copmany name` and `Business ID`.
* If those values were given, they are added
* Otherwise leave the cells empty

## Issues 🐛 
Closes [VEN-442](https://helsinkisolutionoffice.atlassian.net/browse/VEN-442)

## Screenshots 📸 
**Berth Applications**
image

**Winter Storage Applications**
image

### DIFF
--- a/applications/tests/test_reservations_utils.py
+++ b/applications/tests/test_reservations_utils.py
@@ -18,6 +18,8 @@ from .factories import BerthApplicationFactory, WinterStorageApplicationFactory
 
 @freeze_time("2019-01-14T08:00:00Z")
 @pytest.mark.parametrize("berth_switch", [True, False])
+# Parametrised berth_switch_reason inside berth_switch
+@pytest.mark.parametrize("berth_switch_info", [True, False], indirect=True)
 def test_exporting_berth_applications_to_excel(
     berth_switch, boat_type, harbor, berth_switch_info
 ):
@@ -67,6 +69,7 @@ def test_exporting_berth_applications_to_excel(
 
     boat_type.set_current_language("fi")
 
+    expected_berth_switch_reason = ""
     expected_berth_switch_str = ""
     if berth_switch:
         expected_berth_switch_str = "{} ({}): {}".format(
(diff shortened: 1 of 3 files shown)


applications/tests/test_reservations_utils.py
```diff
@@ -18,8 +18,11 @@
 
 @freeze_time("2019-01-14T08:00:00Z")
 @pytest.mark.parametrize("berth_switch", [True, False])
+@pytest.mark.parametrize("customer_private", [True, False])
```
igordavydsson: you could put all of the fixtures in a single decorator, something like:
```python
@pytest.mark.parametrize("berth_switch, customer_private, berth_switch_info", (
    [True, False], [True, False], [True, False])
)
```

igordavydsson: ah, yeah, and that indirect could be added too

applications/utils.py
```diff
@@ -73,6 +76,14 @@ def export_berth_applications_as_xlsx(applications):
                     value = application.boat_type.name
                 elif attr_name == "berth_switch" and application.berth_switch:
                     value = parse_berth_switch_str(application.berth_switch)
+                elif attr_name == "berth_switch_reason":
```
igordavydsson:
```python
elif attr_name == "berth_switch_reason" and application.berth_switch:
    value = application.berth_switch.reason.title if application.berth_switch.reason else "---"
```
?

applications/tests/test_reservations_utils.py
a-rmz: IMO it looks a bit clearer in different decorators.

Also, it's the way the [documentation](https://docs.pytest.org/en/latest/parametrize.html) suggests.
image

applications/utils.py
a-rmz: Yeah, I thought about this, but if `application.berth_switch` is `None`, it will try to get the `berth_switch_reason` from the `Application` instance and throw an error because it doesn't have that attribute.

```python
else:
    value = getattr(application, attr_name)
```

applications/tests/test_reservations_utils.py
igordavydsson: well, the docs are a bit ambiguous 🙂 :
image
but, yeah, it does look cleaner somewhat. aight, let'se keep it.

applications/utils.py
igordavydsson: ah, right! okay then 👍

applications/tests/test_reservations_utils.py
a-rmz: Hmmm... strange, but ok. 👍

PR #129, opened 2020-02-06
VEN-152 | Mutation to assign an Application to a CustomerProfile

a-rmz:
## Description ✨ 
* Add a `customer` field to `BerthApplication`
* Expose an `UpdateApplication` mutation to add a customer to an application
* Add tests for the new mutation
  * Move some fixtures to the global scope so they can be used on different applications

## Issues 🐛
Closes [VEN-152](https://helsinkisolutionoffice.atlassian.net/browse/VEN-152) 

## Notes 🗒 
The mutation is called `UpdateApplication` to support the case in the future when applications can be updated by admins.

## Testing ⚗️ 
Query:
```graphql
mutation UpdateApplication($input: UpdateBerthApplicationInput!) {
  updateBerthApplication(input: $input) {
    application {
      id
      customer {
        id
        comment
      }
    }
  }
}
```

Input:
```json
{
  "input": {
    "berthApplicationId": "",
    "customerId": ""
  }
}
```


applications/new_schema.py
```diff
@@ -71,6 +76,34 @@ def get_node(cls, info, id):
         return super().get_node(info, id)
 
 
+class UpdateBerthApplication(graphene.ClientIDMutation):
+    class Input:
+        berth_application_id = graphene.ID(required=True)
```
igordavydsson: in the resources app you used just `id` field for all the update mutations. I suggest you do the same here and we adopt it as a convention 🙂

applications/tests/test_applications_new_schema_mutations.py
```diff
@@ -0,0 +1,119 @@
...
+
+    assert berth_application.customer is None
+
+    executed = client.execute(
+        query=UPDATE_BERTH_APPLICATION_MUTATION,
+        variables=variables,
+        graphql_url=GRAPHQL_URL,
+        user=superuser,
+    )
+    print(executed)
```
igordavydsson: 👀

applications/new_schema.py
a-rmz: True, it was a leftover from my original idea of having a `AssignApplicationToCustomer` mutation. Good catch.

applications/tests/test_applications_new_schema_mutations.py
a-rmz: 🙈

applications/new_schema.py
```diff
@@ -71,6 +76,34 @@ def get_node(cls, info, id):
         return super().get_node(info, id)
 
 
+class UpdateBerthApplication(graphene.ClientIDMutation):
+    class Input:
+        id = graphene.ID(required=True)
+        customer_id = graphene.ID(required=True)
```
igordavydsson: 🤔 if we want to make it a more general update mutation, we would need to allow for passing only those values, that the frontend wants to update, i.e. only the `id` of the application should be required. other fields should be updated if there was input.

applications/new_schema.py
```diff
@@ -71,6 +76,34 @@ def get_node(cls, info, id):
...
+
+    berth_application = graphene.Field(BerthApplicationNode)
+
+    @classmethod
+    @login_required
+    @superuser_required
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **kwargs):
```
igordavydsson: the `kwargs` here could be renamed to `input` for more clarity, [ref](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/schema.py#L361)

applications/new_schema.py
igordavydsson: hm, could you it the way profile does it with [the `InputObjectType`](https://github.com/City-of-Helsinki/open-city-profile/blob/develop/profiles/schema.py#L272)? we could refactor inputs in resources later on to follow the same convention

a-rmz: Sure. There's also something like that in [resources](https://github.com/City-of-Helsinki/berth-reservations/blob/da62f2ea6868a2f9eda35c66166d84c68b5fbec8/resources/schema.py#L464).

For now, I thought it would make sense to keep the `customer_id` as required since it's the only "updatable" field and it wouldn't make sense to run an empy mutation which just receives the application id. We can either ditch the required right away or add (yet) another `TODO` to remind removing the required once we start allowing other fields.

applications/new_schema.py
a-rmz: Agree, `**input` makes more sense, but we'd also have to refactor all the mutations in [resources](https://github.com/City-of-Helsinki/berth-reservations/blob/da62f2ea6868a2f9eda35c66166d84c68b5fbec8/resources/schema.py)

igordavydsson: yes, let's create a ticket for the refactor

applications/new_schema.py
igordavydsson: OK, fair enough! let's keep the input arguments as is and refactor when the need arises

applications/new_schema.py
igordavydsson: ok, I'll push a refactor commit

PR #131, opened 2020-02-07
VEN-428 | Add Status to BerthApplication

a-rmz:
## Description ✨ 
* Add a `Status` enum with different handling status for applications.
* Add a `status` field to the `BerthApplication` model

## Issues 🐛 
Closes [VEN-428](https://helsinkisolutionoffice.atlassian.net/browse/VEN-428)

## Testing ⚗️ 
```graphql
query GetApplications {
  berthApplications {
    edges {
      node {
        id
        status
      }
    }
  }
}
```
Should return `"PENDING"`


applications/enums.py
```diff
@@ -11,3 +11,10 @@ class Labels:
         ON_TRESTLES = _("On trestles")
         ON_TRAILER = _("On a trailer")
         UNDER_TARP = _("Under a tarp")
+
+
+class ApplicationStatus(Enum):
+    PENDING = "pending"
+    OFFER_GENERATED = "offer_generated"
+    OFFER_SENT = "offer_sent"
+    HANDLED = "handled"
```
igordavydsson: please, add `Labels`

a-rmz: Isn't that gonna be handled by the front-end?

applications/models.py
```diff
@@ -150,6 +150,14 @@ class BerthApplication(BaseApplication):
         related_name="berth_applications",
     )
 
+    status = EnumField(
+        ApplicationStatus,
+        null=False,
```
igordavydsson: `null=False` is already default

applications/enums.py
igordavydsson: these `Labels` are for the django-admin 🙂

a-rmz: Aha! Got it. :)

PR #132, opened 2020-02-07
Ensure permissions when fetching single BerthProfileNode

igordavydsson:


customers/schema.py
```diff
@@ -64,6 +64,20 @@ class Meta:
...
+        node = super().get_node(info, id)
+        if not node:
+            return None
+
+        user = info.context.user
+        # TODO: implement proper permissions
+        if node.user == user or user.is_superuser:
+            return node
+        else:
+            raise GraphQLError(_("You do not have permission to perform this action."))
```
a-rmz: Should we use [VenepaikkaGraphQLError](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/berth_reservations/exceptions.py#L4) instead of plain GQLError?

igordavydsson: we should, good catch! 👍

PR #133, opened 2020-02-10
Hotfix: Add required to harbor ID on HarborChoiceType

a-rmz:
Since we’re treating harbor as an id, it shouldn’t be a nullable value.

```graphql
type HarborChoiceType {
  harbor: String!
  priority: Int!
}
```

### DIFF
--- a/applications/new_schema.py
+++ b/applications/new_schema.py
@@ -17,29 +17,25 @@ ApplicationStatusEnum = graphene.Enum.from_enum(ApplicationStatus)
 
 
 class HarborChoiceType(DjangoObjectType):
-    harbor = graphene.String()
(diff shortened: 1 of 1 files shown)


applications/new_schema.py
```diff
@@ -37,9 +35,7 @@ class Meta:
         exclude = ("berthapplication_set",)
 
     def resolve_harbor(self, info, **kwargs):
-        if self.harbor:
-            return self.harbor.servicemap_id
-        return None
+        return self.harbor.servicemap_id
```
igordavydsson: update also the field on this Type to be `required=True` 🙂

a-rmz: Uhh, good catch!

PR #136, opened 2020-02-11
VEN-348 | Expose leases

a-rmz:
## Description ✨ 
* Split the "old" and "new" GQL schemas in two sparate files to avoid type conflicts
* Add a custom `AuthorizationNode(graphene.relay.Node)` to secure `Node` queries which require permissions
* Add queries to fetch `Leases`

## Issues 🐛 
Close [VEN-348](https://helsinkisolutionoffice.atlassian.net/browse/VEN-348)

## Test ⚗️ 
To test using `pytest`, run:
```sh
$  pytest leases/tests/test_lease_queries.py
```

To test manually, run the following queries:
```graphql
query GetLeases {
  berthLeases {
    edges {
      node {
        id
        createdAt
        modifiedAt
        status
        startDate
        endDate
        comment
        boat {
          id
          name
        }
        customer {
          id
          invoicingType
          comment
          createdAt
          modifiedAt
          boats {
            id
            name
          }
        }
        application {
          id
          customer {
            boats {
              name
            }
          }
        }
        berth {
          id
          number
          berthType {
            id
          }
        }
      }
    }
  }
}

query GetLease {
  berthLease(id: "") {
    id
    application {
      id
      customer {
        boats {
          name
        }
      }
    }
    berth {
      id
      number
      berthType {
        id
        width
        length
      }
    }
  }
}
```

### DIFF
--- a/leases/schema.py
+++ b/leases/schema.py
@@ -1,0 +1,54 @@
+import graphene
+from django.utils.translation import ugettext_lazy as _
+from graphene_django import DjangoConnectionField, DjangoObjectType
+from graphql_jwt.decorators import login_required, superuser_required
+
+from berth_reservations.exceptions import VenepaikkaGraphQLError
+from resources.schema import BerthNode
+
+from .enums import LeaseStatus
+from .models import BerthLease
+
+LeaseStatusEnum = graphene.Enum.from_enum(LeaseStatus)
+
+
+class BerthLeaseNode(DjangoObjectType):
+    berth = graphene.Field(BerthNode)
+    status = LeaseStatusEnum()
+
+    class Meta:
+        model = BerthLease
+        interfaces = (graphene.relay.Node,)
+
+    @classmethod
+    @login_required
+    def get_node(cls, info, id):
+        node = super().get_node(info, id)
+        if not node:
+            return None
+
+        user = info.context.user
+        # TODO: implement proper permissions
+        if (node.customer and node.customer.user == user) or user.is_superuser:
+            return node
+        else:
+            raise VenepaikkaGraphQLError(
+                _("You do not have permission to perform this action.")
+            )
+
+
+class Query:
+    berth_lease = graphene.relay.Node.Field(BerthLeaseNode)
+    berth_leases = DjangoConnectionField(BerthLeaseNode)
+
+    @login_required
+    @superuser_required
+    # TODO: Should check if the user has permissions to access these objects
+    def resolve_berth_leases(self, info, **kwargs):
+        return BerthLease.objects.select_related(
+            "application",
+            "application__customer",
+            "berth",
+            "berth__pier",
+            "berth__pier__harbor",
+        ).prefetch_related("application__customer__boats")
--- a/leases/tests/test_lease_queries.py
+++ b/leases/tests/test_lease_queries.py
@@ -1,0 +1,177 @@
+import pytest
+from graphql_relay import to_global_id
+
+from berth_reservations.tests.utils import (
+    assert_not_enough_permissions,
+    GraphQLTestClient,
+)
+
+client = GraphQLTestClient()
+
+GRAPHQL_URL = "/graphql_v2/"
+
+QUERY_BERTH_LEASES = """
+query GetBerthLeases {
+  berthLeases {
+    edges {
+      node {
+        id
+        status
+        startDate
+        endDate
+        comment
+        boat {
+          id
+        }
+        customer {
+          id
+          boats {
+            id
+          }
+        }
+        application {
+          id
+          customer {
+            id
+          }
+        }
+        berth {
+          id
+          number
+          berthType {
+            id
+          }
+        }
+      }
+    }
+  }
+}
+"""
+
+
+def test_query_berth_leases(superuser, berth_lease, berth_application):
+    berth_application.customer = berth_lease.customer
+    berth_application.save()
+    berth_lease.application = berth_application
+    berth_lease.save()
+
+    executed = client.execute(
+        query=QUERY_BERTH_LEASES, graphql_url=GRAPHQL_URL, user=superuser,
+    )
+
+    berth_type_id = to_global_id("BerthTypeNode", berth_lease.berth.berth_type.id)
+    berth_lease_id = to_global_id("BerthLeaseNode", berth_lease.id)
+    berth_application_id = to_global_id("BerthApplicationNode", berth_application.id)
+    customer_id = to_global_id("BerthProfileNode", berth_lease.customer.id)
+    boat_id = str(berth_lease.boat.id)
+
+    assert executed["data"]["berthLeases"]["edges"][0]["node"] == {
+        "id": berth_lease_id,
+        "status": "OFFERED",
+        "startDate": str(berth_lease.start_date),
+        "endDate": str(berth_lease.end_date),
+        "comment": berth_lease.comment,
+        "boat": {"id": boat_id},
+        "customer": {"id": customer_id, "boats": [{"id": boat_id}]},
+        "application": {"id": berth_application_id, "customer": {"id": customer_id}},
+        "berth": {
+            "id": to_global_id("BerthNode", berth_lease.berth.id),
+            "number": str(berth_lease.berth.number),
+            "berthType": {"id": berth_type_id},
+        },
+    }
+
+
+@pytest.mark.parametrize("user", ["none", "base", "staff"], indirect=True)
+def test_query_berth_leases_not_enough_permissions(user):
+    executed = client.execute(
+        query=QUERY_BERTH_LEASES, graphql_url=GRAPHQL_URL, user=user,
+    )
+
+    assert_not_enough_permissions(executed)
+
+
+QUERY_BERTH_LEASE = """
+query GetBerthLease {
+  berthLease(id: "%s") {
+    id
+    status
+    startDate
+    endDate
+    comment
+    boat {
+      id
+    }
+    customer {
+      id
+      boats {
+        id
+      }
+    }
+    application {
+      id
+      customer {
+        id
+      }
+    }
+    berth {
+      id
+      number
+      berthType {
+        id
+      }
+    }
+  }
+}
+"""
+
+
+def test_query_berth_lease(superuser, berth_lease, berth_application):
+    berth_lease_id = to_global_id("BerthLeaseNode", berth_lease.id)
+
+    berth_application.customer = berth_lease.customer
+    berth_application.save()
+    berth_lease.application = berth_application
+    berth_lease.save()
+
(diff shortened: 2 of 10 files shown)


berth_reservations/relay.py
```diff
@@ -0,0 +1,22 @@
...
+from graphene import relay
+from graphql_jwt.decorators import login_required, superuser_required
+
+from berth_reservations.exceptions import VenepaikkaGraphQLError
+
+
+class AuthorizationNode(relay.Node):
+    @classmethod
+    @login_required
+    @superuser_required
```
igordavydsson: hey now 👀

berth_reservations/relay.py
```diff
@@ -0,0 +1,22 @@
...
+    def get_node_from_global_id(cls, info, global_id, only_type=None):
+        node = super().get_node_from_global_id(info, global_id, only_type)
+
+        user = info.context.user
+        # TODO: implement proper permissions
+        if (hasattr(node, "user") and node.user == user) or user.is_superuser:
```
igordavydsson: hm, quite often it will be either `node.customer.user` or `node.owner.user` 🤔 a what a can a we a do? 🤔 🤔 🤔

leases/tests/test_lease_queries.py
```diff
@@ -0,0 +1,168 @@
...
+    assert_not_enough_permissions,
+    GraphQLTestClient,
+)
+
+client = GraphQLTestClient()
+
+GRAPHQL_URL = "/graphql_v2/"
+
+QUERY_BERTH_LEASES = """
+query GetBerthLeases {
```
igordavydsson: elsewhere indentation was 2 spaces for these query strings 🤔

berth_reservations/relay.py
igordavydsson: I think `AuthorizationNode` should somehow call child node's method, which would have proper check on the user... something along that line.

a-rmz: Can a we a do a something a like:

```python
node_user = (hasattr(node, "user") and node.user) or \
    (hasattr(node, "customer") and node.customer.user) or \
    (hasattr(node, "owner") and node.owner.user)
```
???

leases/tests/test_lease_queries.py
a-rmz: Agh, damn formatting on multi-line strings.

berth_reservations/relay.py
igordavydsson: this seems like a hack IMO, since we can't cover all the possible cases here. we should just allow for sucessful auth checks in child nodes somehow and leave it to them to know what attribute has user connected to it

berth_reservations/relay.py
igordavydsson: we only need `login_required` to ensure that the context has user set, other permissions check are explicit in the method

berth_reservations/relay.py
a-rmz: It does feel a bit hacky. How about this: We add an optional method to the classes implementing `AuthorizationNode`, say 
```python
def authenticate_user(node):
```

and in `get_node_from_global_id` we can do something like:
```python
def get_node_from_global_id(cls, info, global_id, only_type=None)
    node = super().get_node_from_global_id(info, global_id, only_type)
    user = info.context.user

    is_own_node = False
    if hasattr(cls, "authenticate_user"):
        is_own_node = cls.authenticate_user(node)
    
    if is_own_node or user.is_superuser:
        return node
    else:
        raise VenepaikkaGraphQLError(
            _("You do not have permission to perform this action.")
        )
```

and leave the implementation details for handling the user to the implementing class.

berth_reservations/relay.py
a-rmz: Yeap, took a bit to realise. :derp:

berth_reservations/relay.py
igordavydsson: looks pretty good! cool idea 👍 
I am now trying to imagine if it can at all be done through the combination of `get_node_from_global_id` and `get_node`, without custom methods. 🤔

a-rmz: Sorry to break your heart, but after some tinkering, it wouldn't be as straightforward as that.

One option would be to have pre-defined Authorization classes, depending on the user field:
```python
class UserAuthorizationNode(relay.Node):
    pass

class CustomerAuthorizationNode(relay.Node):
    pass

class OwnerAuthorizationNode(relay.Node):
    pass
```
but this might not be as scalable/maintainable.

---

Another option would be to have a custom class for every Node which is using `AuthorizationNode`, which would give us more granularity to check for specific things based on every resource, but we'd need to write and maintain even more custom classes.

```python
class BerthLeaseAuthorizationNode(AuthorizationNode):
    @classmethod
    def is_authorized(cls, node, user):
        return hasattr(node, "user") and node.user == user


class BerthLeaseNode(DjangoObjectType):
    berth = graphene.Field(BerthNode)
    status = LeaseStatusEnum()

    class Meta:
        model = BerthLease
        interfaces = (BerthLeaseAuthorizationNode,)
```

igordavydsson: I think we hit a dead end 🙈

igordavydsson: Discussed IRL, approach not needed after all, since just overriding `get_node` on each class that implements `graphene.relay.Node` and adding permission checks there is actually enough.

a-rmz: That, and assuming that, in the case and invalid id is passed to the query, it will return
```json
{
  "data": {
    "query": null
  }
}
```

whether the user is logged in, has permissions, or not.

igordavydsson: slight correction: 

> invalid id is passed

means entirely undecodable Global ID, i.e. something like `id: "foobar"`.

when a user passes something that could be decoded (e.g. `QmVydGhBcHBsaWNhdGlvbk5vZGU6aW52YWxpZC1pZA==` can be decoded to `BerthApplicationNode:invalid-id`), then `get_node` does get called on the proper node class and the permission check fire first, giving the response ~"You do not have permissions"

leases/schema.py
```diff
@@ -0,0 +1,54 @@
...
+    @classmethod
+    @login_required
+    def get_node(cls, info, id):
+        node = super().get_node(info, id)
+        if not node:
+            return None
+
+        user = info.context.user
+        # TODO: implement proper permissions
+        if (hasattr(node, "user") and node.user == user) or user.is_superuser:
```
igordavydsson: so, lease has a `customer`, which in turn has `user`

leases/tests/test_lease_queries.py
```diff
@@ -0,0 +1,168 @@
...
+        "berth": {
+            "id": to_global_id("BerthNode", berth_lease.berth.id),
+            "number": str(berth_lease.berth.number),
+            "berthType": {"id": berth_type_id},
+        },
+    }
+
+
+@pytest.mark.parametrize("user", ["none", "base", "staff"], indirect=True)
+def test_query_berth_lease_not_enough_permissions(user):
```
igordavydsson: could you actually pass some valid global id, built using the `berth_lease` fixture and then do `assert_not_enough_permissions`? I think that would be more in line with the test name.

leases/tests/test_lease_queries.py
```diff
@@ -0,0 +1,168 @@
...
+    executed = client.execute(query=query, graphql_url=GRAPHQL_URL, user=superuser,)
+
+    berth_type_id = to_global_id("BerthTypeNode", berth_lease.berth.berth_type.id)
+    berth_application_id = to_global_id("BerthApplicationNode", berth_application.id)
+    customer_id = to_global_id("BerthProfileNode", berth_lease.customer.id)
+    boat_id = str(berth_lease.boat.id)
+
+    assert executed["data"]["berthLease"] == {
+        "id": berth_lease_id,
+        "status": "OFFERED",
```
igordavydsson: why is it `"OFFERED"`? 🤔 isn't default `"PENDING"`?

leases/schema.py
```diff
@@ -0,0 +1,58 @@
...
+    @login_required
+    def get_node(cls, info, id):
+        node = super().get_node(info, id)
+        if not node:
+            return None
+
+        user = info.context.user
+        # TODO: implement proper permissions
+        if (
+            hasattr(node, "customer")
```
igordavydsson: the attributes will be there, but they might be `None`. thus, IMO a more clear check is:
```python
if (
    node.customer
    and node.customer.user
    and node.customer.user == user
) or user.is_superuser:
    return node
```

leases/tests/test_lease_queries.py
a-rmz: Idk 🤷‍♂ 
image

---

My guess is that when the lease is created is because it's been offered? Not sure.

igordavydsson: ah, I mixed it up with `BerthApplication`. right `OFFERED` is the default here 😅

a-rmz: image

leases/schema.py
a-rmz: 🤔 fair point, I was doing it like that because it was failing because it didn't have a `user` attr, but I think I should've checked for `node.customer` instead of `node.user` :derp:

leases/schema.py
```diff
@@ -0,0 +1,56 @@
...
+    @login_required
+    def get_node(cls, info, id):
+        node = super().get_node(info, id)
+        if not node:
+            return None
+
+        user = info.context.user
+        # TODO: implement proper permissions
+        if (
+            node.customer and node.customer.user and node.customer.user == user
```
igordavydsson: can we skip `and node.customer.user`? I think it would work fine with just `node.customer and node.customer.user == user`, since if `node.customer.user` is `None` it would simply return `False` 🤔

a-rmz: Good point. You can see the trauma JS left on me 😂 😅

igordavydsson: ahah, yeah, this smells of JS logic 😛

leases/tests/test_lease_queries.py
```diff
@@ -0,0 +1,178 @@
...
+    berth_lease_id = to_global_id("BerthLeaseNode", berth_lease.id)
+
+    query = QUERY_BERTH_LEASE % berth_lease_id
+    executed = client.execute(query=query, graphql_url=GRAPHQL_URL, user=user,)
+
+    assert_not_enough_permissions(executed)
+
+
+@pytest.mark.parametrize("user", ["none", "base", "staff"], indirect=True)
+def test_query_berth_lease_not_enough_permissions_invalid_id(user):
```
igordavydsson: I wouldn't mention permissions in this test's name, since they do not have any affect on this logic. maybe just `test_query_berth_lease_invalid_id`? 🤔 also not sure we need to parametrize it...

PR #140, opened 2020-02-12
VEN-435 | servicemap_id filter

a-rmz:
## Description :sparkles:
* Add a `servicemapIds` filter to the `harbors` query
  * The query takes as filter a string containing a comma serparated list of the desired IDs: `harbors(servicemapIds: ["41234", "41321"]) {}`
* Add a query `harborByServicemapId` to fetch a specific harbor

## Issues :bug:
Closes [VEN-435](https://helsinkisolutionoffice.atlassian.net/browse/VEN-435)
## Testing :alembic:
```graphql
query HarborIds {
  harbors(servicemapIds: ["41234"]) {
    edges {
      node {
        id
        properties {
          name
          servicemapId
        }
      }
    }
  }
}

query Harbor {
  harborByServicemapId(servicemapId: "41234") {
    id
    properties {
      name
      servicemapId
    }
  }
}
```

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -590,6 +590,9 @@ class Query:
     piers = DjangoFilterConnectionField(PierNode)
 
     harbor = relay.Node.Field(HarborNode)
+    harbor_by_servicemap_id = graphene.Field(
+        HarborNode, servicemap_id=graphene.String(required=True)
+    )
     harbors = DjangoFilterConnectionField(
         HarborNode, servicemap_ids=graphene.List(graphene.String)
     )
@@ -633,6 +636,9 @@ class Query:
             "harbor__availability_level__translations",
             "harbor__municipality__translations",
         ).select_related("harbor", "harbor__availability_level", "harbor__municipality")
+
+    def resolve_harbor_by_servicemap_id(self, info, **kwargs):
+        return Harbor.objects.filter(servicemap_id=kwargs.get("servicemap_id")).first()
 
     def resolve_harbors(self, info, **kwargs):
         # TODO: optimize this further
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -590,7 +590,9 @@ class Query:
     piers = DjangoFilterConnectionField(PierNode)
 
     harbor = relay.Node.Field(HarborNode)
-    harbors = DjangoFilterConnectionField(HarborNode)
+    harbors = DjangoFilterConnectionField(
+        HarborNode, servicemap_ids=graphene.List(graphene.String)
+    )
 
     winter_storage_place_type = relay.Node.Field(WinterStoragePlaceTypeNode)
     winter_storage_place_types = DjangoConnectionField(WinterStoragePlaceTypeNode)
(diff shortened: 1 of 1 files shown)


resources/schema.py
```diff
@@ -590,7 +597,10 @@ class Query:
     piers = DjangoFilterConnectionField(PierNode)
 
     harbor = relay.Node.Field(HarborNode)
-    harbors = DjangoFilterConnectionField(HarborNode)
+    harbor_by_servicemap_id = graphene.Field(
+        HarborNode, servicemap_id=graphene.String()
```
igordavydsson: maybe require it as argument?

resources/schema.py
```diff
@@ -590,7 +597,10 @@ class Query:
     piers = DjangoFilterConnectionField(PierNode)
 
     harbor = relay.Node.Field(HarborNode)
-    harbors = DjangoFilterConnectionField(HarborNode)
+    harbor_by_servicemap_id = graphene.Field(
+        HarborNode, servicemap_id=graphene.String(required=True)
+    )
+    harbors = DjangoFilterConnectionField(HarborNode, filterset_class=HarborFilter)
```
igordavydsson: does it still allow using `filter_fields` defined on `HarborNode`?

a-rmz: Now it does. 😄

resources/schema.py
```diff
@@ -89,10 +90,15 @@ class Meta:
         interfaces = (relay.Node,)
 
 
-class HarborNode(graphql_geojson.GeoJSONType):
+class HarborFilter(django_filters.FilterSet):
+    servicemap_ids = django_filters.BaseCSVFilter(
```
igordavydsson: does it expected this format: `servicemapIds: "41234,41235,41236"` or this: `servicemapIds: ["41234", "41235", "41236"]`. I was thinking of the latter one tbh 🙂

resources/schema.py
```diff
@@ -632,6 +646,9 @@ def resolve_piers(self, info, **kwargs):
             "harbor__municipality__translations",
         ).select_related("harbor", "harbor__availability_level", "harbor__municipality")
 
+    def resolve_harbor_by_servicemap_id(self, info, **kwargs):
+        return Harbor.objects.get(servicemap_id=kwargs.pop("servicemap_id", None))
```
igordavydsson: I would probably do `filter().first()` to ensure that no error is raised when wrong `servicemap_id` passed

resources/schema.py
a-rmz: I was thinking of the same, but I think GrapQL (or at least [django-filters](https://django-filter.readthedocs.io/en/master/ref/filters.html)) doesn't support that format. All the filters with expect multiple values receive them with that format.

igordavydsson: combining [`BaseInFilter`](https://django-filter.readthedocs.io/en/master/ref/filters.html#baseinfilter) with [`CharFilter`](https://django-filter.readthedocs.io/en/master/ref/filters.html#charfilter)?

resources/schema.py
```diff
@@ -643,6 +646,11 @@ def resolve_piers(self, info, **kwargs):
             "harbor__municipality__translations",
         ).select_related("harbor", "harbor__availability_level", "harbor__municipality")
 
+    def resolve_harbor_by_servicemap_id(self, info, **kwargs):
+        return Harbor.objects.filter(
+            servicemap_id=kwargs.pop("servicemap_id", None)
```
igordavydsson: you csan just do `kwargs.get("servicemap_id")` or `kwargs["servicemap_id"]` since it's now required and kwargs are not used for any object updates after this line 🙂

resources/schema.py
a-rmz: The `BaseInFilter` "_only validates that the incoming value is comma-separated_". Also, "_The secondary filter is then used to validate the individual values_", so the `CharFilter` would only valditate that each of the CSV values are chars.

Also, the `BaseInFilter` only sets the `lookup_expr="in"`, so there's no additional benefit on using it. It's easier to have the `CSVFilter` and set the `lookup_expr` rather than adding another class just to combine both filters.

a-rmz: FTR, GraphQL _does_ accept arrays as arguments, so it's a Django filters thing.

PR #144, opened 2020-02-18
VEN-447 | Dimension filters for Piers & Berths

a-rmz:
## Description :sparkles:
* Add `min width` and `min length` filters to `BerthNode`
* Add `min berth width` and `min berth length` to `piers` query and `piers` field in `harbors` and `harbor_by_servicemap_id` query
* Add an explanation/warning to the `piers` query regarding which filter to use
* Add a private `_resolve_piers` function that does all the heavylifting for the filtering for piers

## Issues :bug:
Closes [VEN-447](https://helsinkisolutionoffice.atlassian.net/browse/VEN-447)

## Testing :alembic:
```graphql
query Berths {
  piers {
    edges {
      node {
        id
        properties {
          berths(minWidth: 7, minLength: 10) {
            edges {
              node {
                id
                berthType {
                  id
                  width
                  length
                }
              }
            }
          }
        }
      }
    }
  }
}

query Piers {
  piers(minBerthWidth: 5, minBerthLength: 10) {
    edges {
      node {
        id
        properties {
          berths {
            edges {
              node {
                id
                berthType {
                  id
                  width
                  length
                }
              }
            }
          }
        }
      }
    }
  }
}

query ByServicemapID {
  harborByServicemapId(servicemapId: "41390") {
    id
    properties {
      piers(minBerthWidth: 5, minBerthLength: 17) {
        edges {
          node {
            id
            properties {
              berths {
                edges {
                  node {
                    berthType {
                      width
                      length
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}

query Harbors {
  harbors {
    edges {
      node {
        id
        properties {
          piers(minBerthWidth: 5, minBerthLength: 17) {
            edges {
              node {
                id
                properties {
                  berths {
                    edges {
                      node {
                        berthType {
                          width
                          length
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}

```

## Screenshots :camera_flash:
image

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -1,3 +1,4 @@
+import django_filters
 import graphene
 import graphql_geojson
 from django.db import transaction
@@ -122,11 +123,25 @@ class BerthTypeNode(DjangoObjectType):
     mooring_type = BerthMooringTypeEnum(required=True)
 
 
+class BerthNodeFilterSet(django_filters.FilterSet):
+    min_width = django_filters.NumberFilter(
+        field_name="berth_type__width", lookup_expr="gte"
+    )
+    min_length = django_filters.NumberFilter(
+        field_name="berth_type__length", lookup_expr="gte"
+    )
+
+    class Meta:
+        model = Berth
+        fields = ["min_width", "min_length"]
+
+
 class BerthNode(DjangoObjectType):
     class Meta:
         model = Berth
         fields = ("id", "number", "pier", "berth_type", "comment")
         interfaces = (relay.Node,)
+        filterset_class = BerthNodeFilterSet
 
 
 class HarborNode(graphql_geojson.GeoJSONType):
@@ -633,7 +648,7 @@ class Query:
     berth_types = DjangoConnectionField(BerthTypeNode)
 
     berth = relay.Node.Field(BerthNode)
-    berths = DjangoConnectionField(BerthNode)
+    berths = DjangoFilterConnectionField(BerthNode)
 
     pier = relay.Node.Field(PierNode)
     piers = DjangoFilterConnectionField(
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -1,7 +1,7 @@
 import graphene
 import graphql_geojson
 from django.db import transaction
-from django.db.models import Prefetch
+from django.db.models import Count, Prefetch, Q
 from django.db.utils import IntegrityError
 from django.utils.translation import get_language, ugettext_lazy as _
 from graphene import relay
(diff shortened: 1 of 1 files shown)


resources/schema.py
```diff
@@ -41,6 +44,44 @@ def update_object(obj, data):
     obj.save()
 
 
+def _resolve_piers(**kwargs):
+    min_width = kwargs.get("min_berth_width", 0)
+    min_length = kwargs.get("min_berth_length", 0)
+    filter_empty = (
```
igordavydsson: maybe:
```python
should_filter_out_empty = any([
    "min_berth_width" in kwargs,
    "min_berth_length" in kwargs,
])
```
or something along those lines

resources/schema.py
```diff
@@ -41,6 +44,44 @@ def update_object(obj, data):
...
+        kwargs.get("min_berth_width", None) is not None
+        or kwargs.get("min_berth_length", None) is not None
+    )
+
+    suitable_berth_types = BerthType.objects.filter(
+        width__gte=Decimal(min_width), length__gte=Decimal(min_length)
```
igordavydsson: btw, does it work without the `Decimal`? is django magic enough? 😅

resources/schema.py
```diff
@@ -70,17 +112,16 @@ class Meta:
             "lighting",
             "suitable_boat_types",
         ]
-        geojson_field = "location"
-        interfaces = (relay.Node,)
 
 
 class BerthTypeNode(DjangoObjectType):
+    width = graphene.Float(description=_("width (m)"))
```
igordavydsson: please, keep the fields where they were and keep them `required`. as I said, there is no uniformity across apps on whether fields come before `Meta` or after in the GQL type classes, but within each `schema.py` file there is such uniformity, and this file has them after the `Meta` 🙂

resources/schema.py
```diff
@@ -112,22 +153,28 @@ class Meta:
...
             return info.context.build_absolute_uri(self.image_file.url)
         else:
             return None
 
+    def resolve_piers(self, info, **kwargs):
+        return _resolve_piers(**kwargs).filter(harbor_id=self.id)
+
 
 class WinterStoragePlaceTypeNode(DjangoObjectType):
+    width = graphene.Float(description=_("width (m)"))
```
igordavydsson: same comment as above

resources/schema.py
```diff
@@ -61,6 +101,8 @@ class Meta:
 class PierNode(graphql_geojson.GeoJSONType):
     class Meta:
         model = Pier
+        geojson_field = "location"
```
igordavydsson: tiny nitpick: this distorts slightly the git history, as now you have to check the previous commit before this one to figure out why this change was done, and this change, I suppose, does not have any functional causes/requirements.

resources/schema.py
a-rmz: I totally forgot about this one 😅 I ditched all the `tmp/` commits and I think it went down the sink with them.

resources/schema.py
```diff
@@ -125,11 +126,25 @@ class Meta:
...
+        model = Berth
+        fields = ["min_width", "min_length"]
+
+
 class BerthNode(DjangoObjectType):
     class Meta:
         model = Berth
         fields = ("id", "number", "pier", "berth_type", "comment")
         interfaces = (relay.Node,)
+        filterset_class = BerthNodeFilterSet
```
igordavydsson: again, uniformity issue: [here](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/applications/new_schema.py#L122) we inject the filterset at the ConnectionField level and it kinda makes sense that way, since it is more related to this specific connection than to the Node.

a-rmz: That was my first thought as well. However, to be able to use the filter on the node (not query) level, it has to be injected on the node class.

If we inject it on the [`Query`](https://github.com/City-of-Helsinki/berth-reservations/pull/144/files#diff-09f1721fc193a5a442fee1cf24f5bd49R647) level, it will only allow using the filter if the root query is `berths`.

igordavydsson: good catch 👍

a-rmz: Having the `FilterSet` on the `Node` level allows to do both following queries without modifiying the parent (`piers`) node.

```graphql
# (showcased irl)
query Berths {
  berths(minWidth: 7, minLength: 10) {
    edges {
      node {
        id
        berthType {
          id
          width
          length
        }
      }
    }
  }
}

query Piers {
  piers {
    edges {
      node {
        id
        properties {
          berths(minWidth: 7, minLength: 10) {
            edges {
              node {
                id
                berthType {
                  id
                  width
                  length
                }
              }
            }
          }
        }
      }
    }
  }
}```

File: resources/schema.py [hunk 7]
```diff
@@ -41,6 +42,40 @@ def update_object(obj, data):
     obj.save()
 
 
+def _resolve_piers(**kwargs):
+    min_width = kwargs.get("min_berth_width", 0)
+    min_length = kwargs.get("min_berth_length", 0)
+    filter_empty_berths = any(
```
igordavydsson: could you put a comment here: "Filter out piers with no berths, that fit passed dimensions, only if the dimensions were given"?

igordavydsson: afraid it might be a bit confusing to see this without at least some context

PR #154, opened 2020-02-25
Profiles to django admin

igordavydsson:
## Description :sparkles:

- Show CustomerProfile objects in Django admin
- Make customer field blankable on BerthApplication 
- Add customer field to BerthApplication's Django admin page
- Improve Django admin pages' performance for leases

## Screenshots :camera_flash:

Berth application edit page:
image


applications/models.py
```diff
@@ -62,8 +62,6 @@ class BerthSwitch(models.Model):
 class BaseApplication(models.Model):
     created_at = models.DateTimeField(verbose_name=_("created at"), auto_now_add=True)
 
-    is_processed = models.BooleanField(verbose_name=_("is processed"), default=False)
```
a-rmz: Is this now being checked based on the status?

igordavydsson: it will be, yes, don't think FE has implemented that just yet

PR #156, opened 2020-02-26
VEN-455 | Application status filter

a-rmz:
## Description :sparkles:
* Add a filter to pass a list of desired statuses and tests

## Issues :bug:
Closes [VEN-455](https://helsinkisolutionoffice.atlassian.net/browse/VEN-455)

## Testing :alembic:
**Automated tests ⚙️**
```shell
$ pytest applications/tests/test_applications_new_schema_queries.py
```

**Manual testing 👷** 
```graphql
query Applications {
  berthApplications(statuses:["PENDING"]) {
    edges {
      node {
        id
        status
        customer {
          id
        }
      }
    }
  }
}
```

## Screenshots :camera_flash:
image

### DIFF
--- a/applications/new_schema.py
+++ b/applications/new_schema.py
@@ -122,14 +122,29 @@ class UpdateBerthApplication(graphene.ClientIDMutation):
 class Query:
     berth_application = graphene.relay.Node.Field(BerthApplicationNode)
     berth_applications = DjangoFilterConnectionField(
-        BerthApplicationNode, filterset_class=BerthApplicationFilter
+        BerthApplicationNode,
+        filterset_class=BerthApplicationFilter,
+        statuses=graphene.List(ApplicationStatusEnum),
+        description="The `statuses` filter takes a list of `ApplicationStatus` values "
+        "representing the desired statuses. If an empty list is passed, no filter will be applied "
+        "and all the results will be returned."
+        "\n\n**Requires permissions** to access applications."
+        "\n\nErrors:"
+        "\n* A value passed is not a valid status",
     )
 
     @login_required
     @superuser_required
     # TODO: Should check if the user has permissions to access these objects
     def resolve_berth_applications(self, info, **kwargs):
-        return BerthApplication.objects.select_related(
+        statuses = kwargs.pop("statuses", [])
+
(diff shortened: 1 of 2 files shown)


applications/new_schema.py
```diff
@@ -122,14 +122,36 @@ def mutate_and_get_payload(cls, root, info, **input):
 class Query:
     berth_application = graphene.relay.Node.Field(BerthApplicationNode)
     berth_applications = DjangoFilterConnectionField(
-        BerthApplicationNode, filterset_class=BerthApplicationFilter
+        BerthApplicationNode,
+        filterset_class=BerthApplicationFilter,
+        statuses=graphene.List(graphene.String),
```
igordavydsson: can it accept graphene's enums?
image

igordavydsson: would help avoid checking if such an enum exists 😏

a-rmz: Just figured out how 🤯 I was trying passing the enum value as string (in quotes) when it's not needed.

igordavydsson: yeah, according to GQL specs, if it is part of the operation - it should be without quotation marks, but if it is passed in variables, then it can be in quotation marks.

applications/tests/test_applications_new_schema_queries.py
```diff
@@ -64,3 +65,118 @@ def test_berth_applications_no_customer_filter_false(berth_application, superuse
...
+
+    assert executed["data"] == {
+        "berthApplications": {
+            "edges": [
+                {
+                    "node": {
+                        "id": to_global_id(
+                            "BerthApplicationNode", berth_application.id
+                        ),
+                        "status": str(ApplicationStatus.HANDLED).upper(),
```
igordavydsson: could be `"status": ApplicationStatus.HANDLED.name,` to avoid string manipulations 😅

applications/tests/test_applications_new_schema_queries.py
```diff
@@ -64,3 +65,118 @@ def test_berth_applications_no_customer_filter_false(berth_application, superuse
...
```
igordavydsson: samesies

PR #159, opened 2020-02-26
Make all the fields of Profile nodes nullable in our GQL API

igordavydsson:
## Description :sparkles:

1. Since ProfileNode is extended, none of its fields could be
non-nullable (i.e. required=True), because then the entire
ProfileNode will be null at the federation level, if the
profile object has no object in our database.

2. Make `BoatNode` actually implement `relay.Node` interface.

## Screenshots :camera_flash:

image

### DIFF
--- a/customers/schema.py
+++ b/customers/schema.py
@@ -1,17 +1,33 @@
 import graphene
 from django.utils.translation import ugettext_lazy as _
 from graphene import relay
+from graphene_django import DjangoConnectionField
 from graphene_django.filter import DjangoFilterConnectionField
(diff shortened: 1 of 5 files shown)


igordavydsson:
Well, making `BoatNode` actually implement `relay.Node` interface turned out to entail much more changes than I initially anticipated 😅

a-rmz:
> Well, making `BoatNode` actually implement `relay.Node` interface turned out to entail much more changes than I initially anticipated 😅

Been there, decided not to go forward with it since it wasn't necessary (and broke so many things).


Looks good now tho 👍

PR #161, opened 2020-03-05
VEN-479 | Harbor maps

a-rmz:
## Description :sparkles:
* Add an `AbstractAreaMap` with a `map_file: FileField` field
  * Add a `HarborMap` and `WinterStorageAreaMap` that implement the `AbstractAreaMap` and add `FKs` to the corresponding models.
* Add `harbor_maps` field to `HarborNode`
* Add `winter_storage_area_maps` field to `WinterStorageAreaNode`
* Add a `add_map_files` field to `HarborInput` (create & update)
* Add a `remove_map_files` field to `HarborInput` (create & update)
  * It throws an error if any value is passed when creating, since you can't remove maps from a new object

## Issues :bug:
Closes [VEN-479](https://helsinkisolutionoffice.atlassian.net/browse/VEN-479)

## Testing :alembic:
**Automated tests ⚙️**
N/A (refer to [VEN-512](https://helsinkisolutionoffice.atlassian.net/browse/VEN-512))

**Manual testing 👷** 
Using [Altair](https://altair.sirmuel.design/) client or browser extensions
Query:
```graphql
mutation UPDATE_HARBOR($id: ID!, $addMapFiles: [Upload]) {
  updateHarbor(input: {id: $id, addMapFiles: $addMapFiles}) {
    harbor {
      id
      properties {
        name
        harborMaps {
          id
          url
        }
      }
    }
  }
}
```

Variables:
```json
{
  "id": "SGFyYm9yTm9kZTpjMWJmM2UwMy1hYTZkLTQwOGYtOTYzZS1kNjU1YmNiZDQ3YjI=",
    "addMapFiles": [null, null],
}
```

Files:
```
addMapFiles.0 -> File
addMapFiles.1 -> File
```
image

**Once files have been added, you can test removing**
```graphql
mutation UPDATE_HARBOR($id: ID!, $removeMapFiles: [ID!]) {
  updateHarbor(input: {id: $id, removeMapFiles: $removeMapFiles}) {
    harbor {
      id
      properties {
        name
        harborMaps {
          id
          url
        }
      }
    }
  }
}
```

Variables:
```json
{
  "id": "SGFyYm9yTm9kZTpjMWJmM2UwMy1hYTZkLTQwOGYtOTYzZS1kNjU1YmNiZDQ3YjI=",
    "removeMapFiles": ["UUID"],
}
```

## Screenshots :camera_flash:
**createHarbor**
image

**CreateHarborMutationInput**
image

There's no mutations for `WinterStorageArea` at the moment.


resources/admin.py
```diff
@@ -60,3 +62,5 @@ class WinterStorageSectionAdmin(admin.OSMGeoAdmin):
 admin.site.register(WinterStoragePlace)
 admin.site.register(WinterStoragePlaceType)
 admin.site.register(WinterStorageSection, WinterStorageSectionAdmin)
+admin.site.register(HarborMap)
```
igordavydsson: nitpick, but these are in alphabetical order 😅

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
         return self.safe_translation_getter("name", super().__str__())
 
 
+class AbstractAreaMap(UUIDModel):
+    map_file = models.FileField(
+        upload_to=get_map_resource_media_folder,
+        storage=OverwriteStorage(),
```
igordavydsson: I think using the existing `OverwriteStorage` here would mean that when you upload a harbor map or even image_file for the harbor, the entire `/harbors/{harbor_id}/` folder would get removed and created again. I think we do not need `OverwriteStorage` anymore. Folder should remain 4eva now. I see you made some signal handler, so those should suffice.

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
...
+        verbose_name=_("map file"),
+        null=True,
```
igordavydsson: Should not be nullable or blankable, since the entire `HarborMap` or  `WinterStorageAreaMap ` object is about this field.

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
...
+    )
+
+    class Meta:
+        abstract = True
+
+
+class HarborMap(AbstractAreaMap):
+    harbor = models.ForeignKey(
+        Harbor,
+        null=True,
```
igordavydsson: should not be nullable or blankable. we do not want orphan map files 🙂

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
...
+        verbose_name=_("harbor"),
+        related_name="harbor_maps",
+        on_delete=models.CASCADE,
+    )
+
+
+class WinterStorageAreaMap(AbstractAreaMap):
+    winter_storage_area = models.ForeignKey(
+        WinterStorageArea,
+        null=True,
```
igordavydsson: same

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
...
+        abstract = True
+
+
+class HarborMap(AbstractAreaMap):
+    harbor = models.ForeignKey(
+        Harbor,
+        null=True,
+        blank=True,
+        verbose_name=_("harbor"),
+        related_name="harbor_maps",
```
igordavydsson: this could be just `"maps"`, since it is on the `Harbor` object already

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
...
+    )
+
+
+class WinterStorageAreaMap(AbstractAreaMap):
+    winter_storage_area = models.ForeignKey(
+        WinterStorageArea,
+        null=True,
+        blank=True,
+        verbose_name=_("winter storage area"),
+        related_name="winter_storage_area_maps",
```
igordavydsson: this could be just `"maps"` too 🙂

resources/models.py
```diff
@@ -252,6 +263,49 @@ def __str__(self):
...
+        on_delete=models.CASCADE,
+    )
+
+
+@receiver(post_delete, sender=HarborMap)
+@receiver(post_delete, sender=WinterStorageAreaMap)
+def delete_map_file_handler(sender, instance, **kwargs):
+    if instance.map_file:
+        file_path = Path(instance.map_file.path)
+        shutil.rmtree(file_path.parent)
```
igordavydsson: this deletes the entire parent folder? it should just delete the file.

resources/schema.py
```diff
@@ -572,6 +603,8 @@ def mutate_and_get_payload(cls, root, info, **input):
             raise VenepaikkaGraphQLError(e)
 
         update_object(harbor, input)
+        add_map_files(HarborMap, input.pop("add_map_files", []), harbor)
```
igordavydsson: I think it's better to do these before running `update_object`, so it would not try to do `set_attribute(harbor, "add_map_files", [HarborMap])` and so on.

resources/schema.py
```diff
@@ -490,6 +490,17 @@ def mutate_and_get_payload(cls, root, info, **input):
 class HarborInput(AbstractAreaInput):
     municipality_id = graphene.String()
     image_file = Upload()
+    add_map_files = graphene.List(
+        Upload,
+        description="List of map files that will be added to the existing ones belonging to the Harbor.",
+    )
+    remove_map_files = graphene.List(
```
igordavydsson: I think a cleaner solution would be to add this input field on `UpdateHarborMutation`'s `Input` directly. It feels odd, that we have a field user can pass value in, but any value would cause an error.

resources/models.py
a-rmz: AHHH, right. I misunderstood how the media paths/files were being handled, but you're right.

resources/schema.py
a-rmz: Makes more sense. 👍

resources/schema.py
a-rmz: Good catch!

resources/schema.py
```diff
@@ -862,7 +888,9 @@ class Mutation:
         description="The `imageFile` field takes an image as input. "
         "To provide the file, you have to perform the request with a client "
         "that conforms to the [GraphQL Multipart Request Spec]"
-        "(https://github.com/jaydenseric/graphql-multipart-request-spec)."
+        "(https://github.com/jaydenseric/graphql-multipart-request-spec).\n\n"
+        "Errors:"
+        "\n* removeHarborMap is passed"
```
igordavydsson: we probably do not raise this error anymore?

resources/models.py
```diff
@@ -252,6 +262,48 @@ def __str__(self):
         return self.safe_translation_getter("name", super().__str__())
 
 
+class AbstractAreaMap(UUIDModel):
+    map_file = models.FileField(
+        upload_to=get_map_resource_media_folder,
+        storage=FileSystemStorage(),
+        verbose_name=_("map file"),
+        null=False,
```
igordavydsson: `null=False, blank=False` is default anyway, no need to specify that 🙂

resources/models.py
```diff
@@ -252,6 +262,48 @@ def __str__(self):
...
+    )
+
+    class Meta:
+        abstract = True
+
+
+class HarborMap(AbstractAreaMap):
+    harbor = models.ForeignKey(
+        Harbor,
+        null=False,
```
igordavydsson: same

igordavydsson:
and when you change the `image_file` field, make sure, that you implement similar signal handler for it as well to delete the file when it is changed/deleted.

resources/schema.py
a-rmz: I knew I was missing something in the docs. 🤔

a-rmz:
I'll move the `OverwriteStorage` thing to a separate issue :)

PR #162, opened 2020-03-06
VEN-504 | Update BerthApplication status on lease changes

a-rmz:
## Description :sparkles:
* Move application `DRAFTED -> OFFER_SENT` when a lease is created
* Move application `DRAFTED -> PENDING` when a drafted lease is deleted

## Issues :bug:
Close [VEN-504](https://helsinkisolutionoffice.atlassian.net/browse/VEN-504)

## Testing :alembic:
**Automated tests ⚙️**
```shell
pytest leases/tests/test_lease_mutations.py
```

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from django.contrib.postgres.fields import JSONField
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 from enumfields import EnumField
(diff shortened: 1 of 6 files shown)


leases/schema.py
```diff
@@ -92,6 +93,9 @@ def mutate_and_get_payload(cls, root, info, **input):
 
             input["boat"] = boat
 
+        application.status = ApplicationStatus.OFFER_GENERATED
```
igordavydsson: I was thinking that we could maybe enforce in the model's `clean` method, that it cannot be `PENDING` if it has a `BerthLease` related to it. Without the lease it can be `PENDING` and `EXPIRED`, but with the lease other statuses apply. 🤔

a-rmz: Sounds like a good approach. 👍

applications/models.py
```diff
@@ -220,6 +221,36 @@ def get_notification_context(self):
             "application": self,
         }
 
+    def _validate_status(self):
+        """
+        If the BerthApplication does not have a BerthLease related to it, it can only have PENDING or EXPIRED statuses.
```
igordavydsson: wow, very long line 👀 could you split it?

applications/models.py
```diff
@@ -220,6 +221,38 @@ def get_notification_context(self):
...
+                _(
+                    "BerthApplication with no lease can only be "
+                    f"{', '.join([status.name for status in allowed_statuses_without_lease])}; "
+                    f"has {self.status.name}"
+                )
+            )
+
+    def save(self, *args, **kwargs):
+        # Ensure clean is always ran
+        self.clean()
```
igordavydsson: you should call `full_clean`. Django explains [here](https://docs.djangoproject.com/en/3.0/ref/models/instances/#validating-objects) why.

applications/tests/test_applications_new_schema_queries.py
```diff
@@ -22,6 +23,7 @@
 def test_berth_applications_no_customer_filter_true(
     berth_application, superuser_api_client
 ):
+    berth_application.lease = BerthLeaseFactory()
```
igordavydsson: why did you have to add lease to all of these applications? it's allowed to be null 🤔

a-rmz: Because we're adding "invalid" statuses to Applications without leaes, and they fail the validation
```python
 if (
    not hasattr(self, "lease")
    and self.status not in allowed_statuses_without_lease
 ):
```

igordavydsson: but the default application status is `PENDING` and it allows `application.lease` to be `None` 🤔

a-rmz: Ahhh, good catch. The first two (`test_berth_applications_no_customer_filter_true` and `test_berth_applications_no_customer_filter_false`) don't require it. I got too excited with all the copy-pasting going around.

PR #163, opened 2020-03-09
VNE-505 | Expose berth leases

a-rmz:
## Description :sparkles:
* Expose `berth_leases` on `BerthNode`
  * Only allowed for authorised users

If the query is performed without the `leases` field, it will pass for any user. If the `leases` field is added to the query and the user doesn't have `superuser` permissions, it will return an error.

## Issues :bug:
**Closes 🙅**
[VEN-505](https://helsinkisolutionoffice.atlassian.net/browse/VEN-505)

## Testing :alembic:
**Automated tests ⚙️**
```shell
# resources/tests/test_resources_queries.py#L122
$ pytest resources/tests/test_resources_queries.py::test_get_berth_with_leases

# resources/tests/test_resources_queries.py#L151
$ pytest resources/tests/test_resources_queries.py::test_get_berth_with_leases_not_enough_permissions
```

**Manual testing 👷** 
Try the following query in a browser with an existing session and in incognito mode:
```graphql
query BERTH_LEASES {
  berth(id: "QmVydGhOb2RlOjMwNWViMDJhLTk2MGUtNDZhNy05YWE5LWVmNWJhNTRmZGRiNA==") {
    leases {
      edges {
        node {
          id
        }
      }
    }
  }
}
```
The one with the session should return a list (empty or with items) with the available leases. The incognito one should raise a permissions error:
```json
{"message": "You do not have permission to perform this action"}
```

## Screenshots :camera_flash:
**BerthNode field**
image

**Berths query**
image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,11 +1,14 @@
 ## Description :sparkles:
 
 ## Issues :bug:
+**Closes :no_good_woman:**
+
+**Related :handshake:**
 
 ## Testing :alembic:
-**Automated tests ⚙️**
+**Automated tests :gear:️**
 
-**Manual testing 👷** 
+**Manual testing :construction_worker_man:** 
 
 ## Screenshots :camera_flash:
 
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -166,9 +166,14 @@ class BerthNodeFilterSet(django_filters.FilterSet):
 
 
 class BerthNode(DjangoObjectType):
+    leases = DjangoConnectionField(
+        "leases.schema.BerthLeaseNode",
+        description="**Requires permissions** to query this field.",
+    )
+
     class Meta:
         model = Berth
-        fields = ("id", "number", "pier", "berth_type", "comment", "leases")
+        fields = ("id", "number", "pier", "berth_type", "comment")
         interfaces = (relay.Node,)
         filterset_class = BerthNodeFilterSet
 
@@ -751,7 +756,11 @@ class Query:
     berth_types = DjangoConnectionField(BerthTypeNode)
 
     berth = relay.Node.Field(BerthNode)
-    berths = DjangoFilterConnectionField(BerthNode)
+    berths = DjangoFilterConnectionField(
+        BerthNode,
+        description="**Requires permissions** to query `leases` field. "
+        "Otherwise, everything is available",
+    )
 
     pier = relay.Node.Field(PierNode)
     piers = DjangoFilterConnectionField(
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -168,9 +168,14 @@ class BerthNodeFilterSet(django_filters.FilterSet):
 class BerthNode(DjangoObjectType):
     class Meta:
         model = Berth
-        fields = ("id", "number", "pier", "berth_type", "comment")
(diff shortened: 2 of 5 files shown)


.github/pull_request_template.md
```diff
@@ -1,11 +1,14 @@
 ## Description :sparkles:
 
 ## Issues :bug:
+**Closes :no_good_woman:**
+
+**Realted :handshake:**
```
igordavydsson: `Realted` 👀

resources/tests/test_resources_queries.py
```diff
@@ -118,6 +119,59 @@ def test_get_berths(berth):
     }
 
 
+def test_get_berth_with_leases(berth, superuser):
+    berth_lease = BerthLeaseFactory()
```
igordavydsson: you can pass the berth right here: 
```python
berth_lease = BerthLeaseFactory(berth=berth)
```
this way you won't have to assign the attribute on a separate line

resources/tests/test_resources_queries.py
```diff
@@ -115,7 +115,7 @@ def test_get_berths(api_client, berth):
     }
 
 
-def test_get_winter_storage_areas(api_client, winter_storage_area):
+def test_get_winter_storage_areas(winter_storage_area, api_client):
```
igordavydsson: interesting change 👀

resources/schema.py
```diff
@@ -166,6 +166,11 @@ class Meta:
 
 
 class BerthNode(DjangoObjectType):
+    leases = DjangoConnectionField(
```
igordavydsson: when you have explicit fields on the class, it's best to delete the field form the `Meta.fields`. graphene iis not always reliable in how it handles the fields that are mentioned there and might generate it's own schema, not the same as what you specified in the field on the class.

resources/tests/test_resources_queries.py
```diff
@@ -115,6 +116,60 @@ def test_get_berths(api_client, berth):
     }
 
 
+def test_get_berth_with_leases(superuser_api_client, berth):
+    berth_lease = BerthLeaseFactory(berth=berth)
+    berth_lease.save()
```
igordavydsson: I think you do not have to explicitly `save()` it

resources/tests/test_resources_queries.py
```diff
@@ -115,6 +116,60 @@ def test_get_berths(api_client, berth):
...
+        }
+    }
+
+
+@pytest.mark.parametrize(
+    "api_client", ["api_client", "user_api_client", "staff_api_client"], indirect=True
+)
+def test_get_berth_with_leases_not_enough_permissions(api_client, berth):
+    berth_lease = BerthLeaseFactory(berth=berth)
+    berth_lease.save()
```
igordavydsson: same

resources/schema.py
a-rmz: Good to know. I thought that, since I was adding the field, it also had to be added to `Meta.fields` 👍

igordavydsson: yeah, I learned this the hard way. bumped into some issues, when I had two different Nodes for the same model and graphene would use the wrong one, because it would generate its own field, overriding my explicit field on the class (where I used the right node).

PR #164, opened 2020-03-09
Refactor GraphQL Client in tests

igordavydsson:
## Description :sparkles:

1. Switch to using Graphene's test CLient for GQL tests.
2. Initialize clients for anonymous user, regular user, staff user
   and superuser as pytest fixtures.
3. Initialize separate clients for old schema and new schema (v2).

## Issues :bug:

Still needs some tests to ensure that `Accept-Language` header works
as expected using the old `GraphQLDjangoTestClient`. Coming in another PR.

## Testing :alembic:
**Automated tests ⚙️**

```shell
pytest
```

### DIFF
--- a/applications/tests/test_applications_new_schema_mutations.py
+++ b/applications/tests/test_applications_new_schema_mutations.py
@@ -4,12 +4,7 @@ from graphql_relay import to_global_id
 from berth_reservations.tests.utils import (
     assert_field_missing,
     assert_not_enough_permissions,
-    GraphQLTestClient,
 )
-
-client = GraphQLTestClient()
-
-GRAPHQL_URL = "/graphql_v2/"
 
 UPDATE_BERTH_APPLICATION_MUTATION = """
 mutation UpdateApplication($input: UpdateBerthApplicationInput!) {
@@ -26,7 +21,9 @@ mutation UpdateApplication($input: UpdateBerthApplicationInput!) {
 """
 
 
-def test_update_berth_application(superuser, berth_application, customer_profile):
+def test_update_berth_application(
+    superuser_api_client, berth_application, customer_profile
+):
     berth_application_id = to_global_id(
         "BerthApplicationNode", str(berth_application.id)
     )
@@ -39,11 +36,8 @@ def test_update_berth_application(superuser, berth_application, customer_profile
 
     assert berth_application.customer is None
 
-    executed = client.execute(
-        query=UPDATE_BERTH_APPLICATION_MUTATION,
-        variables=variables,
-        graphql_url=GRAPHQL_URL,
-        user=superuser,
+    executed = superuser_api_client.execute(
+        UPDATE_BERTH_APPLICATION_MUTATION, input=variables,
     )
 
     assert executed == {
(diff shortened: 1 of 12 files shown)


berth_reservations/tests/conftest.py
```diff
@@ -38,6 +27,31 @@ def superuser():
...
+    return create_api_client()
+
+
+@pytest.fixture
+def user_api_client(user):
+    return create_api_client(user=UserFactory())
+
+
+@pytest.fixture
+def staff_api_client():
```
a-rmz: Why do the other two user fixtures receive a user/superuser argument, but not `staff`?

igordavydsson: good catch! none of them should receive any arguments. this is remnants of my tinkering. 🙂

berth_reservations/tests/utils.py
```diff
@@ -1,9 +1,53 @@
...
+class ApiClient(GrapheneClient):
+    def execute(self, *args, **kwargs):
+        """
+        Custom wrapper on the execute method, that allows passing
+        "input_variables" as a keyword argument, which get passed
+        to kwargs["variables"]["input"] to comply with the relay
+        spec for mutations.
+        """
+
+        input_variables = kwargs.pop("input_variables", {})
```
a-rmz: Is there any specific reason to name the kwarg `input_variables` instead of just `input` or `variables`?

igordavydsson: I was thinking how to name it, and thought that:
1. `variables` should be the root variables dictionary as per GQL specs
2. `input` could work but it could seem a bit unclear what kind of input

we could go with the `input`, if it seems clear enough

applications/tests/test_applications_new_schema_mutations.py
```diff
@@ -61,39 +55,37 @@ def test_update_berth_application(superuser, berth_application, customer_profile
...
-        user=superuser,
+    executed = superuser_api_client.execute(
+        UPDATE_BERTH_APPLICATION_MUTATION, input_variables=variables,
     )
 
     assert_field_missing("customerId", executed)
 
 
-@pytest.mark.parametrize("user", ["none", "base", "staff"], indirect=True)
+@pytest.mark.parametrize("user", [None, "user", "staff"])
```
a-rmz: By replacing the parametrising with this:
```suggestion
@pytest.mark.parametrize(
    "api_client", ["api_client", "user_api_client", "staff_api_client"], indirect=True
)
```

we could avoid having to manually ask for the fixture. I feel like it's an unnecessary step considering that there's already a way to do it and less hacky.

```python
    if user:
        api_client = request.getfixturevalue("{}_api_client".format(user))
```

I agree that parametrising specifically user might make it a bit clearer, but I'm not 100% convinced on manually getting the fixtures.

igordavydsson: yeah, I spent some time debating in my head whether it's a good idea or not to do `getfixturevalue` 😅 I'll implement it the way you suggest for a couple of tests to see how it behaves and looks

berth_reservations/tests/utils.py
a-rmz: I kinda liked how it was handled in the old `GraphQLDjangoTestClient` by passing `variables`, it felt clearer (or maybe I just got used to it 🤷‍♂ ).

But I would suggest the "standard" that we already have on the [`mutations`](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/resources/schema.py#L343) and keep it as `input`
```python
    pass
```

Thoughts?

applications/tests/test_applications_new_schema_queries.py
```diff
@@ -1,31 +1,33 @@
...
+"""
+
+
+def test_berth_applications_no_customer_filter_true(
+    berth_application, superuser_api_client
+):
+    berth_application.customer = None
+    berth_application.save()
+
+    query = BERTH_APPLICATIONS_WITH_NO_CUSTOMER_FILTER_QUERY % "true"
```
a-rmz: Nicely done! 👍

berth_reservations/tests/utils.py
igordavydsson: yeah, I faced problems with the old way, since not all of the mutations follow relay spec. and it felt kinda hacky that we were forcing variables to be wrapped in an `input`. the main problem was that I needed to pass simply `variables` to the API, don't remember the use case though...
anyway, having an option to either pass `variables` dict or `input` dict, which wraps them in `{"input": variables}` is a decent solution. let's do that 👍

applications/tests/test_applications_new_schema_mutations.py
igordavydsson: ok, I'll switch to that 👍

berth_reservations/tests/utils.py
a-rmz: Yeah, coercing `variables` always to `{"variables": {"input": variables}}` could work but definitely flawed.

Like you say, if we have both options to pass either ```variables -> {"variables": variables}``` or ```input -> {"variables": {"input": input}}``` sounds more reasonable.

berth_reservations/tests/conftest.py
```diff
@@ -38,6 +27,44 @@ def superuser():
...
+    return create_api_client(user=UserFactory(is_staff=True))
+
+
+@pytest.fixture
+def superuser_api_client():
+    return create_api_client(user=UserFactory(is_superuser=True))
+
+
+@pytest.fixture
+def api_client(request, user_api_client, staff_api_client, superuser_api_client):
```
a-rmz: Is it actually ever called with parameters?

a-rmz: Oh wait, I got it. Nvm.

igordavydsson: ? we call it now like this:
```python
@pytest.mark.parametrize(
    "api_client", ["api_client", "user_api_client", "staff_api_client"], indirect=True
)
```
in many places. as we agreed 🙂

PR #167, opened 2020-03-11
Test federated profile

igordavydsson:
## Description :sparkles:

1. Use GQL classes to generate global IDs
Use GQL classes' properties to generate global IDs instead of
just strings of their names. This way potential change of the
class name will be easier to refactor in places like these.

2. Test extended ProfileNodes can be fetched by federation gateway

## Testing :alembic:
**Automated tests :gear:️**
```bash
pytest customers/
```

### DIFF
--- a/customers/tests/test_customers_schema.py
+++ b/customers/tests/test_customers_schema.py
@@ -1,13 +1,17 @@
 import pytest
 from graphql_relay import to_global_id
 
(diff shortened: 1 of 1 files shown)


a-rmz:
Good catch with 1), we should aim to have our whole codebase using that strategy. I'll create a ticket to tackle that at some point of the development.

PR #172, opened 2020-03-13
VEN-506 | Berth is_accessible

a-rmz:
## Description :sparkles:
* Add `is_accessible` field to `Berth` model
* Add tests to check for `is_accessible`
* Abstract a couple of common input fields to a `BerthInput` class
* Clean `create_berth_mutation` a bit

### Closes :no_good_woman:
**[VEN-506](https://helsinkisolutionoffice.atlassian.net/browse/VEN-506)**: add "is_accessible" BooleanField to Berth model

### Automated tests :gear:️
```shell
# Create berth (is_accessible == None)
$ pytest resources/tests/test_resources_mutations.py::test_create_berth

# Update berth (is_accessible set)
$ pytest resources/tests/test_resources_mutations.py::test_create_berth
```

```graphql
mutation UPDATE_BERTH($input: UpdateBerthMutationInput!) {
  updateBerth(input: $input) {
    berth {
      id
      isAccessible
    }
  }
}
```

**Input:**
```json
{
  "input": {
    "id": "",
    "isAccessible": true
  }
}
```


resources/schema.py
```diff
@@ -170,6 +170,7 @@ class BerthNode(DjangoObjectType):
         "leases.schema.BerthLeaseNode",
         description="**Requires permissions** to query this field.",
     )
+    is_accessible = graphene.Boolean()
```
igordavydsson: I think it's one of those fields that graphene could generate itself, if you put it in `Meta.fields.`

a-rmz: I tried that, but it was failing:
```
graphql.error.base.GraphQLError: Cannot return null for non-nullable field BerthNode.isAccessible.
```

so adding it explicitly removed the `non-nullable`-ness. But it's a strange situation, I also think just adding it to the `fields=[]` should be enough.

igordavydsson: ah, ok, probably a bug in graphene where it does not recognize nullable boolean fields 😅

PR #178, opened 2020-03-19
Catch error raised when trying to remove files from filesystem

igordavydsson:


igordavydsson:
good idea! returning it to the user is a bit too technical, since most users do not care about the file in the filesystem, but sending it to sentry is something we could benefit from as devs 👍

PR #179, opened 2020-03-24
VEN-453 | User permissions

a-rmz:
## Description :sparkles:
- Add Group permission fixtures
- Add a management command to set model permissions
- Add decorators (and util functions) to check for permissions
- Add the required pytest fixtures to test for the new permissions/groups
- Update all the tests checking for permissions to fit the new schema

### Closes :no_good_woman:
[VEN-453](https://helsinkisolutionoffice.atlassian.net/browse/VEN-523)

### Related :handshake:
[VEN-363](https://helsinkisolutionoffice.atlassian.net/browse/VEN-363)

### Automated tests :gear:️
```shell
$ pytest
```

1. Load the data from the `perms.json` fixture
```shell
$ python manage.py loaddata groups.json
```
2. Run the script to set the model permissions
```shell
$ python manage.py set_group_model_permissions
```
3. Add your user to the one of the groups via [django admin](http://localhost:8000/admin)
4. Try to access the protected resources

The `superuser_api_client` works for cases when the permissions are not relevant for the test, for example:
- Passing the ID of an object that doesn't exist
- Passing an invalid ID

The document stating the pemissions and roles is [User roles @ confluence](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/DD/pages/53837874/User+roles)

### DIFF
--- a/applications/new_schema.py
+++ b/applications/new_schema.py
@@ -3,11 +3,12 @@ import graphene
 from django.db import transaction
 from graphene_django.filter import DjangoFilterConnectionField
 from graphene_django.types import DjangoObjectType
-from graphql_jwt.decorators import login_required, superuser_required
 from graphql_relay import from_global_id
 
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.models import CustomerProfile
+from leases.models import BerthLease
+from users.decorators import change_permission_required, view_permission_required
 
 from .enums import ApplicationStatus
 from .models import BerthApplication, BerthSwitch, HarborChoice
@@ -84,9 +85,7 @@ class BerthApplicationNode(DjangoObjectType):
         return None
 
     @classmethod
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access this specific object
+    @view_permission_required(BerthApplication, BerthLease, CustomerProfile)
     def get_node(cls, info, id):
         return super().get_node(info, id)
 
@@ -103,11 +102,10 @@ class UpdateBerthApplication(graphene.ClientIDMutation):
     berth_application = graphene.Field(BerthApplicationNode)
 
     @classmethod
-    @login_required
-    @superuser_required
+    @view_permission_required(CustomerProfile, BerthLease)
+    @change_permission_required(BerthApplication)
     @transaction.atomic
     def mutate_and_get_payload(cls, root, info, **input):
-        # TODO: Should check if the user has permissions to perform the following changes
         berth_application_id = from_global_id(input.get("id"))[1]
         customer_id = from_global_id(input.get("customer_id"))[1]
 
@@ -138,9 +136,7 @@ class Query:
         "\n* A value passed is not a valid status",
     )
 
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access these objects
+    @view_permission_required(BerthApplication, BerthLease, CustomerProfile)
     def resolve_berth_applications(self, info, **kwargs):
         statuses = kwargs.pop("statuses", [])
 
--- a/applications/tests/test_applications_new_schema_mutations.py
+++ b/applications/tests/test_applications_new_schema_mutations.py
@@ -21,9 +21,10 @@ mutation UpdateApplication($input: UpdateBerthApplicationInput!) {
 """
 
 
-def test_update_berth_application(
-    superuser_api_client, berth_application, customer_profile
-):
+@pytest.mark.parametrize(
+    "api_client", ["berth_services"], indirect=True,
+)
+def test_update_berth_application(api_client, berth_application, customer_profile):
     berth_application_id = to_global_id(
         "BerthApplicationNode", str(berth_application.id)
     )
@@ -36,9 +37,7 @@ def test_update_berth_application(
 
     assert berth_application.customer is None
 
-    executed = superuser_api_client.execute(
-        UPDATE_BERTH_APPLICATION_MUTATION, input=variables,
-    )
+    executed = api_client.execute(UPDATE_BERTH_APPLICATION_MUTATION, input=variables)
 
     assert executed == {
         "data": {
@@ -55,36 +54,36 @@ def test_update_berth_application(
     }
 
 
-def test_update_berth_application_no_application_id(
-    superuser_api_client, customer_profile
-):
+@pytest.mark.parametrize(
+    "api_client", ["berth_services"], indirect=True,
+)
+def test_update_berth_application_no_application_id(api_client, customer_profile):
     variables = {
         "customerId": to_global_id("BerthProfileNode", str(customer_profile.id)),
     }
 
-    executed = superuser_api_client.execute(
-        UPDATE_BERTH_APPLICATION_MUTATION, input=variables,
-    )
+    executed = api_client.execute(UPDATE_BERTH_APPLICATION_MUTATION, input=variables)
 
     assert_field_missing("id", executed)
 
 
-def test_update_berth_application_no_customer_id(
-    superuser_api_client, berth_application
-):
+@pytest.mark.parametrize(
+    "api_client", ["berth_services"], indirect=True,
+)
+def test_update_berth_application_no_customer_id(api_client, berth_application):
     variables = {
         "id": to_global_id("BerthApplicationNode", str(berth_application.id)),
     }
 
-    executed = superuser_api_client.execute(
-        UPDATE_BERTH_APPLICATION_MUTATION, input=variables,
-    )
+    executed = api_client.execute(UPDATE_BERTH_APPLICATION_MUTATION, input=variables)
 
     assert_field_missing("customerId", executed)
 
 
 @pytest.mark.parametrize(
-    "api_client", ["api_client", "user_api_client", "staff_api_client"], indirect=True
+    "api_client",
+    ["api_client", "berth_handler", "berth_supervisor", "harbor_services", "user"],
+    indirect=True,
 )
 def test_update_berth_application_not_enough_permissions(
     api_client, berth_application, customer_profile
--- a/leases/schema.py
+++ b/leases/schema.py
@@ -5,15 +5,21 @@ from django.db import transaction
 from django.utils.translation import ugettext_lazy as _
 from graphene_django import DjangoObjectType
 from graphene_django.filter import DjangoFilterConnectionField
-from graphql_jwt.decorators import login_required, superuser_required
+from graphql_jwt.decorators import login_required
 from graphql_relay import from_global_id
 
 from applications.enums import ApplicationStatus
 from applications.models import BerthApplication
 from berth_reservations.exceptions import VenepaikkaGraphQLError
-from customers.models import Boat
+from customers.models import Boat, CustomerProfile
 from resources.models import Berth
 from resources.schema import BerthNode
+from users.decorators import (
+    add_permission_required,
+    delete_permission_required,
+    view_permission_required,
+)
+from users.utils import user_has_view_permission
 
 from .enums import LeaseStatus
 from .models import BerthLease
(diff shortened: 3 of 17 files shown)


users/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    if perm not in VALID_PERMS:
+        raise ValueError("The permission to check is not a valid type")
+
+    def wrapper(user):
+        return user.has_perms([_build_permstring(perm, model) for model in models])
+
+    return wrapper
+
+
+def user_has_view_permission(user, *models):
```
igordavydsson: were you trying to add these or delete them? 😅 they are in code purgatory right now, since they are not used anywhere

applications/new_schema.py
```diff
@@ -84,9 +85,7 @@ def resolve_harbor_choices(self, info, **kwargs):
         return None
 
     @classmethod
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access this specific object
+    @view_permission_required(BerthApplication)
```
igordavydsson: what do you think? should we list all the types that `BerthApplicationNode` has relations to, like `HarborChoice, CustomerProfile, Company, Boat, Lease`? or is it safe / sane enough to only check for the top level model, i.e. `BerthApplication`? they all are within the same business logic category in the `User roles` table, but technically different models. I am not sure. 🤔

applications/new_schema.py
```diff
@@ -138,9 +136,9 @@ class Query:
         "\n* A value passed is not a valid status",
     )
 
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access these objects
+    @view_permission_required(
+        BerthApplication, BerthSwitch, Harbor, BerthSwitchReason, HarborChoice
```
igordavydsson: kinda related to the comment above. however, here you check also for `Harbor`, which is an open model with no view permissions required. should we maybe skip it from this list?

applications/tests/test_applications_new_schema_mutations.py
```diff
@@ -55,36 +54,36 @@ def test_update_berth_application(
...
-    )
+    executed = api_client.execute(UPDATE_BERTH_APPLICATION_MUTATION, input=variables,)
 
     assert_field_missing("customerId", executed)
 
 
 @pytest.mark.parametrize(
-    "api_client", ["api_client", "user_api_client", "staff_api_client"], indirect=True
+    "api_client",
+    ["api_client", "berth_handler", "berth_supervisor", "harbor_services", "user"],
```
igordavydsson: I think that `"api_client"` on this line might be a leftover from the copypasta 😅

applications/tests/test_applications_new_schema_mutations.py
```diff
@@ -36,9 +37,7 @@ def test_update_berth_application(
 
     assert berth_application.customer is None
 
-    executed = superuser_api_client.execute(
-        UPDATE_BERTH_APPLICATION_MUTATION, input=variables,
-    )
+    executed = api_client.execute(UPDATE_BERTH_APPLICATION_MUTATION, input=variables,)
```
igordavydsson: too bad black does not delete the trailing commas, after it combines several lines into one. 😞 could you delete the trailing commans manually? :derp: I mean the one after `input=variables`

applications/tests/test_applications_new_schema_mutations.py
igordavydsson: ah, right, it will have `AnonymousUser` as `info.context.user` right. Already forgot what that I did it during my refactor 😄

users/utils.py
igordavydsson: ah, you use this within resolvers as utils funcs. 👍 gotcha!

leases/schema.py
```diff
@@ -65,12 +72,10 @@ class Input:
     berth_lease = graphene.Field(BerthLeaseNode)
 
     @classmethod
-    @login_required
-    @superuser_required
+    @view_permission_required(BerthLease, BerthApplication, Berth, Boat)
+    @add_permission_required(BerthLease)
```
igordavydsson: also interesting question: do we need to specify `view_permission_required(BerthLease)` if it is included in `add_permission_required(BerthLease)` 🤔

leases/tests/test_lease_queries.py
```diff
@@ -168,7 +178,7 @@ def test_query_berth_lease(superuser_api_client, berth_lease, berth_application)
 
 
 @pytest.mark.parametrize(
-    "api_client", ["api_client", "user_api_client", "staff_api_client"], indirect=True
+    "api_client", ["api_client", "user", "harbor_services"], indirect=True
```
igordavydsson: should `test_query_berth_lease_invalid_id` down below use `superuser_ api_client`? 🤔

resources/schema.py
```diff
@@ -733,12 +711,9 @@ class Input:
         id = graphene.ID(required=True)
 
     @classmethod
-    @login_required
-    @superuser_required
+    @change_permission_required(Pier)
```
igordavydsson: `@delete_permission_required(Pier)`? 👀

resources/tests/test_resources_queries.py
```diff
@@ -10,6 +10,18 @@
...
+    [
+        "api_client",
+        "user",
+        "harbor_services",
+        "berth_supervisor",
+        "berth_handler",
+        "berth_services",
+    ],
+    indirect=True,
+)
```
igordavydsson: I think it's enough if we use the non-parametrized `api_client` with the `AnonymousUser` in it 🙂

applications/new_schema.py
a-rmz: Good catch, I think we should check for the non-public models, since the gruop permissions could _theoretically_ change, so better safe than sorry.

applications/new_schema.py
a-rmz: And good catch here as well, I'll be sure to remove the public models from the `view_permissions` tests.

a-rmz: On top of it: `BerthSwitchReason` belongs to the `applications` app, so as per the `User roles` doc it has the same permissions as `BerthApplication`, but it doesn't make much sense to "hide" it since it's only general information. Should we change the perms so that every role can `"view"` it?

igordavydsson: good point: `BerthSwitchReason` was not properly mentioned anywhere, but it is a public model. every role should have `"view"` permissions, but also since it is public we should not check if the user has permissions to it in the schema files.

leases/schema.py
a-rmz: I was thinking about whether to include it or not because, at least logically, the `add` would mean that you can `view` it as well.

Technically, in practice, you could have `add` perms _without_ `view` perms, so I settled for explicitly checking for both. BUT if we agree that on our logic `add`, `change` or `delete` _require_ to have `view`, then we can remove it.

Thoughts?

applications/new_schema.py
igordavydsson: Ok, let's see how it will look if we check for all relevant models. 👍

resources/schema.py
a-rmz: Whoops, copypasta. 😅

leases/schema.py
igordavydsson: > if we agree that on our logic add, change or delete require to have view, then we can remove it.

I think the code will be cleaner if we do accept this. 👍

a-rmz: Alrighto! Let's go on with this then :)

leases/tests/test_lease_queries.py
a-rmz: Terrrrrrific catch. 👍

applications/new_schema.py
```diff
@@ -138,9 +135,7 @@ class Query:
         "\n* A value passed is not a valid status",
     )
 
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access these objects
+    @view_permission_required(BerthApplication, BerthSwitch, HarborChoice)
```
igordavydsson: our GQL types are very intertwined. from `BerthApplicationNode` you can access `ProfileNode` and from `ProfileNode` you can access `BoatNode` and `CompanyType` 🤔

a-rmz: Then we can replace it for:
```suggestion
    @view_permission_required(BerthApplication, CustomerProfile)
```

So `BerthApplication` covers `Application`, `Switch` and `HarborChoice` (since they all share the same perms) and `CustomerProfile` that covers `Boat` and `Company`.

igordavydsson: yes, let's follow this pattern

applications/new_schema.py
```diff
@@ -84,9 +84,7 @@ def resolve_harbor_choices(self, info, **kwargs):
         return None
 
     @classmethod
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access this specific object
+    @view_permission_required(BerthApplication, CustomerProfile)
```
igordavydsson: you can also access `BerthLease` from `BerthApplication`

igordavydsson: please, check for it in other decorators in this file too

leases/schema.py
```diff
@@ -44,8 +50,9 @@ def get_node(cls, info, id):
             return None
 
         user = info.context.user
-        # TODO: implement proper permissions
-        if (node.customer and node.customer.user == user) or user.is_superuser:
+        if (node.customer and node.customer.user == user) or user_has_view_permission(
+            user, BerthLease, BerthApplication, Berth
```
igordavydsson: `Berth` is a public model

igordavydsson: `CusromerProfile` is not and you can access it from `BerthLease`

leases/schema.py
```diff
@@ -65,12 +72,10 @@ class Input:
     berth_lease = graphene.Field(BerthLeaseNode)
 
     @classmethod
-    @login_required
-    @superuser_required
+    @view_permission_required(BerthApplication, Boat)
```
igordavydsson: replace `Boat` with `CustomerProfile`, since you can access the entire profile from the lease

leases/schema.py
```diff
@@ -119,12 +124,10 @@ class Input:
         id = graphene.ID(required=True)
 
     @classmethod
-    @login_required
-    @superuser_required
+    @view_permission_required(BerthApplication)
```
igordavydsson: well, you do not have access to the application's info through this mutation, so no need to check for this

leases/schema.py
```diff
@@ -154,9 +157,7 @@ class Query:
         description="`BerthLeases` are ordered by `createdAt` in ascending order by default.",
     )
 
-    @login_required
-    @superuser_required
-    # TODO: Should check if the user has permissions to access these objects
+    @view_permission_required(BerthLease, BerthApplication, CustomerProfile, Boat)
```
igordavydsson: `Boat` is covered by `CustomerProfile`

applications/new_schema.py
```diff
@@ -103,11 +102,10 @@ class Input(BerthApplicationInput):
     berth_application = graphene.Field(BerthApplicationNode)
 
     @classmethod
-    @login_required
-    @superuser_required
+    @view_permission_required(CustomerProfile)
```
igordavydsson: you will also be able to see the `BerthLease` :derp:

resources/schema.py
```diff
@@ -178,8 +189,7 @@ class Meta:
         interfaces = (relay.Node,)
         filterset_class = BerthNodeFilterSet
 
-    @login_required
-    @superuser_required
+    @view_permission_required(BerthLease, BerthApplication, BerthLease)
```
igordavydsson: second `BerthLease` should probably be `CustomerProfile` 👀

applications/new_schema.py
a-rmz: So close yet so far 😭 I even thought about it, but apparently thought it wasn't necessary.

resources/schema.py
a-rmz: :derp: good catch, again.

applications/new_schema.py
```diff
@@ -103,11 +102,10 @@ class Input(BerthApplicationInput):
...
+    @change_permission_required(BerthApplication, BerthLease)
```
igordavydsson: why in `change`? should be in `view`

a-rmz: 🤦‍♂ Quarantine day 3 got me like:

PR #182, opened 2020-03-27
VEN-478 | Block resources associated to leases from deletion

a-rmz:
## Description :sparkles:
* Raise an error when a resource (`Harbor`, `Pier`, or `Berth`) associated to a Lease is being deleted
  * _Only_ raises the error when the Lease has a `PAID` status
  * If the resource has associated resources but neither is `PAID`, the leases associated to the resource will be deleted, and then the resource.

### Closes :no_good_woman:
**[VEN-478](https://helsinkisolutionoffice.atlassian.net/browse/VEN-478):** Do not allow deleting resources if they have connected active leases

### Automated tests :gear:️
```shell
# Harbor
$ pytest resources/tests/test_resources_mutations.py::test_delete_harbor_protected_with_lease

# Pier
$ pytest resources/tests/test_resources_mutations.py::test_delete_pier_protected_with_lease

# Berth
$ pytest resources/tests/test_resources_mutations.py::test_delete_berth_protected_with_lease
```

1. Create a Berth Lease
2. Set the status as `PAID`
3. Try to delete the `berth`, `pier`, or `harbor` associated to the `lease`
  3.1. You should get an error saying `"Cannot delete {resource} because it has some related leases"`
4. Move the status to any other than `PAID`
5. Try to delete the resource again
6. The delete should be successful
  6.1. The lease should have been removed as well

Currently, there are no mutations that allow deleting `Winter` resources, but it has to be implemented at some point.

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -16,6 +16,7 @@ from munigeo.models import Municipality
 from applications.models import BerthApplication
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.models import CustomerProfile
+from leases.enums import LeaseStatus
 from leases.models import BerthLease
 from users.decorators import (
     add_permission_required,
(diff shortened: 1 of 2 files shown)


resources/schema.py
```diff
@@ -120,6 +121,20 @@ def _resolve_piers(info, **kwargs):
     return query
 
 
+def delete_inactive_leases(lookup, model_name):
+    # Get all the leases related to the Harbor
+    leases = BerthLease.objects.filter(lookup)
+    print(leases)
```
igordavydsson: 👀 why did it pass travis/flake8 🤔

a-rmz: LOL :derp:

PR #183, opened 2020-03-27
Allow JWT tokens to be valid for 1 hour

igordavydsson:
## Description :sparkles:

This will allow us using JWT's `exp` value for up to 60 mins 
in the future. Default `OIDC_LEEWAY` considers them invalid
after 15 mins.


igordavydsson:
@a-rmz seems to be sane / safe default. tunnistamo toekns are valid for an hour by default

Nizar-Rahme:
So we're not going to consider the `exp` value from JWT?

igordavydsson:
@Nizar-Rahme we are considering it. this `OIDC_LEEWAY` is the first check and then we check `exp`.

PR #184, opened 2020-03-30
VEN-509 | Safely decode GQL global ids

a-rmz:
## Description :sparkles:
- Add a util function to safely decode global GQL ids
- Update mutations to use safe decode of global ids

### Closes :no_good_woman:
**[VEN-509](https://helsinkisolutionoffice.atlassian.net/browse/VEN-509):** Decoding global IDs has potential problems.

### Automated tests :gear:️
```shell
$ pytest
```

Execute a mutation that requires decoding an ID.
```graphql
mutation HARBOR {
  updateHarbor(input: { id: "clearly_an_invalid_id" }) {
    harbor {
      id
    }
  }
}
```
It should return an error: `"Harbor matching query does not exist."`.

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -4,16 +4,16 @@ import graphql_geojson
 from django.db import transaction
 from django.db.models import Count, Prefetch, Q
 from django.db.utils import IntegrityError
-from django.utils.translation import get_language, ugettext_lazy as _
+from django.utils.translation import ugettext_lazy as _
 from graphene import relay
 from graphene_django.fields import DjangoConnectionField, DjangoListField
 from graphene_django.filter import DjangoFilterConnectionField
 from graphene_django.types import DjangoObjectType
 from graphene_file_upload.scalars import Upload
-from graphql_relay import from_global_id
 from munigeo.models import Municipality
 
 from applications.models import BerthApplication
+from applications.new_schema import BerthApplicationNode
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.models import CustomerProfile
 from leases.enums import LeaseStatus
@@ -25,6 +25,7 @@ from users.decorators import (
     view_permission_required,
 )
 from users.utils import user_has_view_permission
+from utils.relay import get_node_from_global_id
 
 from .enums import BerthMooringType
 from .models import (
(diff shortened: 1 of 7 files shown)


resources/schema.py
```diff
@@ -268,6 +267,10 @@ class Meta:
         "\n * Filter `forApplication` combined with either dimension (width, length) filter",
     )
 
+    @classmethod
+    def get_node(cls, info, id):
+        return Harbor.objects.language(get_language()).get(pk=id)
```
igordavydsson: what made you add this?

a-rmz: We were fetching the harbor using the `language()` queryset on the [`UpdateHarborMutation`](https://github.com/City-of-Helsinki/berth-reservations/pull/184/files#diff-09f1721fc193a5a442fee1cf24f5bd49L597)
```python
lang = get_language()
harbor = get_node_from_global_id(info, input.pop("id"), only_type=HarborNode)

try:
    harbor = Harbor.objects.language(lang).get(pk=id)
```

But since we don't have direct access to the queryset anymore using the `get_node_from_global_id`, you can't explicitly call the language filter.

However, `get_node_from_global_id` internally calls `get_node`, so, by overriding `get_node` and forcing the query to use `language` we ensure that it's always returning the one with the correct language even if we can manually get it.

resources/schema.py
```diff
@@ -454,11 +454,7 @@ class Input:
     @add_permission_required(BerthType)
     @transaction.atomic
     def mutate_and_get_payload(cls, root, info, **input):
-        berth_type = BerthType.objects.create(
-            mooring_type=input.get("mooring_type"),
-            width=input.get("width"),
-            length=input.get("length"),
-        )
+        berth_type = BerthType.objects.create(**input)
```
igordavydsson: yeah, good catch, nice and clean 👍

resources/tests/test_resources_mutations.py
```diff
@@ -971,8 +971,8 @@ def test_create_pier_not_enough_permissions(api_client):
     assert_not_enough_permissions(executed)
 
 
-def test_create_harbor_harbor_doesnt_exist(superuser_api_client):
-    variables = {"harborId": to_global_id(BerthNode._meta.name, uuid.uuid4())}
+def test_create_pier_harbor_doesnt_exist(superuser_api_client):
```
igordavydsson: good catch 😅

a-rmz: Yeah, sneaky test. 😅

resources/schema.py
igordavydsson: as locally tested, Django's `LocaleMiddleware` can handle the language itself, no need for us to pass it explicitly in resolvers.

igordavydsson: you can probably delete other lines with `get_language` from this file 🙂

PR #185, opened 2020-04-02
VEN-362 | Dynamically calculate dimensions and places

a-rmz:
## Description :sparkles:
* Replace hard-coded "maximums" (width, length, depth) with `annotated` fields
* Replace hard-coded place counts with a `Count` of all available berths/places
  * Show those fields as `read-only` fields on the Admin site
* Remove "maximums" from mutations input
* Remove hard-coded values from `helsinki-resources.json` fixture 
* Change `WinterStorageArea.max_length_of_section_spaces` to `DecimalField` for consistency with the rest of the dimension properties
* Update README.md to include fixtures missing

### Closes :no_good_woman:
**[VEN-362](https://helsinkisolutionoffice.atlassian.net/browse/VEN-362):** Calculate dynamically total # of berths and max dimensions

### Related :handshake:
**[VEN-462](https://helsinkisolutionoffice.atlassian.net/browse/VEN-462):** Add number of total places and free places

### Automated tests :gear:️
```shell
$ pytest resources/tests/test_resources_queries.py::test_get_harbors
$ pytest resources/tests/test_resources_queries.py::test_create_harbor
$ pytest resources/tests/test_resources_queries.py::test_upate_harbor
```

**Admin**
1. Go to the `Harbor` page for any harbor: [`http://localhost:8000/admin/resources/harbor/70481d5c-9264-4c81-87af-e5cf7fb4efa0/change/`](http://localhost:8000/admin/resources/harbor/70481d5c-9264-4c81-87af-e5cf7fb4efa0/change/)
2. Scroll to the bottom, you should see the dynamic fields as read-only fields (go to `Screeenshots` section for details)

## Screenshots :camera_flash:
**Harbor admin page:**
image

**Create Harbor input does not have a `number_of_place` field anymore:**
image

`Winter` resources still don't have mutations, so those changes don't apply to those.


README.md
```diff
@@ -106,3 +107,11 @@ You can also import default images for Helsinki harbors / winter storage areas:
...
 
     ./manage.py loaddata switch-reasons.json
+
+Lastly, you can load the User Groups:
+
+    ./manage.py loaddata groups.json
+
+And install the model permissions:
+
+    ./manage.py set_group_model_permissions
```
igordavydsson, line 117: yes, very welcome addition 👍

resources/models.py
```diff
@@ -101,6 +129,8 @@ class Meta:
 
 
 class Harbor(AbstractArea, TranslatableModel):
+    objects = HarborManager()
```
igordavydsson, line 146: convention is to declare this after all the fields and before the `class Meta`

resources/models.py
```diff
@@ -164,6 +185,8 @@ def __str__(self):
 
 
 class WinterStorageArea(AbstractArea, TranslatableModel):
+    objects = WinterStorageAreaManager()
```
igordavydsson, line 198: same as above

resources/schema.py
```diff
@@ -233,28 +233,38 @@ class Meta:
...
         interfaces = (relay.Node,)
+        filterset_class = HarborFilter
 
     name = graphene.String()
     street_address = graphene.String()
     municipality = graphene.String()
     image_file = graphene.String()
     maps = graphene.List(HarborMapType, required=True)
+    max_width = graphene.Float()
+    max_length = graphene.Float()
```
igordavydsson, line 266: we could mark these as `required=True`

resources/schema.py
```diff
@@ -311,28 +321,37 @@ class Meta:
...
         interfaces = (relay.Node,)
+        filterset_class = WinterStorageAreaFilter
 
     name = graphene.String()
     street_address = graphene.String()
     municipality = graphene.String()
     image_file = graphene.String()
     maps = graphene.List(WinterStorageAreaMapType, required=True)
+    max_width = graphene.Float()
+    max_length = graphene.Float()
```
igordavydsson, line 355: these are okay like this, since not all WSAs have marked places 👍

resources/admin.py
```diff
@@ -35,17 +38,57 @@ class BoatTypeAdmin(TranslatableAdmin):
...
+        return obj.max_width
+
+    def max_length(self, obj):
+        return obj.max_length
+
+    def max_depth(self, obj):
+        return obj.max_depth
+
+    max_width.short_description = _("Maximum allowed width")
+    max_width.admin_order_field = _("max_width")
```
igordavydsson, line 66: probably `admin_order_field` attributes do not need lazy translation wrappers?

resources/admin.py
```diff
@@ -35,17 +38,57 @@ class BoatTypeAdmin(TranslatableAdmin):
...
+
+    max_length.short_description = _("Maximum allowed length")
```
igordavydsson, line 68: nitpick, but could you put these right after the property declaration and without empty lines in between, i.e.
```python
def max_width(self, obj):
    return obj.max_width
max_width.short_description = _("Maximum allowed width")
max_width.admin_order_field = "max_width"

def max_length(self, obj):
    return obj.max_length
max_length.short_description = _("Maximum allowed length")
max_length.admin_order_field = "max_length"

def max_depth(self, obj):
    return obj.max_depth
max_depth.short_description = _("Maximum allowed depth")
max_depth.admin_order_field = "max_depth"
```
this is the convention that I've found in many projects and which seems easier to navigate 🙂

resources/schema.py
```diff
@@ -265,6 +265,7 @@ class Meta:
     max_width = graphene.Float()
     max_length = graphene.Float()
     max_depth = graphene.Float()
+    number_of_places = graphene.Int()
```
igordavydsson, line 268: this will be `required=True` too, since the `Count` annotation will at least return `0`

resources/schema.py
```diff
@@ -352,6 +353,7 @@ class Meta:
     maps = graphene.List(WinterStorageAreaMapType, required=True)
     max_width = graphene.Float()
     max_length = graphene.Float()
+    number_of_marked_places = graphene.Int()
```
igordavydsson, line 356: same

resources/schema.py
igordavydsson: OTOH, hm, what if a harbor has no berths. like if we just created it. will the `Max` annotations return `null`s? if so, maybe it's not a bad idea to have these as just `graphene.Float()` 🤔

resources/admin.py
```diff
@@ -35,17 +38,71 @@ class BoatTypeAdmin(TranslatableAdmin):
     pass
 
 
-class HarborAdmin(CustomTranslatableAdmin, admin.OSMGeoAdmin):
-    list_display = ("name",)
+class HarborAdmin(CustomTranslatableAdmin, admin.OSMGeoAdmin, HarborManager):
```
igordavydsson, line 41: why do you need the `HarborManager` here? Django will do `Harbor.objects.all()` or `Harbor.objects.get(pk=obj.id)`, which will already have all the annotations. 🤔

resources/admin.py
```diff
@@ -35,17 +38,71 @@ class BoatTypeAdmin(TranslatableAdmin):
...
+    max_depth.admin_order_field = _("max_depth")
 
 
 class PierAdmin(admin.OSMGeoAdmin):
     filter_horizontal = ("suitable_boat_types",)
 
 
-class WinterStorageAreaAdmin(CustomTranslatableAdmin, admin.OSMGeoAdmin):
+class WinterStorageAreaAdmin(
+    CustomTranslatableAdmin, admin.OSMGeoAdmin, WinterStorageAreaManager
```
igordavydsson, line 80: same question about the manager

resources/models.py
a-rmz: By `this` you mean the `objects/manager`? If so, where can I find more info about it? Since all the model fields are defined before the `class Meta`.

igordavydsson: yes, I mean `objects = HarborManager()` usually goes after the model fields and before the `class Meta`. [here in the first example](https://docs.djangoproject.com/en/3.0/topics/db/managers/#custom-managers-and-model-inheritance) it is implied and [here in this section and following sections](https://docs.djangoproject.com/en/3.0/topics/db/managers/#custom-managers) it is done exactly like that.

resources/schema.py
a-rmz: I think it's better to keep it as `required=False` unless we want to return `0` instead of `null`. 🤔

resources/admin.py
a-rmz: Whoops 😅 I got carried away by the copy-paste. Good catch.

resources/schema.py
igordavydsson: yeah, let's keep it as is 👍 I think it makes more sense if we return `null` when there are no berths in a harbor

resources/models.py
igordavydsson: here are some examples from wagtail's repo:
https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L73
https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L1594
https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L2187

a-rmz: Gotcha 👍 good to know, thanks. 😄

resources/admin.py
a-rmz: Black automatically adds a whitespace in between tho.

igordavydsson: yeah, thought so 😅 that's ok 👍

resources/admin.py
a-rmz: It was part of my experiments trying to get the new annotated fields to the admin site, gone now. 👍

PR #188, opened 2020-04-06
VEN-537 | Mark berths as reserved

a-rmz:
### `renew automatically`
- Add a field to allow `Berth`s to be renewed automatically (to be implemented) for the following season
- Add an `is_available` field (annotated) to the `Berth` model (and tests)
- Add the field to `BerthAdmin`

### `number of places`
- Automatically calculate `number of places` and `number of free places` depending on the number of available places per section (`pier` or `wssection`)
- Move the dimension annotations (`max width`, `max length`, and `max depth`) to the `Section` level (`pier` or `wssection`), and expose them on both levels (`section` and `area`) on GraphQL

### `lease validation`
- Add validation for `Lease` dates:
  - A lease `start date` cannot be _after_ the `end date`
  - A lease cannot be assigned to a place (`berth` or `wsplace`) if there is a valid lease that ends _AFTER_ the start date of the new lease
- Minor tweaks to project organisation

### Closes :no_good_woman:
**[VEN-537](https://helsinkisolutionoffice.atlassian.net/browse/VEN-537):** Mark berths as reserved if they have lease assigned

### Related :handshake:
**[VEN-462](https://helsinkisolutionoffice.atlassian.net/browse/VEN-462):** Add number of total places and free places

### Automated tests :gear:️
```shell
$ pytest
```

#### `renew automatically`
1. Create a lease and assign it to a berth
2. Go to the berth detail page and look for the `Berth available` property

#### `number of places`
```graphql
query ALL_HARBORS {
  harbors {
    edges {
      node {
        id
        properties {
          maxWidth
          maxLength
          maxDepth
          piers {
            edges {
              node {
                properties {
                  maxWidth
                  maxLength
                  maxDepth
                  numberOfPlaces
                  numberOfFreePlaces
                  berths {
                    edges {
                      node {
                        id
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}
```

## Screenshots :camera_flash:
image


resources/models.py
```diff
@@ -451,6 +453,45 @@ class Meta:
...
+        paid_status = Q(status=LeaseStatus.PAID)
+        auto_renew = Q(renew_automatically=True)
+
+        active_leases = (
+            BerthLease.objects.filter(berth=OuterRef("pk"))
+            .filter(
+                Q(in_current_season & active_current_status)
+                | Q(in_last_season & auto_renew & paid_status)
+            )
+            .values("id")
```
igordavydsson, line 535: hm, I think this is not needed, since it fires the `EXISTS NOT ... ` query, which does not care about the values. at least, when tested locally, I could not see any difference, if I tested it correctly of course

resources/models.py
```diff
@@ -451,6 +453,45 @@ class Meta:
...
+                    * It needs to have a "valid" status (DRAFTED, OFFERED, PAID)
+                + If a lease ended during the last season:
+                    * It needs to have a "valid" status (PAID)
+                    * It needs to have renew automatically set
+        """
+        from leases.models import BerthLease
+
+        current_season_end = calculate_berth_lease_end_date()
+
+        in_current_season = Q(end_date=current_season_end)
```
igordavydsson, line 519: should we maybe check for the current season using the start_date? what if the lease starts on the season opening day but is only for 2 months and not till the end of the season? probably:
1. today is before season opening day -> just use calculate start date
2. today is after season start date -> start date after today

resources/models.py
a-rmz: Yeah, it doesn't seem to have any performance improvement. However, it simplifies a bit the SQL generated by that QuerySet, removing some unncessary fields. We can remove it anyway since it's no affecting direclty and we (thankfully) don't get to see the raw SQL often.

```sql
NOT EXISTS(SELECT
        U0."id",
        U0."created_at",
        U0."modified_at",
        U0."customer_id",
        U0."boat_id",
        U0."status",
        U0."comment",
        U0."berth_id",
        U0."application_id",
        U0."start_date",
        U0."end_date",
        U0."renew_automatically" 
    FROM
        "leases_berthlease" U0 
    WHERE
        (
            U0."berth_id" = (
                "resources_berth"."id"
            ) 
            AND (
                (
                    U0."end_date" = 2020-09-14 
                    AND U0."status" IN (
                        drafted, offered, paid
                    )
                ) 
                OR (
                    U0."end_date" = 2019-09-14 
                    AND U0."renew_automatically" = True 
                    AND U0."status" = paid
                )
            )
        )) AS "is_available"
```

```sql
NOT EXISTS(SELECT
        U0."id" 
    FROM
        "leases_berthlease" U0 
    WHERE
        (
            U0."berth_id" = (
                "resources_berth"."id"
            ) 
            AND (
                (
                    U0."end_date" = 2020-09-14 
                    AND U0."status" IN (
                        drafted, offered, paid
                    )
                ) 
                OR (
                    U0."end_date" = 2019-09-14 
                    AND U0."renew_automatically" = True 
                    AND U0."status" = paid
                )
            )
        )) AS "is_available"
```

igordavydsson: I think at the end of the day postgres does not care for columns to be selected in the `SELECT` subquery. Deleting `.values("id")` however simplifies our python code (although just by a bit 😅)

resources/schema.py
```diff
@@ -151,6 +151,12 @@ class Meta:
 
 
 class PierNode(graphql_geojson.GeoJSONType):
+    number_of_free_places = graphene.Int(required=True)
+    number_of_places = graphene.Int(required=True)
+    max_width = graphene.Float()
+    max_length = graphene.Float()
```
igordavydsson, line 157: I think this two are non-nullable non-blankable, so there will always be values

igordavydsson: ah, but if there are no places/berths... right

a-rmz: Yeah, if there are no places, it will return `null`. 👍

resources/models.py
igordavydsson: as discussed on slack: this is not actually the logic we want. no changes needed here.

leases/models.py
```diff
@@ -90,6 +93,15 @@ def clean(self):
             raise ValidationError(
                 _("BerthLease start and end year have to be the same")
             )
+        existing_leases = BerthLease.objects.filter(
```
igordavydsson, line 96: can we rename this to `active_leases_for_given_period` or something?

igordavydsson: maybe just `leases_for_given_period`

resources/tests/test_resources_queries.py
```diff
@@ -53,6 +55,7 @@ def test_get_harbors(api_client, berth):
         }
     """
     executed = api_client.execute(query)
+    print(executed)
```
igordavydsson, line 58: 👀

a-rmz: :derp:

PR #192, opened 2020-04-08
VEN-558 | Validate that BerthLease season occurs during the same year

a-rmz:
## Description :sparkles:
- Add a validation for `BerthLease` years

### Closes :no_good_woman:
**[VEN-558](https://helsinkisolutionoffice.atlassian.net/browse/VEN-558):** Validate that BerthLease season occurs during the same year

### Related :handshake:
#188 

### Automated tests :gear:️
```shell
$  pytest leases/tests/test_lease_models.py::test_lease_should_be_the_same_year
```

## Screenshots 📸 
**Error returned on GraphQL**
image

### DIFF
--- a/leases/models.py
+++ b/leases/models.py
@@ -115,6 +115,13 @@ class BerthLease(AbstractLease):
         verbose_name_plural = _("berth leases")
         default_related_name = "berth_leases"
 
+    def clean(self):
+        if self.start_date.year != self.end_date.year:
+            raise ValidationError(
+                _("BerthLease start and end year have to be the same")
+            )
+        super().clean()
+
     def __str__(self):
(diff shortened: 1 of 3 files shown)


leases/models.py
```diff
@@ -115,6 +115,13 @@ class Meta:
         verbose_name_plural = _("berth leases")
         default_related_name = "berth_leases"
 
+    def save(self, *args, **kwargs):
+        if self.start_date.year != self.end_date.year:
```
igordavydsson, line 119: we should maybe do it in the `clean()` method? and call `self.full_clean()` from `save()`. like we do [here](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/applications/models.py#L256)

PR #199, opened 2020-04-21
VEN-548 | Update Lease mutation

a-rmz:
## Description :sparkles:
- Add extra validation for leases changing application
  - Check that they belong to the same owner
  - WSApplication doesn't have a `customer` field, so it can't be checked atm
- Add the `UpdateBerthLeaseMutation`
- Create a separate `BerthLeaseInput` class to share the common input fields between mutations
- Minor changes to `CreateBerthLeaseMutation` to fail faster when the application has no customer

### Closes :no_good_woman:
**[VEN-548](https://helsinkisolutionoffice.atlassian.net/browse/VEN-548):** 

### Automated tests :gear:️
```shell
```

```graphql
mutation UpdateBerthLease($input: UpdateBerthLeaseMutationInput!) {
    updateBerthLease(input:$input){
        berthLease {
            id
            startDate
            endDate
            comment
            boat {
                id
            }
            application {
                id
                customer {
                    id
                }
            }
        }
    }
}
```

```json
{
  "id": "<BerthLeaseID>",
  "startDate": "2019-01-01",
  "endDate": "2020-09-04",
  "comment": "Such updated, very wow",
  "boatId": "<OPTIONAL>",
  "applicationId": "<OPTIONAL>",
}
```

## Screenshots :camera_flash:
**GraphQL schema**
image

### DIFF
--- a/leases/tests/test_lease_mutations.py
+++ b/leases/tests/test_lease_mutations.py
@@ -2,6 +2,7 @@ import uuid
 from random import randint
 
 import pytest
+from dateutil.utils import today
 from freezegun import freeze_time
 from graphql_relay import to_global_id
 
@@ -167,7 +168,11 @@ def test_create_berth_lease_application_doesnt_exist(superuser_api_client, berth
     assert_doesnt_exist("BerthApplication", executed)
 
 
-def test_create_berth_lease_berth_doesnt_exist(superuser_api_client, berth_application):
+def test_create_berth_lease_berth_doesnt_exist(
+    superuser_api_client, berth_application, customer_profile
+):
+    berth_application.customer = customer_profile
+    berth_application.save()
     variables = {
         "applicationId": to_global_id(
             BerthApplicationNode._meta.name, berth_application.id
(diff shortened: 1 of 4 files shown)


leases/tests/test_lease_mutations.py
```diff
@@ -322,3 +327,154 @@ def test_delete_berth_lease_inexistent_lease(superuser_api_client):
...
+    }
+
+    executed = api_client.execute(UPDATE_BERTH_LEASE_MUTATION, input=variables)
+    assert executed["data"]["updateBerthLease"]["berthLease"] == {
+        "id": berth_lease_id,
+        "startDate": str(berth_lease.start_date),
+        "endDate": str(berth_lease.end_date),
+        "comment": berth_lease.comment,
+        "boat": {"id": boat_id},
+        "application": None,
```
igordavydsson, line 423: hm, yeah, but also `berth_lease` fixture does not have application from the start. so, what you are testing is not removing, but the fact, that `application` is non-required for `updateBerthLease` mutation.

a-rmz: Good catch, fixing. 👍

igordavydsson: nice 👍

PR #204, opened 2020-04-28
VEN-585 | CountConnection

a-rmz:
## Description :sparkles:
* Add `CountConnection` to expose the entity count on GQL connection fields
* Add tests for the connections using it

### Closes :no_good_woman:
**[VEN-585](https://helsinkisolutionoffice.atlassian.net/browse/VEN-585):** Add custom connection with count attribute

### Automated tests :gear:️
```shell
$ pytest
```

All the objects exposed as connections extend this class, so you can test with any resource
```graphql
query CONNECTION_TEST {
  allHarbors:harbors {
    count
    totalCount
  }
  harborsElectricity:harbors(piers_Electricity: true) {
    count
    totalCount
  }
  harborsNoElectricity:harbors(piers_Electricity: false) {
    count
    totalCount
  }
}
```
**Result:**
```json
{
  "data": {
    "allHarbors": {
      "count": 42,
      "totalCount": 42
    },
    "harborsElectricity": {
      "count": 28,
      "totalCount": 42
    },
    "harborsNoElectricity": {
      "count": 16,
      "totalCount": 42
    }
  }
}
```

## Screenshots :camera_flash:
image

Some of the `Node`s don't expose a query, so they cannot be tested. However, the base `connection_class` was added for consistency among nodes.

### DIFF
--- a/resources/tests/test_resources_queries.py
+++ b/resources/tests/test_resources_queries.py
@@ -1,4 +1,5 @@
 import json
+import random
 
 import pytest
 from graphql_relay import to_global_id
@@ -11,7 +12,15 @@ from berth_reservations.tests.utils import (
 from leases.schema import BerthLeaseNode
 from leases.tests.factories import BerthLeaseFactory
 from resources.schema import BerthNode, HarborNode, PierNode
-from resources.tests.factories import BerthFactory
+from resources.tests.factories import (
+    BerthFactory,
+    BerthTypeFactory,
+    HarborFactory,
+    PierFactory,
+    WinterStorageAreaFactory,
+    WinterStoragePlaceFactory,
+    WinterStorageSectionFactory,
+)
 
 
 def test_get_boat_type(api_client, boat_type):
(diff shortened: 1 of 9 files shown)


utils/schema.py
```diff
@@ -0,0 +1,21 @@
...
+    total_count = graphene.Int(
+        description="Total count of nodes on this connection regardless of filters",
+        required=True,
+    )
+
+    def resolve_count(self, info):
+        return self.iterable.distinct().count()
+
+    def resolve_total_count(self, info, **kwargs):
+        return self.iterable.model.objects.distinct().count()
```
igordavydsson, line 21: does it need to be `distinct` here? since it just implies `objects.count()` I don;t think it should have any duplicates. 🤔 or did you get duplicates here?

resources/tests/test_resources_queries.py
```diff
@@ -562,3 +571,277 @@ def test_get_harbor_available_active_berths(api_client, pier, available):
...
+        }
+    """
+
+    executed = api_client.execute(query)
+    assert executed["data"] == {"harbors": {"count": count, "totalCount": count}}
+
+
+def test_get_harbor_count_filtered(api_client):
+    filtered = random.randint(1, 10)
+    unfiltered = random.randint(1, 10)
```
igordavydsson, line 596: this naming is a bit misleading IMO. `unfiltered` is what you get for `totalCount`, both of these will be `filtered` results, but one `withElectricityInPiers` while the other one is `withoutElectricityInPiers` 🤔 same applies for other test cases too

resources/tests/test_resources_queries.py
```diff
@@ -562,3 +571,277 @@ def test_get_harbor_available_active_berths(api_client, pier, available):
...
+    assert executed["data"] == {"berths": {"count": count, "totalCount": count}}
+
+
+def test_get_berth_count_filtered(api_client):
+    filtered = random.randint(1, 10)
+    unfiltered = random.randint(1, 10)
+    filtered_bt = BerthTypeFactory(
+        width=round(random.uniform(0.1, 4.99), 2),
+        length=round(random.uniform(0.1, 4.99), 2),
+        depth=round(random.uniform(0.1, 4.99), 2),
```
igordavydsson, line 694: you could have skipped `length` and `depth`, if you are only filtering by `width` 🙂  would make the test shorter

utils/schema.py
a-rmz: The tests seem to work fine without the `distinct`, so we can get rid of this. 👍

resources/tests/test_resources_queries.py
a-rmz: Yeah, I was having the same concern but couldn't come up with a clearer naming. I'll think about it and suggestions are also welcome. 😅

resources/tests/test_resources_queries.py
a-rmz: Terrrrrific catch. 👍

resources/tests/test_resources_queries.py
igordavydsson: Great job with the names!

a-rmz:
Thanks Haymitch. :D

PR #207, opened 2020-04-29
VEN-591 | Hide BerthType from client

a-rmz:
## Description :sparkles:
Hide the `BerthType` logic from the client by exposing the dimensions + mooring type directly on the `BerthNode` rather than having a nested field. When creating or updating a `berth`, only the dimensions and mooring type are expected, not the `berth_type_id` anymore.

When a berth is being created, a `BerthType` with the passed properties will try to be fetched from the DB, if it doesn't exist, a new one will be created and immediately associated with the new `berth`.

When a berth is being updated, a `BerthType` will try to be fetched _with the new properties_ combined with the old ones for the props not provided, if there's no matching `type`, a new one will be created and associated.

### Closes :no_good_woman:
**[VEN-591](https://helsinkisolutionoffice.atlassian.net/browse/VEN-591):** Hide BerthType from client

### Automated tests :gear:️
```shell
$ pytest resources/tests/test_resources_mutations.py
```

**Query:**
```graphql
query BERTHS {
  berths(first: 5) {
    edges {
      node {
        id
        width
        length
        depth
        mooringType
      }
    }
  }
}
```

**Create:**
```graphql
mutation CREATE_BERTH($input: CreateBerthMutationInput!) {
    createBerth(input: $input) {
        berth {
            id
            width
            length
            depth
            mooringType
        }
    }
}
```
Input:
```json
{
    "number": 666,
    "pierId": "<PIER_ID>",
    "width": 3.33,
    "length": 11.1,
    "depth": 5.55,
    "mooringType": "STERN_BUOY_PLACE"
}
```

**Update:**
```graphql
mutation UPDATE_BERTH($input: UpdateBerthMutationInput!) {
    updateBerth(input: $input) {
        berth {
            id
            width
            length
            depth
            mooringType
        }
    }
}
```
Input:
```json
{
    "id": "<BERTH_ID>",
    "width": 6.66
}
```

## Screenshots :camera_flash:
**Berth node exposes the dimensions directly:**
image

**Create berth mutation input (no more berth_type_id):**
image

**Update berth mutation input (no more berth_type_id):**
image

### DIFF
--- a/resources/schema.py
+++ b/resources/schema.py
@@ -40,7 +40,6 @@ from .models import (
     WinterStorageArea,
     WinterStorageAreaMap,
     WinterStoragePlace,
-    WinterStoragePlaceType,
     WinterStorageSection,
 )
 
@@ -174,17 +173,6 @@ class PierNode(graphql_geojson.GeoJSONType):
         connection_class = CountConnection
 
 
-class BerthTypeNode(DjangoObjectType):
-    class Meta:
-        model = BerthType
-        interfaces = (relay.Node,)
-
-    width = graphene.Float(description=_("width (m)"), required=True)
-    length = graphene.Float(description=_("length (m)"), required=True)
-    depth = graphene.Float(description=_("depth (m)"))
-    mooring_type = BerthMooringTypeEnum(required=True)
-
-
 class BerthNodeFilterSet(django_filters.FilterSet):
     min_width = django_filters.NumberFilter(
         field_name="berth_type__width", lookup_expr="gte"
@@ -206,6 +194,10 @@ class BerthNode(DjangoObjectType):
     )
     is_accessible = graphene.Boolean()
     is_available = graphene.Boolean(required=True)
+    width = graphene.Float(description=_("width (m)"), required=True)
+    length = graphene.Float(description=_("length (m)"), required=True)
+    depth = graphene.Float(description=_("depth (m)"))
+    mooring_type = BerthMooringTypeEnum(required=True)
 
     class Meta:
         model = Berth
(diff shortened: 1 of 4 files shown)


resources/schema.py
```diff
@@ -446,41 +447,12 @@ class BerthInput(AbstractBoatPlaceInput):
...
-            )
-        else:
-            raise VenepaikkaGraphQLError(_("Missing fields to associate a BerthType"))
-
-    return berth_type
-
-
 class CreateBerthMutation(graphene.ClientIDMutation):
     class Input(BerthInput):
         number = graphene.Int(required=True)
```
igordavydsson, line 458: `CreateBerthMutationInput` should have `width`, `length` and `mooringType` as `required=True`

resources/schema.py
```diff
@@ -492,7 +464,21 @@ class Input(BerthInput):
     @add_permission_required(Berth)
     @transaction.atomic
     def mutate_and_get_payload(cls, root, info, **input):
-        input["berth_type"] = get_berth_type(info, input)
+        width = input.pop("width", None)
+        length = input.pop("length", None)
+        depth = input.pop("depth", None)
+        mooring_type = input.pop("mooring_type", None)
+
+        if all([width, length, mooring_type]):
```
igordavydsson, line 472: I guess this `if` won't be necessary, since dimensions and mooring type should be required (except for `depth` of course)

resources/schema.py
a-rmz: Good catch! This also fixes the other comment.

image

PR #210, opened 2020-05-15
VEN-187 | Payments

a-rmz:
## Description :sparkles:
* Add first version of payments app

### Closes :no_good_woman:
**[VEN-187](https://helsinkisolutionoffice.atlassian.net/browse/VEN-187):** Add pricing related data models

### Automated tests :gear:️
```shell
```

http://localhost:8000/admin/payments/

**Add a Berth Price Group:**
- Create the new group
http://localhost:8000/admin/payments/berthpricegroup/

- Go to a `BerthType`
http://localhost:8000/admin/resources/berthtype/3ebb3feb-dd5e-4480-817a-d2dcfdda52c1/change/

- Specify the price group
- Go back to the created Price Group

**Add a Berth Product:**
http://localhost:8000/admin/payments/berthproduct/add/

## Screenshots :camera_flash:
**Berth Price Group**
image

**Berth Product**
image

**Additional Product**
image


payments/admin.py
```diff
@@ -0,0 +1,34 @@
...
+    BerthPriceGroup,
+    BerthProduct,
+    WinterStorageProduct,
+)
+
+
+class BerthPriceGroupAdmin(admin.ModelAdmin):
+    readonly_fields = ("berth_types",)
+
+    def berth_types(self, obj):
```
igordavydsson, line 15: can this be `filter_horizontal` instead of the readonly field?

a-rmz: Don't think it will be possible
```
<class 'payments.admin.BerthPriceGroupAdmin'>: (admin.E020) The value of 'filter_horizontal[0]' must be a many-to-many field.
```

payments/models.py
```diff
@@ -0,0 +1,146 @@
...
+                name="unique_product_for_harbor_pricegroup",
+            ),
+        ]
+
+    def __str__(self):
+        return f"{self.price_group} - {self.harbor} ({self.price_value}€)"
+
+
+class WinterStorageProduct(AbstractPlaceProduct):
+    price_unit = EnumField(
```
igordavydsson, line 96: this field is probably handled by `AbstractPlaceProduct`?

a-rmz: Right! I forgot to remove it from here. Good catch 👍

payments/models.py
```diff
@@ -0,0 +1,146 @@
...
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+
+class BerthPriceGroup(UUIDModel):
+    name = models.CharField(
+        verbose_name=_("berth price group name"), max_length=128, null=True, blank=True
```
igordavydsson, line 44: let's make it non-blankable and non0nullable, since it's the only real field here

payments/admin.py
igordavydsson: yeah... maybe we should try some inline here? I would like to have some kind of selection interface here to avoid setting it from the berth type page.

igordavydsson: I mean, you should also be able to set it on the berth type page, but it is easier to just choose several berth types from this one page

payments/models.py
```diff
@@ -0,0 +1,139 @@
...
+    class Meta:
+        abstract = True
+
+
+class BerthProduct(AbstractPlaceProduct):
+    price_group = models.ForeignKey(
+        BerthPriceGroup,
+        on_delete=models.CASCADE,
+        verbose_name=_("price group"),
+        related_name=_("products"),
```
igordavydsson, line 72: should not be translatable 👀

payments/models.py
```diff
@@ -0,0 +1,139 @@
...
+    )
+    harbor = models.ForeignKey(
+        "resources.Harbor",
+        verbose_name=_("harbor"),
+        related_name=_("product"),
```
igordavydsson, line 77: should not be translatable 👀

payments/models.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def __str__(self):
+        return f"{self.price_group} - {self.harbor} ({self.price_value}€)"
+
+
+class WinterStorageProduct(AbstractPlaceProduct):
+    winter_storage_area = models.ForeignKey(
+        "resources.WinterStorageArea",
+        verbose_name=_("winter storage area"),
+        related_name=_("product"),
```
igordavydsson, line 97: should not be translatable 👀

payments/models.py
```diff
@@ -0,0 +1,139 @@
...
+                name="unique_product_for_harbor_pricegroup",
+            ),
+        ]
+
+    def __str__(self):
+        return f"{self.price_group} - {self.harbor} ({self.price_value}€)"
+
+
+class WinterStorageProduct(AbstractPlaceProduct):
+    winter_storage_area = models.ForeignKey(
```
igordavydsson, line 94: shouldn't it be `OneToOneField`?

a-rmz: Right! I just re-read the specs. I understood that it should behave the same way as BerthProduct, but if the price will always be the same for any space on the WSArea, then it makes sense that it's a `OneToOne`.

payments/admin.py
a-rmz: Yeah, I thought about inlining as well. The problem is that it only lets you create new BerthTypes, not select existing ones, but I can dig a bit deeper to try and find something.

payments/models.py
igordavydsson: yeah, for now one WSArea always has one price per square meter and that is what the product is going to be.

payments/models.py
```diff
@@ -0,0 +1,139 @@
...
+    price_group = models.ForeignKey(
+        BerthPriceGroup,
+        on_delete=models.CASCADE,
+        verbose_name=_("price group"),
+        related_name="products",
+    )
+    harbor = models.ForeignKey(
+        "resources.Harbor",
+        verbose_name=_("harbor"),
+        related_name="product",
```
igordavydsson, line 77: I guess, one harbor can have multiple `productS`

payments/models.py
```diff
@@ -0,0 +1,139 @@
...
+    service = EnumField(ServiceType, verbose_name=_("service"), max_length=40)
+    period = EnumField(PeriodType, max_length=8)
+
+    @property
+    def product_type(self):
+        if self.service.is_fixed_service():
+            return AdditionalProductType.FIXED_SERVICE
+        elif self.service.is_optional_service():
+            return AdditionalProductType.OPTIONAL_SERVICE
+        return None
```
igordavydsson, line 115: hmhmhm, not a fan of returning `None`, I would rather do
```python
if self.service.is_fixed_service():
    return AdditionalProductType.FIXED_SERVICE
else:
    return AdditionalProductType.OPTIONAL_SERVICE
```
'cause at. the API we. would need to have a filter for this and would be nice in `product_type` could be only either ... or ... .

a-rmz: Yes, I agree, and if everything is done right (which it should), it _won't ever_ fall on the `None` case, except when creating a new instance from the Admin.

If you have `else OPTIONAL_SERVICE`, the admin will display `Product type: Optional Service` when creating a new instance, which can be missleading. 

image

igordavydsson: aahhh, okay, yeah, let's keep it that way

payments/tests/test_payments_models.py
```diff
@@ -0,0 +1,105 @@
...
+    errors = str(exception.value)
+    assert "tax_percentage" in errors
+    assert "is not a valid choice" in errors
+
+
+def test_berth_price_group_delete_berth_type_cascade(berth_price_group):
+    berth_price_group.berth_types.first().delete()
+
+    with pytest.raises(BerthProduct.DoesNotExist):
+        BerthProduct.objects.get(id=berth_price_group.id)
```
igordavydsson, line 40:
```python
BerthProduct.objects.get(price_group_id=berth_price_group.id)
```
?

payments/tests/test_payments_models.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def test_berth_product_delete_berth_type_product_group_is_none(berth_price_group):
+    berth_type = berth_price_group.berth_types.first()
+    BerthPriceGroup.objects.get(id=berth_price_group.id).delete()
+
+    assert BerthType.objects.get(id=berth_type.id).price_group is None
```
igordavydsson, line 47: are you testing that nullable FK with `SET_NULL` works properly? I think Django already covers a test like this 🙂

a-rmz: Yup 😅 double checking that it's actually deleted, but I can remove it if it's redundant. :)

payments/tests/test_payments_models.py
a-rmz: :derp:

payments/tests/test_payments_models.py
igordavydsson: I think we can trust Django on this one

a-rmz: Alright! In Django we trust. 🙏

payments/tests/test_payments_models.py
```diff
@@ -0,0 +1,90 @@
...
+        AdditionalProductFactory(service=service, period=period)
+
+    errors = str(exception.value)
+    assert (
+        'duplicate key value violates unique constraint "optional_services_per_period"'
+        in errors
+    )
+
+
+def test_additional_product_fixed_service_tax_value():
```
igordavydsson, line 81: you are testing `optional` services, not `fixed` ones

payments/migrations/0001_initial.py
```diff
@@ -0,0 +1,266 @@
...
+                        decimal_places=2,
+                        default=Decimal("24.0"),
+                        max_digits=5,
+                        verbose_name="tax percentage",
+                    ),
+                ),
+                (
+                    "price_unit",
+                    enumfields.fields.EnumField(
+                        default="amount", enum=payments.enums.PriceUnits, max_length=10
```
igordavydsson, line 158: hm, `django-enumfields` does not pass `choices` to this autogenerated migrations file?

payments/tests/test_payments_models.py
a-rmz: Right! It should actually be testing `FIXED_SERVICES`

>fixed_service products should always be VAT 24%

payments/migrations/0001_initial.py
a-rmz: It's strange: it doesn't pass it, but it actually only allows the right choices. Maybe it's handled directly at the Django level and not DB?

igordavydsson: probably, yeah

payments/tests/test_payments_models.py
```diff
@@ -0,0 +1,92 @@
...
+    with pytest.raises(ValidationError) as exception:
+        AdditionalProductFactory(
+            service=random.choice(ServiceType.FIXED_SERVICES()),
+            tax_percentage=Decimal("999.9"),
```
igordavydsson, line 87: maybe check that it can't be even `Decimal("10.0")`?

payments/models.py
```diff
@@ -0,0 +1,148 @@
...
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+
+class BerthPriceGroupManager(models.Manager):
+    def get_or_create_for_width(self, width):
```
igordavydsson, line 43: oh, neat 👍

payments/models.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class BerthProduct(AbstractPlaceProduct):
+    price_group = models.ForeignKey(
+        BerthPriceGroup,
+        on_delete=models.CASCADE,
+        verbose_name=_("price group"),
+        related_name="products",
+    )
+    harbor = models.ForeignKey(
```
igordavydsson, line 83: yeah, this should be nullable :derp:

igordavydsson: if it's null, then it is the default price for this price group in all harbors

payments/admin.py
igordavydsson: yes, let's keep it as read-only for now, since we are assigning / generating BPGs automatically for now

a-rmz: 🆒 🆒 🆒

PR #211, opened 2020-05-20
[VEN-396] Payments API

a-rmz:
## Description :sparkles:
Add a first version for the Payments API
* Add queries and mutations for: `BerthProduct`, `WinterStorageProduct`, and `AdditionalProduct`
* Add queries for `BerthPriceGroups`
* Add model permissions for the `payments` models
* Split the tax percentages into two different lists for easier handling
* Move some of the existing util functions to separate files and add new ones

### Closes :no_good_woman:
**[VEN-396](https://helsinkisolutionoffice.atlassian.net/browse/VEN-396):** Expose pricing data models via GraphQL API

### Related :handshake:
**#210**: VEN-187 | Payments

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_queries.py
```

To avoid cluttering the PR, I'm not including all the queries and mutations, examples can be found on the respective test files (`test_payments_queries` or `test_payments_mutations`).

## Screenshots :camera_flash:
image

Once the PR is merged, the command to assign the permissions has to be run
```shell
python manage.py set_group_model_permissions
```


payments/tests/factories.py
```diff
@@ -31,16 +26,9 @@ class Meta:
 
 
 class BerthPriceGroupFactory(factory.django.DjangoModelFactory):
-    name = factory.Faker("word")
```
igordavydsson: as we have discussed, you should probably leave this line in this Factory, otherwise it won't work

payments/tests/conftest.py
```diff
@@ -1,17 +1,22 @@
...
     WinterStorageProductFactory,
 )
+from resources.tests.conftest import *  # noqa
+from resources.tests.factories import BerthTypeFactory
 
 
 @pytest.fixture
 def berth_price_group():
-    berth_price_group = BerthPriceGroupFactory()
+    width = randgen.uniform(1, 999)
```
igordavydsson, line 16: some comment why we are doing it this way would be welcome here for future generations 🙂

payments/schema/types.py
```diff
@@ -0,0 +1,61 @@
...
+        interfaces = (graphene.relay.Node,)
+        connection_class = CountConnection
+
+    @classmethod
+    @view_permission_required(BerthPriceGroup)
+    def get_queryset(cls, queryset, info):
+        return super().get_queryset(queryset, info)
+
+    def resolve_products(self, info):
+        return self.products.all()
```
igordavydsson, line 62: if this is all this resolver does, it is kinda redundant, as this is the default logic anyway

a-rmz: Seems to work 👍 it was probably a (functioning) leftover from some previous tests. Gone now.

payments/tests/test_payments_queries.py
```diff
@@ -0,0 +1,255 @@
...
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_supervisor", "berth_handler", "berth_services"],
+    indirect=True,
+)
+def test_get_berth_price_groups(api_client, berth_price_group):
+    product = BerthProductFactory(price_group=berth_price_group, harbor=None)
+    product_dict = {
+        "id": to_global_id(BerthProductNode, product.id),
+        "priceValue": str(Decimal(product.price_value).quantize(Decimal("1.00"))),
```
igordavydsson, line 68: I think all of these could be just `str(product.price_value)`, since `product.price_value` is already a `Decimal` and simply casting it to string will preserve the decimal points

payments/tests/test_payments_mutations.py
```diff
@@ -0,0 +1,262 @@
...
+"""
+
+
+@pytest.mark.parametrize(
+    "api_client", ["berth_services"], indirect=True,
+)
+def test_create_berth_product(api_client, berth_price_group, harbor):
+    variables = {
+        "priceValue": str(
+            Decimal(round(random.uniform(1, 999), 2)).quantize(Decimal("1.00"))
```
igordavydsson, line 56: since you are casting this to string anyway for the sake of simplicity I would just do 
```python
"priceValue": str(round(random.uniform(1, 999), 2)),
```
here and in all similar places

payments/schema/types.py
```diff
@@ -55,21 +56,39 @@ def resolve_default_product(self, info):
...
+
+
+class BerthProductNode(DjangoObjectType, AbstractPlaceProductNode):
     price_group = graphene.Field("payments.schema.BerthPriceGroupNode", required=True)
     harbor = graphene.Field("resources.schema.HarborNode")
 
     class Meta:
         model = BerthProduct
         interfaces = (graphene.relay.Node,)
+        view_permissions = (BerthProduct,)
```
igordavydsson, line 80: what does this do and where does it come from? couldn't find mentions of it in `graphene-django`

payments/schema/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .mutations import Mutation
+from .queries import Query
+from .types import BerthPriceGroupNode, BerthProductNode, PriceUnitsEnum
+
+__all__ = [
+    "Mutation",
+    "Query",
+    "PriceUnitsEnum",
+    "BerthProductNode",
+    "BerthPriceGroupNode",
```
igordavydsson, line 15: can we do these in alphabetical order? would be easier to find where to put new things later

payments/schema/mutations.py
```diff
@@ -171,6 +176,76 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+    additional_product = graphene.Field(AdditionalProductNode)
+
+    @classmethod
+    @add_permission_required(AdditionalProduct)
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **input):
+        input["tax_percentage"] = AdditionalProductTaxEnum.get(
+            input.pop("tax_percentage")
+        ).value
```
igordavydsson, line 201: hm, tested this locally just now, seems to work fine without this, as `input` already has `Decimal` object

payments/schema/queries.py
```diff
@@ -31,3 +35,28 @@ class Query:
         WinterStorageProductNode,
         description="**Requires permissions** to access payments.",
     )
+
+    additional_products = DjangoConnectionField(
+        AdditionalProductNode,
+        product_type=AdditionalProductTypeEnum(required=False),
```
igordavydsson, line 41: `required=False` is the default

payments/models.py
```diff
@@ -10,7 +10,12 @@
 from payments.enums import AdditionalProductType, PeriodType, PriceUnits, ServiceType
 from utils.models import TimeStampedModel, UUIDModel
 
-TAX_PERCENTAGES = [Decimal(x) for x in ("24.00", "10.00")]
+PLACE_PRODUCT_TAX_PERCENTAGES = [Decimal(x) for x in ("24.00",)]
+ADDITIONAL_PRODUCT_TAX_PERCENTAGES = [Decimal(x) for x in ("24.00", "10.00")]
+
+TAX_PERCENTAGES = list(
+    set(PLACE_PRODUCT_TAX_PERCENTAGES + ADDITIONAL_PRODUCT_TAX_PERCENTAGES)
+)
```
igordavydsson, line 18: so, we aren't really using this anywhere, right? I'd say, let's sweep it under the rug for now

payments/schema/mutations.py
igordavydsson: i.e. graphene converts it from name to value internally

igordavydsson:
one bigger question that I am trying to think about right now is: how can we expose the list of fixed services to the frontend? 
Screenshot 2020-05-20 at 19 36 39
imagine we don't yet have those services on the left (satamapalvelut == fixed services), how can the frontend know, what fixed services are missing additional products. 🤔

payments/tests/test_payments_queries.py
a-rmz: The problem is that the returned value comes with 2 decimals (I guess because the `DecimalField` has a fixed number of `decimal_places=2`?).

image

So by doing the `Decimal(...).quantize(...)` it creates a decimal with 2 decimal places.

Another option would be to do `round(Decimal(...), 2)`

payments/schema/types.py
a-rmz: Whoops 😅 that was left from some tests I was doing. Removed now.

a-rmz: Whoops 😅 that was left from some tests I was doing. Removed now.

payments/schema/mutations.py
a-rmz: Ohh good catch, once again Graphene outsmarts me.

payments/models.py
a-rmz: We use it [here](https://github.com/City-of-Helsinki/berth-reservations/blob/ven-396/pricing-api/payments/models.py#L36):
```python
# payments/models.py#L36
choices=[(tax, str(tax)) for tax in TAX_PERCENTAGES],
```
only for the abstract base product. We could also do the 
```python
list(
    set(PLACE_PRODUCT_TAX_PERCENTAGES + ADDITIONAL_PRODUCT_TAX_PERCENTAGES)
)
```
directly on the model choices.

a-rmz:
> one bigger question that I am trying to think about right now is: how can we expose the list of fixed services to the frontend?
> Screenshot 2020-05-20 at 19 36 39
> imagine we don't yet have those services on the left (satamapalvelut == fixed services), how can the frontend know, what fixed services are missing additional products. 🤔
> 
> should we maybe have a separate query ~ `AdditionalProductServices(type: FIXED | OPTIONAL) {}` to let the frontend fetch lists for both? it would be quite minimal, but would do the job.

Hm, good question. We have the `ServiceType` enum, but that only gives you the "raw" types, doesn't give you any information whether they're fixed or optional.

I added a simple version of the `additionalProductServices` query as a POC, let's see if it's useful. 👍

payments/schema/__init__.py
igordavydsson: what about this? 👀

payments/schema/mutations.py
```diff
@@ -0,0 +1,118 @@
+import graphene
+from django.core.exceptions import ValidationError
+from django.db import IntegrityError, transaction
+
+from berth_reservations.exceptions import VenepaikkaGraphQLError
+from payments.models import BerthProduct
```
igordavydsson: please, do relative import for all in-app imports

payments/schema/mutations.py
```diff
@@ -0,0 +1,118 @@
...
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **input):
+        product = get_node_from_global_id(
+            info, input.pop("id"), only_type=BerthProductNode, nullable=False
+        )
+        if input.get("harbor_id"):
+            input["harbor"] = get_node_from_global_id(
+                info, input.pop("harbor_id", None), only_type=HarborNode, nullable=True,
+            )
+        if "price_group_id" in input:
```
igordavydsson, line 77: should these two if statements follow the same pattern? i.e. either `if input.get("input_field")` or `if "input_field" in input`

payments/schema/types.py
```diff
@@ -0,0 +1,58 @@
+import graphene
+from graphene_django import DjangoConnectionField, DjangoObjectType
+
+from payments.enums import PriceUnits
+from payments.models import BerthPriceGroup, BerthProduct, TAX_PERCENTAGES
```
igordavydsson: relative imports

payments/schema/mutations.py
a-rmz: True! Good catch, especially since harbor is nullable.

payments/tests/test_payments_queries.py
a-rmz: Solved by changing how the `AbstractBaseProductFactory` generates the `price_value`


[payments/tests/factories.py#L23](https://github.com/City-of-Helsinki/berth-reservations/pull/211/files#diff-b5be5d575a3ebb66c94f3efd538238c5R23)
```python
price_value = factory.LazyFunction(
    lambda: round(Decimal(random.uniform(1, 999)), 2)
)
```

igordavydsson: so, this should now work with just `str(product.price_value)`?

a-rmz: Yup! It has [already been fixed](https://github.com/City-of-Helsinki/berth-reservations/pull/211/files#diff-ba338259a0da5db71cb5e6bb5bd5b724R69)

payments/models.py
igordavydsson: if we override the `tax_percentage` field on both child models of the abstract model, then we probably do not really need it there 🤔

payments/schema/queries.py
```diff
@@ -61,3 +65,27 @@ def resolve_additional_products(self, info, **kwargs):
                 )
 
         return AdditionalProduct.objects.all()
+
+    def resolve_additional_product_services(self, info, **kwargs):
+        services = []
+        service_list = []
+        product_type = kwargs.get("product_type")
+        if product_type:
+            product_type = AdditionalProductType(product_type)
```
igordavydsson, line 74: not sure this line is needed 🤔

payments/schema/queries.py
```diff
@@ -61,3 +65,27 @@ def resolve_additional_products(self, info, **kwargs):
...
+        else:
+            service_list = list(ServiceType)
+
+        for service in service_list:
+            services.append(
+                AdditionalProductServiceNode(
+                    service=service,
+                    product_type=AdditionalProductType.FIXED_SERVICE
+                    if service.is_fixed_service()
+                    else AdditionalProductType.OPTIONAL_SERVICE,
```
igordavydsson, line 88: maybe a more generic way of figuring out `product_type` would be to have a resolver on the node?
```python
class AdditionalProductServiceNode(graphene.ObjectType):
    service = ServiceTypeEnum(required=True)
    product_type = AdditionalProductTypeEnum(required=True)

    class Meta:
        interfaces = (graphene.relay.Node,)
        connection_class = CountConnection
    
    def resolve_product_type(self, info):
        return AdditionalProductType.FIXED_SERVICE if self.service.is_fixed_service() else AdditionalProductType.OPTIONAL_SERVICE
```

this wayyou can only pass the service whenever you need to init the node

igordavydsson: also you could skip initing `services` list and just do:
```python
def resolve_additional_product_services(self, info, **kwargs):
    service_list = list(ServiceType)
    product_type = kwargs.get("product_type")
    if product_type:
        if product_type == AdditionalProductType.FIXED_SERVICE:
            service_list = ServiceType.FIXED_SERVICES()
        elif product_type == AdditionalProductType.OPTIONAL_SERVICE:
            service_list = ServiceType.OPTIONAL_SERVICES()
    return [AdditionalProductServiceNode(service=service) for service in service_list]
```

igordavydsson: apart from these two nitpicks, a solid implementation of the enum query 👍 👍 👍

payments/schema/queries.py
a-rmz: The value that the resolver is receiving is a `str`, so we need to either (1) cast it to the Enum value
```python
or (2) check against the enum `.name`, `if product_type == AdditionalProductType.FIXED_SERVICE.name`

File: payments/models.py (hunk 10 above)
Comment (a-rmz) [reply payments/models.py:18]: Yeah, I was thinking that... but I decided to keep the field to indicate that any product needs to have a taxable value and to have a set of defaults in case in the future, another model would not override it.

However, we can remove it if it's doing no good to anyone. :)

Comment (igordavydsson) [reply payments/models.py:18]: I feel like it adds more complexity for people who would be new to the codebase. They would see this field and its choices and will have to track how each product overrides it and will have to understand that the field on the base class is an example of sorts or a default for potential future child classes. I think it would make people more cautious if we remove the field, so that they would see that `tax_percentage` is missing unless they add it explicitly where they need it.

File: payments/schema/queries.py (hunk 14 above)
Comment (igordavydsson) [reply payments/schema/queries.py:74]: ah, okay, then let's cast it 👍

File: payments/models.py (hunk 10 above)
Comment (a-rmz) [reply payments/models.py:18]: Makes sense, it would require more caution when implementing other child products.

The dream would be to have kinda "configurable" choices so we could have something like this:
```python
class AbstractTaxableProduct:
    tax_percentage = DecimalField(
        choices=self._meta.TAX_OPTIONS
    )

class ConcreteProduct(AbstractTaxableProduct):
    class Meta:
        TAX_OPTIONS = [10, 20]
```

But I guess removing it for the parent class will do for now. 😅

igordavydsson: configurable choices would be super cool 😻 let's see what will happen with the webshop microservice and then add more sugar like this, if we will still have this code in our codebase

payments/schema/queries.py
igordavydsson: I see, graphene converts its Enums to django-enumfields' Enums when we use them as fields on DjangoObjectTypes, but does not do so when they are just arguments on queries 😢 well, let's just cast, yeah

PR #214, opened 2020-06-01
[VEN-304] Invoicing models

a-rmz:
## Description :sparkles:
- Add models for `Order`, `OrderLine` and `OrderLogEntry`
- Add those models to the admin
- Add tests
- Add some logic to the `Order` creation:
  - When placing an order associated with a `BerthLease`, it determines the pier services and creates the related `OrderLine`s with the price according to the duration of the lease.

### Closes :no_good_woman:
**[VEN-304](https://helsinkisolutionoffice.atlassian.net/browse/VEN-304):** data model for invoices

### Automated tests :gear:️
```shell
```

Try adding instances from the [admin](http://localhost:8000/admin/payments/)

## Screenshots :camera_flash:
**Admin models:**
image

**Add Order admin:**
image

The `Order` model uses the [`ContentTypes`](https://docs.djangoproject.com/en/2.2/ref/contrib/contenttypes/) to accepth both `BerthProduct` or `WinterStorageProduct` on the same `product` field; same thing for the `lease` field.

Because of that, the `Add Order` admin page cannot look up for leases as it would do with a normal foreign key. The `Product` has a choice field since, ideally, there wouldn't be a big number, so it would be reasonable to display it on a `select`.


payments/models.py
```diff
@@ -154,3 +169,309 @@ def __str__(self):
...
+
+    objects = OrderManager()
+
+    def __str__(self):
+        return f"{self.product} [{self.status}]"
+
+    @property
+    @rounded
+    def full_price(self):
+        return convert_pretax_to_aftertax(self.base_price, self.tax_percentage)
```
igordavydsson, line 250: I think this is the other way around: we want to store taxful price and then subtract taxes programmatically 🤔 
e.g. the orders inported from Timmi will have their total taxful price and tax percentage

payments/models.py
```diff
@@ -154,3 +169,309 @@ def __str__(self):
...
+
+class Order(UUIDModel, TimeStampedModel):
+    customer = models.ForeignKey(
+        "customers.CustomerProfile",
+        verbose_name=_("customer"),
+        on_delete=models.CASCADE,
+        related_name="orders",
+    )
+    product = GenericForeignKey("_product_content_type", "_product_object_id")
+    lease = GenericForeignKey("_lease_content_type", "_lease_object_id")
```
igordavydsson, line 200: these look very nice 😻

payments/models.py
```diff
@@ -154,3 +169,309 @@ def __str__(self):
...
+    status = EnumField(OrderStatus, default=OrderStatus.WAITING)
+    comment = models.TextField(blank=True, null=True)
+    base_price = models.DecimalField(
+        verbose_name=_("price"),
+        max_digits=10,
+        decimal_places=2,
+        validators=[MinValueValidator(Decimal("0.01"))],
+        blank=True,
+        null=True,
```
igordavydsson, line 209: not sure that the base price (be in taxul or taless) should be blankable or nullable 🤔  it is one of the main attributes of the order, and we probably want to have always defined

payments/models.py
```diff
@@ -154,3 +169,309 @@ def __str__(self):
...
+    )
+    tax_percentage = models.DecimalField(
+        verbose_name=_("tax percentage"),
+        max_digits=5,
+        decimal_places=2,
+        blank=True,
+        null=True,
```
igordavydsson, line 216: same with tax - also one of the main attributes

payments/models.py
```diff
@@ -154,3 +169,309 @@ def __str__(self):
...
+        if all([not self.product, not self._product_object_id, not self.base_price]):
+            raise ValidationError(
+                _("Order must have either product object or price value")
+            )
+        if self.lease and self.lease.customer != self.customer:
+            raise ValidationError(
+                _("The lease provided belongs to a different customer")
+            )
+
+        super().clean()
```
igordavydsson, line 285: no need to call `super()` in clean, since it is an empty hook on the parent model ([ref](https://github.com/django/django/blob/master/django/db/models/base.py#L990))

payments/models.py
a-rmz: I agree, but _in theory_, they will never actually be null. The reason to have them as `blank`/`null`, is so that we can receive "empty" values which then will be set based on the `product` values.

Another option would be to make them non-nullable, and call the `clean()` after the values have been set.

https://github.com/City-of-Helsinki/berth-reservations/pull/214/files#diff-409ee34417f21f35e34b41dfabf4720cR382-R385

payments/models.py
a-rmz: I thought about that as well 🤔 it's not really necessary, but I left it in case we would, at some point in the future, do some validation on any parent model. But since it probably won't be the case, I'll get rid of it.

payments/models.py
a-rmz: What would be the the advantage of having taxful vs taxless? Is it only the Timmi import? _In my mind_, it makes more sense to store the actual base price, but what say you?

payments/models.py
igordavydsson: yeah, ok, this _base_ price can be nullable and blankable 👍

payments/models.py
igordavydsson: as per Slack discussion these are now changed to `price with tax` and `tax percentage`. thread is resolved 👍

payments/models.py
igordavydsson: 👀

payments/models.py
```diff
@@ -154,3 +169,334 @@ def __str__(self):
...
+                )
+        else:
+            if (
+                self.price
+                and self.tax_percentage
+                and (self.product or self._product_object_id)
+            ):
+                raise ValidationError(
+                    _("Cannot receive product and plain price/tax_percentage")
+                )
```
igordavydsson, line 399: all of these are pure validation, maybe put them in the `clean`?

payments/models.py
```diff
@@ -154,3 +169,334 @@ def __str__(self):
...
+                    _("Cannot change the lease associated to this order")
+                )
+
+            # If the price/tax are being changed
+            if (
+                old_instance.price != self.price
+                or old_instance.tax_percentage != self.tax_percentage
+            ):
+                raise ValidationError(
+                    _("Cannot receive product and plain price/tax_percentage")
```
igordavydsson, line 389: seems that message needs updating

payments/models.py
```diff
@@ -154,3 +169,334 @@ def __str__(self):
...
+                if lease
+                else WinterStorageLease.objects.filter(id=self._lease_object_id).first()
+            )
+            # If for some reason neither was found, raise an error
+            if not lease:
+                raise ValidationError(_("The lease passed is not valid"))
+
+            self.lease = lease
+
+        if not creating:
```
igordavydsson, line 369: usually you achieve better readability when you start with `if creating` (or in general with the `True` logic)

payments/models.py
```diff
@@ -154,3 +169,345 @@ def __str__(self):
...
+                if product
+                else WinterStorageProduct.objects.filter(
+                    id=self._product_object_id
+                ).first()
+            )
+            # If for some reason neither was found (shouldn't be the case), raise an error
+            if not product:
+                raise ValidationError(_("The product passed is not valid"))
+
+            self.product = product
```
igordavydsson, line 378: maybe this could be encapsulated in a helper function `self._assign_product_from_object_id()`? 🤔  would clean up `save` a bit

payments/models.py
```diff
@@ -154,3 +169,345 @@ def __str__(self):
...
+            if not lease:
+                raise ValidationError(_("The lease passed is not valid"))
+
+            self.lease = lease
+
+        creating = self._state.adding
+        if creating:
+            # If the product instance is being passed
+            # Price has to be assigned before saving but only if creating
+            if self.product:
```
igordavydsson, line 400: you can decrease one `if` by doing `if creating and self.product:` maybe

a-rmz: Good catch! The structure was left from when there were more things inside the `if`, but now that there's not, it can be simplified. 😄

payments/models.py
```diff
@@ -154,3 +169,345 @@ def __str__(self):
...
+                if not self.tax_percentage:
+                    self.tax_percentage = self.product.tax_percentage
+        else:
+            old_instance = OrderLine.objects.get(id=self.id)
+            # If the product is being changed
+            if old_instance.order != self.order:
+                raise ValidationError(
+                    _("Cannot change the order assigned to this order line")
+                )
+        return super().save(*args, **kwargs)
```
igordavydsson, line 487: you don't need to `return` it 😉

payments/models.py
```diff
@@ -154,3 +169,345 @@ def __str__(self):
...
+        blank=True,
+        null=True,
+    )
+
+    def save(self, *args, **kwargs):
+        # TODO: validate that order and product are not being changed
+        self.full_clean()
+
+        if self._state.adding:
+            if self.product:
```
igordavydsson, line 459: could be one one line

payments/models.py
```diff
@@ -154,3 +169,356 @@ def __str__(self):
...
+            if lease
+            else WinterStorageLease.objects.filter(id=self._lease_object_id).first()
+        )
+        # If for some reason neither was found, raise an error
+        if not lease:
+            raise ValidationError(_("The lease passed is not valid"))
+
+        self.lease = lease
+
+    def save(self, *args, **kwargs):  # noqa: C901  is too complex
```
igordavydsson, line 389: maybe flake would not complain anymore? 👀

payments/models.py
```diff
@@ -154,3 +169,356 @@ def __str__(self):
...
+        Order,
+        verbose_name=_("order"),
+        related_name="order_lines",
+        on_delete=models.CASCADE,
+    )
+    product = models.ForeignKey(
+        AdditionalProduct,
+        verbose_name=_("product"),
+        related_name="orders",
+        on_delete=models.CASCADE,
```
igordavydsson, line 436: hm, what if at some point someone deletes the product? it would cascade this line, thus affecting the total order price and everything 🤔 maybe this should be nullable?

payments/models.py
```diff
@@ -154,3 +169,356 @@ def __str__(self):
...
+    quantity = models.PositiveSmallIntegerField(
+        verbose_name=_("quantity"), default=1, validators=[MinValueValidator(1)],
+    )
+    price = models.DecimalField(
+        verbose_name=_("price"),
+        max_digits=10,
+        decimal_places=2,
+        validators=[MinValueValidator(Decimal("0.01"))],
+        blank=True,
+        null=True,
```
igordavydsson, line 447: but then the `price` (and `tax_percentage`) down below would not be nullable, maybe? they would store the price and tax of the product at the time of order line creation...

payments/models.py
a-rmz: True, it would make sense since we're keeping price and tax, so if the product would be deleted, we can still keep track.

payments/models.py
a-rmz: The only reason atm to have them blankable/nullable would be to allow creating `OrderLine`s from the admin without explicitly passing them (when passing a product).

igordavydsson: hm, let's have them a blankable then, maybe? since that would allow leaving them empty in Django admin, but would block not setting any value

igordavydsson: i.e. without `null=True`

a-rmz: Great idea! 👍 I'll do the same for price/tax on `Order` and `OrderLine`

igordavydsson: yeah, I was on the edge about allowing `null`s for prices and taxes on the `Order`. this might lead to orders with empty prices and distort the bookkeeping, better to ensure the price is always set on the DB level

payments/models.py
```diff
@@ -154,3 +169,348 @@ def __str__(self):
...
+    comment = models.TextField(blank=True, null=True)
+    price = models.DecimalField(
+        verbose_name=_("price"),
+        max_digits=10,
+        decimal_places=2,
+        validators=[MinValueValidator(Decimal("0.01"))],
+        blank=True,
+    )
+    tax_percentage = models.DecimalField(
+        verbose_name=_("tax percentage"), max_digits=5, decimal_places=2, blank=True,
```
igordavydsson, line 211: should we have validation here? `validators=[MinValueValidator(Decimal("0.00"))],` or something?

payments/models.py
```diff
@@ -154,3 +169,348 @@ def __str__(self):
...
+    def __str__(self):
+        return f"{self.product} [{self.status}]"
+
+    @property
+    @rounded
+    def pretax_price(self):
+        return (
+            convert_aftertax_to_pretax(self.price, self.tax_percentage)
+            if self.price and self.tax_percentage
+            else None
```
igordavydsson, line 248: if it ever ends up in `None` it will crash `rounded` decorator, since it expects a decimal. should it ever be `we maybe drop the ip statement here?

payments/models.py
```diff
@@ -154,3 +169,348 @@ def __str__(self):
...
+        return sum(
+            [ol.pretax_price for ol in self.order_lines.all()], self.pretax_price
+        )
+
+    @property
+    @rounded
+    def total_tax_percentage(self):
+        return ((self.total_price - self.total_pretax_price) / self.total_price) * 100
+
+    def clean(self):  # noqa: C901  is too complex
```
igordavydsson, line 268: is this now really too complex for flake? 😭  maybe enscapsulate it into some `_validate_order_lease` etc.?

a-rmz: Yes:( I'll try to encapsulate it a bit.

payments/models.py
a-rmz: The reason to do it like that was (again) because of the admin. This time, the `OrderLine` inline on `Order` creation.

image

However, when adding that `else None`, everything goes fine.

image

igordavydsson: maybe then add the logic on the model admin, like:
```python
def pretax_price(self):
    return self.pretax_price or None
```
or something similar, just to not have that on the model directly?

payments/models.py
```diff
@@ -154,3 +169,348 @@ def __str__(self):
...
+    )
+
+    def clean(self):
+        creating = self._state.adding
+        if not creating:
+            old_instance = OrderLine.objects.get(id=self.id)
+            # If the order is being changed
+            if old_instance.order and self.order != old_instance.order:
+                raise ValidationError(
+                    _("Cannot change the order associated to this order")
```
igordavydsson, line 456: `associated WITH this order LINE`?

payments/models.py
```diff
@@ -154,3 +169,348 @@ def __str__(self):
...
+                )
+
+            # If the product is being changed
+            if old_instance.product != self.product:
+                raise ValidationError(
+                    _("Cannot change the product assigned to this order")
```
igordavydsson, line 462: `to this order LINE`?

payments/models.py
a-rmz: That would be ideal, but the error is raised on the division inside `convert_aftertax_to_pretax` when a`None` is passed instead of a number:
```
 return aftertax_price / (1 + tax_percentage / 100)
TypeError: unsupported operand type(s) for /: 'NoneType' and 'int'
```

(I tried doing that on the admin and didn't work either)

a-rmz: Scratch that, if you check for `price/tax` on the admin, you can avoid the error. 👍

payments/models.py
```diff
@@ -154,3 +171,376 @@ def __str__(self):
...
+            ):
+                # Calculate the actual price for the amount of days on the period
+                # price = (days_on_lease * product.price_value) / season_days
+                price = calculate_product_partial_season_price(
+                    price, self.order.lease.start_date, self.order.lease.end_date
+                )
+            elif self.product.period == PeriodType.YEAR:
+                price = calculate_product_partial_year_price(
+                    price, self.order.lease.start_date, self.order.lease.end_date
+                )
```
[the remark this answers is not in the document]
a-rmz: Hummm good catch. We should check for the lease _before_.

Follow up question: if there's no lease, we can't determine if there's a "partial" price, so, should we charge for the whole product price?

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,632 @@ def test_additional_product_fixed_service_tax_value():
...
+
+def test_order_berth_product_winter_storage_lease_raise_error():
+    lease = WinterStorageLeaseFactory()
+    with pytest.raises(ValidationError) as exception:
+        Order.objects.create(
+            product=BerthProductFactory(), lease=lease, customer=lease.customer
+        )
+
+    errors = str(exception.value)
+    assert f"A BerthProduct must be associated with a BerthLease" in errors
```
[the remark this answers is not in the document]
a-rmz: Copy-paste :derp:

payments/models.py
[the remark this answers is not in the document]
igordavydsson: yeah, I think full product price will be the default then

payments/tests/test_payments_models.py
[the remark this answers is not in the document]
igordavydsson: :pepejam:

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,632 @@ def test_additional_product_fixed_service_tax_value():
...
+            product=winter_storage_product,
+            lease=winter_storage_lease,
+            customer=customer_profile,
+        )
+
+    errors = str(exception.value)
+    assert "The lease provided belongs to a different customer" in errors
+
+
+def test_order_berth_product_creates_pier_order_lines(harbor):
```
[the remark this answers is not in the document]
a-rmz: 🤔 ok, ok, makes sense.

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,632 @@ def test_additional_product_fixed_service_tax_value():
...
+            order_line.order.lease.end_date,
+        )
+    elif product.period == PeriodType.YEAR:
+        expected_price = calculate_product_partial_year_price(
+            product.price_value,
+            order_line.order.lease.start_date,
+            order_line.order.lease.end_date,
+        )
+    assert order_line.price == expected_price
+    assert order_line.tax_percentage == product.tax_percentage
```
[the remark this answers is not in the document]
a-rmz: Good idea 😸

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,632 @@ def test_additional_product_fixed_service_tax_value():
...
+            assert product_price == order_price
+
+
+@freeze_time("2020-01-01T08:00:00Z")
+def test_order_winter_storage_product_creates_pier_for_partial_month(
+    winter_storage_area,
+):
+    services = {
+        "summer_storage_for_docking_equipment": random_bool(),
+        "summer_storage_for_trailers": random_bool(),
```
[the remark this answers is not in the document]
a-rmz: Or add a loop that loops infinitely until at least one is true? :derp:

Okay, maybe just setting one to `True` is enough.

payments/utils.py
```diff
@@ -0,0 +1,104 @@
...
+    return date.today() + timedelta(days=delta)
+
+
+@rounded
+def calculate_product_partial_month_price(
+    base_price: Decimal, start_date: date, end_date: date
+) -> Decimal:
+    _, days_in_month = calendar.monthrange(start_date.year, start_date.month)
+    delta = (end_date - start_date).days
+    price = (delta * base_price) / days_in_month
```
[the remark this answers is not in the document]
a-rmz: How would we bill for every month? The `order` is generated for/with a `lease`, with a determined duration. But how are the clients billed for that? I.e. where would we say what's the price that the customer has to pay for either month (in this case)?

payments/utils.py
```diff
@@ -0,0 +1,104 @@
...
+def calculate_product_partial_season_price(
+    base_price: Decimal, start_date: date, end_date: date, summer_season: bool = True
+) -> Decimal:
+    # If it's the "normal" (summer) season, calculate with the normal dates
+    season_days = calculate_berth_lease_end_date() - calculate_berth_lease_start_date()
+    # If it's for the opposite season ("winter season"), calculate the inverse
+    if not summer_season:
+        # Calculate the amount of days in the year, in case it's a leap year
+        day = today()
+        days_on_year = day.replace(year=day.year) - day
```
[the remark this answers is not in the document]
a-rmz: Whoops, it was supposed to be `year=day.year + 1`. 😅

payments/utils.py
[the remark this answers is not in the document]
igordavydsson: nonono, I donät mean billing every months, I mean the total price calculation when you have a product with `period == MONTH`

payments/tests/factories.py
```diff
@@ -67,10 +73,56 @@ class AdditionalProductFactory(AbstractBaseProductFactory):
     # been assigned to the model.
     @factory.post_generation
     def service_taxes(self, created, extracted, **kwargs):
+        if self.price_unit == PriceUnits.PERCENTAGE:
+            self.price_value = random_price(1, 100, decimals=0)
```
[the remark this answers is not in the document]
a-rmz: It has it! `AdditionalProductFactory` inherits from `AbstractBaseProductFactory`, and
```python
class AbstractBaseProductFactory(factory.django.DjangoModelFactory):
    price_value = factory.LazyFunction(
        lambda: round(Decimal(random.uniform(1, 999)), 2)
    )
```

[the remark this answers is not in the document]
igordavydsson: ahhh, well, then very good catch and welcome addition that the `percentage` is generated between `0` and `100` :pepejam: 🙂

payments/utils.py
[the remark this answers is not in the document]
igordavydsson: beware, `days_on_year = day.replace(year=day.year + 1) - day` will always be `datetime.timedelta(1)` 🤔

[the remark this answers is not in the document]
igordavydsson: ah, scratch that :derp:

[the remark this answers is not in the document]
igordavydsson: yeah, makes sense with the `+ 1` 👍

payments/utils.py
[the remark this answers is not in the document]
a-rmz: AHHHH, okay, gotcha. I was under the assumption that monthly products would only be assigned to "monthly" products. So, I guess that instead of counting `days_in_month` we can just calculate the day difference.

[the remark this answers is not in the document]
a-rmz: How about the "full month" case? Say:
```
start_date = june 15
end_date = september 22
```

We calculate partially `june` and `september`, but for `july + august` there's two options:
1. Add the full month price _as-is_.
2. Again, calculate the "partial" month price to adjust the base monthly price to either 30-31 days (or 28-29)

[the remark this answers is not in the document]
igordavydsson: so:
1. full month price as is for july and for august, since those months are rented from the first day to the last
2. for june you: (take month price / 30 days in june) * 16 days of june that the user is renting it for
3. for september you: (take month price / 30 days in september) * 22 days of september that the user is renting it for
(4). if it would have been october 22 as the end date: (take month price / 31 days in october) * 22 days of october that the user is renting it for

payments/models.py
```diff
@@ -154,3 +171,376 @@ def __str__(self):
...
+
+            if not self.price:
+                self.price = price
+            if not self.tax_percentage:
+                self.tax_percentage = self.product.tax_percentage
```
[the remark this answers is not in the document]
a-rmz: That validation is done for [`Order`s](https://github.com/City-of-Helsinki/berth-reservations/pull/214/files#diff-409ee34417f21f35e34b41dfabf4720cR307-R315)
```python
if (
    self.price
    and self.tax_percentage
    and (self.product or self._product_object_id)
):
    raise ValidationError(
        _("Cannot receive product and plain price/tax_percentage")
    )
```

But not for `OrderLine`. Should we also add it here?

payments/models.py
```diff
@@ -154,3 +171,376 @@ def __str__(self):
...
+
+        creating = self._state.adding
+        # If the product instance is being passed
+        # Price has to be assigned before saving but only if creating
+        if creating and self.product:
+            # Assign the price and tax from the product
+            if not self.price:
+                self.price = self.product.price_value
+            if not self.tax_percentage:
+                self.tax_percentage = self.product.tax_percentage
```
igordavydsson, line 422: ok, then this needs changing 🙂 since here validation exists

payments/models.py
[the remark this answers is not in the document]
igordavydsson: still not checking if `order` has `lease`? 👀

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,686 @@ def test_additional_product_fixed_service_tax_value():
...
+def test_order_winter_storage_product_berth_lease_raise_error():
+    lease = BerthLeaseFactory()
+    with pytest.raises(ValidationError) as exception:
+        Order.objects.create(
+            product=WinterStorageProductFactory(), lease=lease, customer=lease.customer
+        )
+
+    errors = str(exception.value)
+    assert (
+        f"A WinterStorageProduct must be associated with a WinterStorageLease" in errors
```
igordavydsson, line 156: formatted string :simple-smile: :hessu: :derp:

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,686 @@ def test_additional_product_fixed_service_tax_value():
...
+        if created:
+            assert (
+                OrderLine.objects.filter(
+                    order=order, product__service=ProductServiceType(service)
+                ).count()
+                == 1
+            )
+
+
+def test_order_no_lease_no_products(customer_profile, harbor):
```
igordavydsson, line 325: `..._no_lease_no_ORDER_LINES`? 🤔

igordavydsson: since you are testing that order lines would not get generated...

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,686 @@ def test_additional_product_fixed_service_tax_value():
...
+            )
+            order_price = order_line.price
+            assert partial_product_price == order_price
+
+
+@freeze_time("2020-01-01T08:00:00Z")
+def test_order_winter_storage_lease_right_price_for_partial_year(winter_storage_area,):
+    services = {
+        "summer_storage_for_docking_equipment": random_bool(),
+        "summer_storage_for_trailers": random_bool(),
```
igordavydsson, line 549: also here only two options and both roll the dice with the `random_bool()` 🤔

payments/tests/test_payments_models.py
a-rmz: Whoopsie-daisie :derp:

payments/utils.py
```diff
@@ -0,0 +1,134 @@
...
+@rounded
+def calculate_product_partial_season_price(
+    base_price: Decimal, start_date: date, end_date: date, summer_season: bool = True
+) -> Decimal:
+    # If it's the "normal" (summer) season, calculate with the normal dates
+    season_days = calculate_berth_lease_end_date() - calculate_berth_lease_start_date()
+    # If it's for the opposite season ("winter season"), calculate the inverse
+    if not summer_season:
+        # Calculate the amount of days in the year, in case it's a leap year
+        day = today()
```
igordavydsson, line 114: shouldn't this be `start_date` btw? 🤔

payments/utils.py
```diff
@@ -0,0 +1,134 @@
...
+        days_on_year = (day + relativedelta(years=1)) - day
```
igordavydsson, line 115: `days_IN_year`? 🤔

payments/utils.py
a-rmz: 🤔 makes sense.

payments/utils.py
a-rmz: 🇬🇧

igordavydsson: [google not sure](https://www.google.com/search?sxsrf=ALeKk02qZSkRC2VFP43qfE5qwJqQ3PW6-w%3A1591879522792&ei=YifiXqbhL6qqrgSY1JPACQ&q=%22days+on+year%22&oq=%22days+on+year%22&gs_lcp=CgZwc3ktYWIQAzIGCAAQFhAeMgYIABAWEB4yBggAEBYQHjIGCAAQFhAeMgYIABAWEB4yBggAEBYQHjIGCAAQFhAeMgYIABAWEB4yBggAEBYQHjIGCAAQFhAeOgQIABBHOgQIABBDOgYIABAHEB46BAgAEApQl94PWMvxD2CR9A9oAHABeACAAXWIAdcBkgEDMS4xmAEAoAEBqgEHZ3dzLXdpeg&sclient=psy-ab&ved=0ahUKEwimmuyr5fnpAhUqlYsKHRjqBJgQ4dUDCAw&uact=5)

a-rmz: 🤷 days`days_IN_year` works fine.

payments/utils.py
```diff
@@ -0,0 +1,132 @@
...
+                MONTHLY,
+                dtstart=start_date + relativedelta(day=1, months=1),
+                until=end_date - relativedelta(day=1, months=1),
+            )
+        )
+        # Calculate the added price for all the complete months in between
+        price += len(months_between) * base_price
+
+        # Calculate price for the end month
+        price += _calculate_price_for_partial_month(end_date, base_price)
```
igordavydsson, line 99: so, for if the end date is 22nd of september, you would charge for `(30 - 22) * price_per_day`? 🤔

a-rmz: Sneaky catch! For the start month we should do
```python
abs(start.day - days_IN_month) * price
```

and for the end month
```python
abs(days_IN_month - end.days) * price
```

igordavydsson: yep 👍

payments/tests/test_payments_models.py
```diff
@@ -107,3 +131,686 @@ def test_additional_product_fixed_service_tax_value():
...
+                order_line.product.price_value,
+                order.lease.start_date,
+                order.lease.end_date,
+            )
+            order_price = order_line.price
+            assert partial_product_price == order_price
+
+
+@freeze_time("2020-01-01T08:00:00Z")
+def test_order_winter_storage_lease_right_price_for_partial_season(harbor):
```
igordavydsson, line 503: berth lease in this test case?

payments/models.py
```diff
@@ -154,3 +171,374 @@ def __str__(self):
...
+            self.price
+            and self.tax_percentage
+            and (self.product or self._product_object_id)
+        ):
+            raise ValidationError(
+                _("Cannot receive product and plain price/tax_percentage")
+            )
+
+    def _check_same_product(self, old_instance) -> None:
+        if old_instance.product != self.product:
```
igordavydsson, line 318: should this be `if old_instance.product and old_instance.product != self.product:`? or what did we agree on?

payments/models.py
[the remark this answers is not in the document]
igordavydsson: still valid thread 🙂

payments/utils.py
```diff
@@ -0,0 +1,132 @@
...
+    delta = (end_date - start_date).days
+    price = (delta * base_price) / season_days.days
+    return price
+
+
+@rounded
+def calculate_product_partial_year_price(
+    base_price: Decimal, start_date: date, end_date: date
+) -> Decimal:
+    days_in_year = 366 if calendar.isleap(start_date.year) else 365
```
igordavydsson, line 124: what if the start date is 30th of September and end date is 30th of September next year AND start year is not a leap year, but end year is 🤔

payments/models.py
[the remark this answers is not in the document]
igordavydsson: ok, I gave it some more thought and I believe now that this is a good rule for both `Order` and `OrderLine` models: allow passing both a product and a price. since if it is possible to override the price after an object is created anyway, it is kinda silly that we do not allow it during `create`. if both were passed, price overrides product's price.

payments/models.py
```diff
@@ -154,3 +171,374 @@ def __str__(self):
...
+        if self._lease_object_id and not self.lease:
+            self._assign_lease_from_object_id()
+
+        creating = self._state.adding
+        # If the product instance is being passed
+        # Price has to be assigned before saving but only if creating
+        if creating and self.product:
+            # Assign the price and tax from the product
+            self.price = self.product.price_value
+            self.tax_percentage = self.product.tax_percentage
```
igordavydsson, line 420: so this should be something like: 
```python
if creating and any([not self.price, not self.tax_percentage]):
    price = self.product.price
    tax_percentage = self.product.tax_percentage
    if self.lease and not self.is_full_season():
        price = calculate_lease_partial_price()
    self.price = self.price or price
    self.tax_percentage = self.tax_percentage or tax_percentage
```

payments/models.py
```diff
@@ -154,3 +171,374 @@ def __str__(self):
...
+            self._check_same_product(old_instance)
+
+            # If the lease is being changed
+            self._check_same_lease(old_instance)
+
+    def is_full_season(self):
+        return (
+            self.lease.start_date == calculate_berth_lease_start_date()
+            and self.lease.end_date == calculate_berth_lease_end_date()
+        )
```
igordavydsson, line 356: this should probably handle no `self.lease` more gracefully...

payments/models.py
```diff
@@ -154,3 +171,374 @@ def __str__(self):
...
+            # Check that product and lease are from the same type
+            if self.product:
+                self._check_product_and_lease()
+
+        # Check that it has either product or price
+        self._check_product_or_price()
+
+        if self._state.adding:
+            # Don't allow creating with product and price
+            self._check_product_and_price()
```
igordavydsson, line 343: and then, I guess, this is not needed 🤔

payments/models.py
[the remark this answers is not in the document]
igordavydsson: btw, would be nice to only do these calculations `if not self.price` 🤔

payments/models.py
a-rmz: Hmmm... :big-hm:

`old_instance` _should_ always be `not None`

```python
if self._state.adding:
    # Don't allow creating with product and price
    self._check_product_and_price()
else:
    old_instance = Order.objects.get(id=self.id)
    # If the product is being changed
    self._check_same_product(old_instance)
```

So `_check_same_product` (or `_check_same_lease`) are only called when the instance is updating.

igordavydsson: yeah, but the old instance might have NO product ☝️

payments/utils.py
a-rmz: Even worse: what if:
```
start = date(2019, 9, 30)
end = date(2021, 9, 30)
```
(would that be even possible?)
Given that 2020 _is_ leap year, how should we determine the yearly price? Would it be an average of days (`(365 + 366 + 365)/3`) or do we take the price for each year individually and then add it or..?

If start/end dates weren't allowed to be over a year apart (i.e. delta < 366), we can just check for something like this?
```python
if (
    is_leap(start) or
    end.month >= MARCH and is_leap(end):
):
    days = 366
```

payments/models.py
a-rmz: Would it the

a-rmz:
```python
if self.lease and not self.is_full_season():
        price = calculate_lease_partial_price()
```
So that would `calculate_product_partial_season_price` with the `berth/ws product` as "base" price?

payments/utils.py
igordavydsson: yeah, I was thinking about this too yesterday. we _should not_ have lease longer than a year business-wise. they are however possible technically. I like the idea of not allowing creating orders for lease longer than 366 days 🤔

payments/models.py
a-rmz: Hmmm good catch. We could either check it _before_ calling `is_full_season` or... I don't think returning `False` would make sense in this context. Raising an error maybe?

payments/models.py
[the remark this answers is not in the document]
a-rmz: Alrighty then, let's aim for a standard behaviour and rely on the front-end/API to only allow either of them at a time.

payments/models.py
a-rmz: In that case, `if old_instance.product != self.product ---> None != BerthProduct()`

payments/utils.py
a-rmz: So we can restrict them from the `leases` app..?

payments/models.py
igordavydsson: hm, but it would already raise an `AttrivbuteError` :simple_smile: 🤔

igordavydsson: ok, let's assume that this check should always follow `if order.lease` 🙂

payments/models.py
igordavydsson: yeah, berth/ws products are restricted to season, so just calculate based on that

payments/utils.py
igordavydsson: hm, good idea, maybe we should do it in the leases. let's disallow for now any lease longer than 1 year

payments/models.py
igordavydsson:
```python
if all([old_instance.product, self.product, old_instance.product != self.product]):
    raise ValidationError()
```
like this? 🤔

payments/utils.py
a-rmz: Ok ok, then let's assume that leases won't be over a year and work on that later. 👍

igordavydsson: sure, let's do that 👍

payments/models.py
a-rmz: Ok ok, checking for checks. 👍

payments/models.py
a-rmz: Sure 🤷 worth checking for everything.

payments/models.py
igordavydsson: I mean, this function can maybe stay as is. 🙂

a-rmz: Yups, I meant that I'll check that we check for `lease` before calling it. :derp:

payments/models.py
```diff
@@ -154,3 +171,377 @@ def __str__(self):
...
+        if all(
+            [old_instance.product, self.product, old_instance.product != self.product]
+        ):
+            raise ValidationError(_("Cannot change the product assigned to this order"))
+
+    def _check_same_lease(self, old_instance) -> None:
+        if old_instance.lease and self.lease != old_instance.lease:
+            raise ValidationError(
+                _("Cannot change the lease associated with this order")
+            )
```
igordavydsson, line 317: 🙈  looking at this two and thinking more about it. it seems that the way `_check_same_lease` is safer, since with the check in `_check_same_product` you can now unlink some product from the order 🤔

a-rmz: Hmm. Thinking about it, I think the `_check_same_product` is safer, because on `_check_same_lease` you could set the `lease` to `None` (since we're not checking for `self.lease`).

igordavydsson: exactly why `_check_same_lease` is safer :simple_smile: if an order had a lease, you can't set it to `None` anymore. with the product check you can.

a-rmz: Also, on `_check_same_product`, `old_instance.product` would always be `True` since, product is `null=False`, so there will _always_ be a previous product. And _I think_ `self.product` would also always be `True` since the model constraint checks are run before, so if `None` is being passed, it would raise an error.

igordavydsson: > product is `null=False`
is it? 🤔

a-rmz: AHHHHH NO. My bad, I was confused with `price`.

Also, now I got your point from it being safer. I think we should then also add a check for:
```python
if old_instance.lease and not self.lease:
    raise ValidationError(_("Cannot remove the lease associated with this order"))
```

igordavydsson: but 
```python
if old_instance.lease and self.lease != old_instance.lease:
            raise ValidationError(
                _("Cannot change the lease associated with this order")
            )
```
would already cover this case 🙂

a-rmz: 🤯 Always one step ahead of myself.

payments/models.py
```diff
@@ -154,3 +171,375 @@ def __str__(self):
...
+        # If the product instance is being passed
+        # Price has to be assigned before saving but only if creating
+        if creating and self.product:
+            # Assign the price and tax from the product
+            price = self.product.price_value
+            tax_percentage = self.product.tax_percentage
+
+            # If the lease doesn't last for the whole season
+            if self.lease and not self.is_full_season():
+                price = calculate_product_partial_season_price(
```
igordavydsson, line 411: should this pass `summer_season=False` for WS leases?

payments/utils.py
```diff
@@ -0,0 +1,147 @@
...
+    if start_date.month == end_date.month and start_date.year == end_date.year:
+        price = _calculate_price_for_partial_month(
+            start_date, base_price, days=lambda _: (end_date - start_date).days
+        )
+    else:
+        # Calculate price for the start month
+        price += _calculate_price_for_partial_month(
+            start_date,
+            base_price,
+            days=lambda days_in_month: abs(start_date.day - days_in_month),
```
igordavydsson, line 91: very cool approach with the `abs`, but I think just `lambda days_in_month: days_in_month - start_date.day` would also work :simple_smile:

payments/utils.py
```diff
@@ -0,0 +1,147 @@
...
+            )
+        )
+        # Calculate the added price for all the complete months in between
+        price += len(months_between) * base_price
+
+        # Calculate price for the end month
+        price += _calculate_price_for_partial_month(
+            end_date,
+            base_price,
+            days=lambda days_in_month: abs(days_in_month - start_date.day),
```
igordavydsson, line 109: and here I think you just need to pass the `lambda _: end_date.day`

payments/utils.py
igordavydsson: aaaand probably `lambda days_in_month: days_in_month - start_date.day + 1` since we are including the first day

payments/utils.py
igordavydsson: hm, I kinda feel like here it should be then `lambda _: end_date.day - 1` since prooobably excluding the last day, but maybe we should check with Ville

payments/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    if start_date.month == end_date.month and start_date.year == end_date.year:
+        price = _calculate_price_for_partial_month(
+            start_date, base_price, days=lambda _: (end_date - start_date).days
+        )
+    else:
+        # Calculate price for the start month
+        price += _calculate_price_for_partial_month(
+            start_date,
+            base_price,
+            days=lambda days_in_month: start_date.day - days_in_month + 1,
```
igordavydsson, line 91: well, should be `days_in_month - start_date.day + 1`

a-rmz: True :derp:

payments/models.py
```diff
@@ -154,3 +171,379 @@ def __str__(self):
...
+                        )
+                    elif self.product.period == PeriodType.YEAR:
+                        price = calculate_product_partial_year_price(
+                            price,
+                            self.order.lease.start_date,
+                            self.order.lease.end_date,
+                        )
+                self.price = price
+            if not self.tax_percentage:
+                self.tax_percentage = self.product.tax_percentage
```
igordavydsson, line 525: maybe we could decrease at least some if by doing it like so:
```python

PR #216, opened 2020-06-11
[VEN-398] Orders API

a-rmz:
## Description :sparkles:
- Add queries:
  - `orders(orderType: BERTH | WINTER_STORAGE)`
  - `order(id: ID)`
- Add mutations:
  - `createOrder`
  - `updateOrder`
  - `deleteOrder`
  - `createOrderLine`
  - `updateOrderLine`
  - `deleteOrderLine`

### Closes :no_good_woman:
**[VEN-398](https://helsinkisolutionoffice.atlassian.net/browse/VEN-398):** Expose invoices via GraphQL API

### Related :handshake:
**#214:** [VEN-304] Invoicing models

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_queries.py
```

Example query for `Orders`:
(For more example, check the tests `payments/tests/test_payments_queries.py` or `payments/tests/test_payments_mutations.py`
```graphql
query BP {
  orders {
    edges {
      node {
        id
        customer {
          id
        }
        status
        comment
        price
        taxPercentage
        pretaxPrice
        totalPrice
        totalPretaxPrice
        totalTaxPercentage
        dueDate
        orderLines {
          edges {
            node {
              id
              product {
                id
                priceUnit
                priceValue
                taxPercentage
                service
                period
              }
            }
          }
        }
        logEntries {
          edges {
            node {
              id
            }
          }
        }
        product {
          ... on BerthProductNode {
            id
            priceUnit
            priceValue
            taxPercentage
            priceGroup {
              name
            }
          }
          ... on WinterStorageProductNode {
            id
          }
        }
        lease {
          ... on BerthLeaseNode {
            id
          }
        }
      }
    }
  }
}
```

## Screenshots :camera_flash:
**Orders queries:**
image

**Orders mutations:**
image
image
image

This still depends on #214 to be merged

[VEN-304]: https://helsinkisolutionoffice.atlassian.net/browse/VEN-304


payments/schema/types.py
```diff
@@ -138,3 +149,90 @@ def resolve_product_type(self, info):
...
+    tax_percentage = graphene.Decimal(required=True)
+    pretax_price = graphene.Decimal(required=True)
+    total_price = graphene.Decimal(required=True)
+    total_pretax_price = graphene.Decimal(required=True)
+    total_tax_percentage = graphene.Decimal(required=True)
+    due_date = graphene.Date(required=True)
+    order_lines = DjangoConnectionField("payments.schema.OrderLineNode", required=True)
+    log_entries = DjangoConnectionField(
+        "payments.schema.OrderLogEntryNode", required=True
+    )
```
igordavydsson, line 181: I would probably init those classes before `OrderNode` to avoid string refs

payments/schema/types.py
```diff
@@ -138,3 +149,90 @@ def resolve_product_type(self, info):
...
+    @classmethod
+    @view_permission_required(Order)
+    def get_queryset(cls, queryset, info):
+        return super().get_queryset(queryset, info)
+
+    def resolve_order_lines(self, info, **kwargs):
+        return OrderLine.objects.filter(order=self)
+
+    def resolve_log_entries(self, info, **kwargs):
+        return OrderLogEntry.objects.filter(order=self)
```
igordavydsson, line 203: do you really need these resolvers? 🤔 isn't graphene `magic` enough to resolve them itself?

payments/schema/types.py
```diff
@@ -138,3 +149,90 @@ def resolve_product_type(self, info):
...
+
+
+class OrderLineNode(DjangoObjectType):
+    product = graphene.Field("payments.schema.AdditionalProductNode")
```
igordavydsson, line 207: why is this a string import? 🤔

payments/schema/mutations.py
```diff
@@ -241,6 +247,115 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+            info, input.pop("customer_id"), ProfileNode, nullable=False
+        )
+        product_id = input.pop("product_id", None)
+        if product_id:
+            product = None
+            try:
+                product = get_node_from_global_id(
+                    info, product_id, BerthProductNode, nullable=True
+                )
+            # If a different node type is received, it get_node raises an assertion error
```
igordavydsson, line 286: `, it ... ` seems to be a dangling typo 👀

payments/schema/mutations.py
```diff
@@ -356,6 +363,76 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+class OrderLineInput:
+    quantity = graphene.Int(description="Defaults to 1")
+
+
+class CreateOrderLineMutation(graphene.ClientIDMutation):
+    class Input(OrderLineInput):
+        order_id = graphene.ID(required=True)
+        product_id = graphene.ID(required=True)
+
+    order_line = graphene.Field(OrderLineNode)
```
igordavydsson, line 375: hm, maybe it would make sense to also return the order as a separate field? since you can't traverse it back from the order line

payments/schema/mutations.py
```diff
@@ -356,6 +363,76 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        except (ValidationError, IntegrityError) as e:
+            raise VenepaikkaGraphQLError(e)
+        return CreateOrderLineMutation(order_line=order_line)
+
+
+class UpdateOrderLineMutation(graphene.ClientIDMutation):
+    class Input(OrderLineInput):
+        id = graphene.ID(required=True)
+
+    order_line = graphene.Field(OrderLineNode)
```
igordavydsson, line 402: same question with the order as return field 🤔

payments/schema/types.py
a-rmz: I was thinking about doing _all_ the imports as string imports, but I guess it makes sense to use "normal" imports when they're on the same file.

igordavydsson: "normal" import is the default import, as it helps you traverse the codebase by "Go To definition" etc. and makes refactoring a tiny bit easier. string imports are reasonable when you want to avoid circular import errors

customers/schema.py
```diff
@@ -106,6 +107,7 @@ class BaseProfileFieldsMixin:
     boats = DjangoConnectionField(BoatNode)
     berth_applications = DjangoConnectionField(BerthApplicationNode)
     berth_leases = DjangoConnectionField(BerthLeaseNode)
+    orders = DjangoConnectionField("payments.schema.OrderNode")
```
igordavydsson, line 110: so this could maybe be normal import as well? 🙂

leases/schema.py
```diff
@@ -41,6 +41,7 @@ class BerthLeaseNode(DjangoObjectType):
     berth = graphene.Field(BerthNode, required=True)
     status = LeaseStatusEnum(required=True)
     customer = graphene.Field("customers.schema.ProfileNode", required=True)
+    order = graphene.Field("payments.schema.OrderNode")
```
igordavydsson, line 44: and this? 👀

payments/schema/types.py
a-rmz: Good catch! Smart smart, I didn't know if the filtering was gonna work, but it did. 👍

igordavydsson: it does not really filter, it just execute `self.<field_name>.all()` which is == to `self.order_lines.all()` and `self.log_entries.all()`. so as long as the field_name == related_name on the FK then it should be _magic_ 🧙‍♂️

payments/schema/types.py
a-rmz: I can't remember on which docs I read it (either python-graphene or Django) that the recommended way to handle `Field` (or `ForeignKey` 🤔 ) references was through string. I'll keep looking for it, but we can go on with the "normal" ones.

payments/schema/mutations.py
a-rmz: Hmmm true. We can add an `order` return field and remove later if its' not necessary. 👍

a-rmz: But :big-hm:, for `createOrderLine`, `order` is a required field, so, would it make sense to also return it since you'd in theory know which order it belongs to?

igordavydsson: hm, well, you are kinda mutating Order by adding more lines to it and maybe you'll want to refresh some of the data displayed... 🤔  I see some potential use for it on the client's side, but in theory you could just resend the `order(id: <id>)` query, but not sure about race conditions for it with the mutation. seems like giving the client a possibility to fetch the order as return field can have some use still.

payments/schema/types.py
a-rmz: Aaaahh, alright, makes sense. 👍

payments/schema/mutations.py
a-rmz: Ok ok, I guess there's no harm in having it, we can just remove it if it's not used anymore.

payments/schema/types.py
a-rmz: Couldn't find the docs I was looking for :sad-kitten: so let's go on with the "normal" imports until proven otherwise.

a-rmz: But :hm: (again), we were already doing it on the other types (which of course can also be changed).

https://github.com/City-of-Helsinki/berth-reservations/blob/ven-398/invoicing-api/payments/schema/types.py#L63
```python
class BerthPriceGroupNode(DjangoObjectType):
    name = graphene.String(required=True)
    products = DjangoConnectionField("payments.schema.BerthProductNode", required=True)
    default_product = graphene.Field(
        "payments.schema.BerthProductNode",
        description="Returns the first product associated to the price group "
        "that does not belong to any specific `Harbor`.",
    )
```

igordavydsson: I did not comment there about the import format, because I thought that it was due to circular import. If you can have a normal impor there, then we should refactor it at some point. 🙂

a-rmz: Ahh, fair point. I'll refactor the ones that won't cause import errors.

leases/schema.py
a-rmz: This one causes some circular import errors, so better keep it there.

payments/schema/mutations.py
```diff
@@ -241,6 +247,115 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+            info, input.pop("customer_id"), ProfileNode, nullable=False
+        )
+        product_id = input.pop("product_id", None)
+        if product_id:
+            product = None
+            try:
+                product = get_node_from_global_id(
+                    info, product_id, BerthProductNode, nullable=True
+                )
+            # If a different node type is received it get_node raises an assertion error
```
igordavydsson, line 286: is `it` here redundant?

a-rmz: True 😅

PR #218, opened 2020-06-23
VEN-662 | Sort profiles berthApplications by created_at

santtul:
## Description :sparkles:
- Sort profiles berthApplications by created_at ASC as default
- Add `orderBy` param to profiles berthApplications

### Closes :no_good_woman:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-662

### Automated tests :gear:️
`test_customers_queries.py::test_query_extended_profile_applications_order_by_created`
Fire a query against `graphql_v2`
```
query { 
    berthProfiles { 
        edges { 
            node { 
              id
              comment
              berthApplications {
                edges {
                  node {
                    id
                    createdAt
                  }
                }
              }
            }
        }
    }
}
```
You should see that berthApplications are sorted by createdAt ascending.

Modify query, add param `berthApplications(orderBy: "-createdAt")`, you should see the applications in createdAt descending order.

### DIFF
--- a/customers/schema.py
+++ b/customers/schema.py
@@ -1,3 +1,4 @@
+import django_filters
 import graphene
 from django.core.exceptions import ValidationError
 from django.db import transaction
@@ -88,6 +89,13 @@ PROFILE_NODE_FIELDS = (
 )
 
 
+class BerthApplicationFilter(django_filters.FilterSet):
+    order_by = django_filters.OrderingFilter(
+        fields=(("created_at", "createdAt"),),
+        label="Supports `createdAt` and `-createdAt`.",
+    )
+
+
 class BaseProfileFieldsMixin:
     """
     Mixin that stores the attributes that are exactly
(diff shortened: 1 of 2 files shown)


customers/schema.py
```diff
@@ -107,6 +107,11 @@ class BaseProfileFieldsMixin:
     berth_applications = DjangoConnectionField(BerthApplicationNode)
     berth_leases = DjangoConnectionField(BerthLeaseNode)
 
+    def resolve_berth_applications(self, info, **kwargs):
+        qs = BerthApplication.objects
+
+        return qs.filter(customer=self.id).order_by("created_at")
```
igordavydsson, line 113: you don't haave to initiate a totally new queryset with this `filter(customer=self.id)`, you can just do `return self.berth_applications.order_by("created_at")`. reverse FK relationships are available at this level, and in fact without this custom resolver that it exactly what graphene does: `return self.<field_name>.all()`  (or in this case `return self.berth_applications.all()`)

a-rmz:
The idea for this was under my suggestion to be consistent with [VEN-514](https://helsinkisolutionoffice.atlassian.net/browse/VEN-514):

> _**Backend: sort berth applications and leases query by the created_at field**
> Ascending order as default_

customers/schema.py
santtul: 👍 will try it out

santtul: works like a charm

igordavydsson:
Preferrably unit tests come in the same commit as the changes being tested. In this case, i think, we can squash all commits into one. 🙂

igordavydsson:
This way it's easier to track what is being tested and also easier to see from the tests what the change does.

santtul:
ok, will squash them!

santtul:
done.

PR #220, opened 2020-06-24
[Hotfix] Season date calculation

a-rmz:
## Description :sparkles:
The function that was being used to calculate the beginning of the season was used both for determine the lease start date but also to just calculate the season start. However, the default start date for a lease should be `today` if `season start < today < season end`, so there were some miscalculations by using the same function.

By adding a `use_default_date` param to the function, we can decide whether to use `today` (given the conditions) or the computed `default` start date.

### Automated tests :gear:️
```shell
$ pytest
```

### DIFF
--- a/leases/utils.py
+++ b/leases/utils.py
@@ -1,28 +1,38 @@
 from datetime import date
 
 
-# If a lease object is being created before 10.6, then the dates are in the same year.
-# If the object is being created between those dates, then the start date is
-# the date of creation and end date is 14.9 of the same year.
-# If the object is being created after 14.9, then the dates are from next year.
-def calculate_berth_lease_start_date():
-    # Leases always start on 10.6 the earliest
+def calculate_season_start_date(lease_start: date = None) -> date:
+    """Return the date when the summer season starts
+
+    If the current date is after the season end date,
+    it returns next year's default start date.
+
+    Leases always start on 10.6 the earliest
+    """
     today = date.today()
     default = date(day=10, month=6, year=today.year)
+
+    if lease_start:
+        return default.replace(year=lease_start.year)
 
     # If today is gte than the date when all the leases end,
     # return the default start date for the next year
     if today >= date(day=14, month=9, year=today.year):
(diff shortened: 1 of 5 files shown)


leases/utils.py
```diff
@@ -1,11 +1,19 @@
 from datetime import date
 
 
-# If a lease object is being created before 10.6, then the dates are in the same year.
-# If the object is being created between those dates, then the start date is
-# the date of creation and end date is 14.9 of the same year.
-# If the object is being created after 14.9, then the dates are from next year.
-def calculate_berth_lease_start_date():
+def calculate_berth_lease_start_date(use_default_date: bool = False) -> date:
```
santtul, line 4: Could we have two separate functions: `calculate_season_start_date` and `calculate_berth_lease_start_date`? Then there would be no need for this flag param.

a-rmz: :hm: yeah, that could be a cleaner approach 🤔

igordavydsson: I like @santtul 's suggestions :simple_smile:

leases/utils.py
```diff
@@ -1,12 +1,14 @@
...
-def calculate_berth_lease_start_date():
-    # Leases always start on 10.6 the earliest
+def calculate_season_start_date() -> date:
+    """Return the date when the summer season starts
+
+    If the current date is after the season end date,
+    it returns next year's default start date.
+
+    Leases always start on 10.6 the earliest
+    """
```
igordavydsson, line 11: I think this util should be more flexible in terms of the year. It should probably accept a date as parameter and take the year from that date, because in a similar case to the case where I found this bug - creating order for last year's lease (or for the next year's lease for that matter) - it would today's year and thus `is_full_season` would be `False` again 🤔

a-rmz: I guess there should be a similar behavior for `calculate_season_end_date`? Also allowing to pass a date to base the year.

igordavydsson: yep

PR #226, opened 2020-06-26
VEN-726 | Get WS Applications

santtul:
## Description :sparkles:
Schema for following Queries
- winterStorageApplication
- winterStorageApplications

### Resolves
**[VEN-734](https://helsinkisolutionoffice.atlassian.net/browse/VEN-734)** 

### Automated tests :gear:️
`test_applications_new_schema_ws_queries.py`

Execute query via console:
```
query {
  winterStorageApplications(first: 10) {
    count
    edges {
      node {
        id
        customer {
          id
          __typename
        }
        createdAt
        boatType
        winterStorageAreaChoices {
          winterStorageArea
          winterStorageAreaName
          priority
        }
        __typename
      }
      __typename
    }
    __typename
  }
}
```


applications/new_schema/types.py
```diff
@@ -87,3 +93,62 @@ def resolve_harbor_choices(self, info, **kwargs):
     @view_permission_required(BerthApplication, BerthLease, CustomerProfile)
     def get_node(cls, info, id):
         return super().get_node(info, id)
+
+
+WinterStorageMethodEnum = graphene_enum(WinterStorageMethod)
```
igordavydsson, line 98: nitpick: I would keep all enum declarations in the beginning of the file together, as they are one-liners that might easily get lost between bigger blocks of code. makes it easier to navigate things in github's web UI, if you need to look something up

applications/new_schema/types.py
```diff
@@ -87,3 +93,62 @@ def resolve_harbor_choices(self, info, **kwargs):
...
+
+
+class WinterStorageAreaChoiceType(DjangoObjectType):
+    winter_storage_area = graphene.String()
```
igordavydsson, line 102: why isn't this `required=True`? 🤔

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,407 @@
...
+        }
+    }
+}
+"""
+
+
+def create_handled_application(winter_storage_application) -> None:
+    winter_storage_application.lease = WinterStorageLeaseFactory()
+    winter_storage_application.status = ApplicationStatus.HANDLED
+    winter_storage_application.save()
```
igordavydsson, line 100: might make sense to just create a fixture out of it, it could look like this if you put it after your `winter_storage_application` fixture:
```python
@pytest.fixture
def handled_ws_application(winter_storage_application):
    winter_storage_application.status = ApplicationStatus.HANDLED
    winter_storage_application.save()
    return winter_storage_application
```

applications/tests/factories.py
```diff
@@ -36,6 +36,7 @@ class Meta:
 
 class WinterStorageApplicationFactory(BaseApplicationFactory):
     storage_method = WinterStorageMethod.ON_TRAILER
+    municipality = "Helsinki"
```
igordavydsson, line 39: checked this out. actually field `municipality` on the abstract `BaseApplication` model is just a `CharField` and not a `ForeignKey` to `django-munigeo`'s `Municipality` model. I do not know why `municipality = factory.SubFactory(MunicipalityFactory)` works for `BerthApplicationFactory`, that would need a deeper dive into this, but I think we should change it to `municipality = factory.Faker("word")` on `BaseApplicationFactory` to make it consistent with what the model expects.

applications/new_schema/types.py
santtul: this maps to `winter_storage_area.servicemap_id` (like in `HarborChoiceType -> harbor`), but the field is nullable which causes api call to fail in case of nulls.

applications/new_schema/types.py
santtul: ok! will move it

applications/new_schema/types.py
igordavydsson: ah, yes, true, WS areas do not have servicemap ids. this is ok then, we will change this later, if needed.

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,394 @@
...
+
+
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_services", "berth_handler", "berth_supervisor"],
+    indirect=True,
+)
+def test_winter_storage_applications_statuses_filter_invalid_enum(
+    handled_ws_application, api_client
+):
```
a-rmz, line 152: For cases like this when you're not necessarily testing that the user has the right permissions, you can use the `superuser_api_client`, you save a couple of unnecessary tests and emphasize that the "fail" you're checking for doesn't directly concern user permissions.

applications/new_schema/types.py
```diff
@@ -87,3 +95,59 @@ def resolve_harbor_choices(self, info, **kwargs):
     @view_permission_required(BerthApplication, BerthLease, CustomerProfile)
     def get_node(cls, info, id):
         return super().get_node(info, id)
+
+
+class WinterStorageAreaChoiceType(DjangoObjectType):
+    winter_storage_area = graphene.String(required=True)
```
igordavydsson, line 101: wait, but you said that this causes errors 😅

igordavydsson: and I said, that "yes, servicemap id are not there, so we can keep this non-required" 🙂

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,394 @@
...
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_services", "berth_handler", "berth_supervisor"],
+    indirect=True,
+)
+def test_winter_storage_application_query(winter_storage_application, api_client):
+    first_choice = WinterAreaChoiceFactory()
+    first_choice.priority = 1
+    first_choice.application = winter_storage_application
+    first_choice.save()
```
a-rmz, line 365: You can pass those values directly to the factory so you don't have to save it later.
```suggestion
    first_choice = WinterAreaChoiceFactory(priority=1, application=winter_storage_application)
```

applications/new_schema/types.py
santtul: when re-testing it,  I accidentally committed it. good catch, will remove :D

applications/tests/test_applications_new_schema_ws_queries.py
santtul: good point, I change it.

applications/tests/test_applications_new_schema_ws_queries.py
santtul: sweet, I'll change it

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,389 @@
...
+        "invalid value [%s]." % nonexistent_enum_str, executed,
+    )
+
+
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_services", "berth_handler", "berth_supervisor"],
+    indirect=True,
+)
+def test_winter_storage_applications_statuses_filter_empty_list(
```
a-rmz, line 166: You can also use the `superuser_api_client` here. Since you have already tested that the query _works_, and this is just a variation of the query, it doesn't matter if you don't explicitly check for all the possible user groups.

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,389 @@
...
+
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_services", "berth_handler", "berth_supervisor"],
+    indirect=True,
+)
+@pytest.mark.parametrize(
+    "order_by,ascending", [("createdAt", True), ("-createdAt", False)]
+)
+def test_winter_storage_applications_order_by_created(order_by, ascending, api_client):
```
a-rmz, line 214: Same here

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,389 @@
...
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_services", "berth_handler", "berth_supervisor"],
+    indirect=True,
+)
+def test_winter_storage_application_query(winter_storage_application, api_client):
+    first_choice = WinterAreaChoiceFactory(
+        priority=1, application=winter_storage_application
+    )
+    first_choice.save()
```
igordavydsson, line 360: you don't need to save these. when you init an object with a factory, it is a fully functioning model instance

applications/tests/test_applications_new_schema_ws_queries.py
```diff
@@ -0,0 +1,389 @@
...
+}
+"""
+
+
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_services", "berth_handler", "berth_supervisor"],
+    indirect=True,
+)
+def test_winter_storage_applications_order_by_created_at_default(api_client):
```
a-rmz, line 257: Samesies

PR #227, opened 2020-06-30
VEN-726 | Choice ordering

santtul:
## Description :sparkles:
- Return choices in priority order
### Related :handshake:
**[VEN-726](https://helsinkisolutionoffice.atlassian.net/browse/VEN-726)** 

### Automated tests :gear:️
`applications/tests/test_applications_new_schema_berth_queries.py::test_winter_storage_application_query`


applications/models.py
```diff
@@ -31,6 +32,7 @@ class WinterStorageAreaChoice(models.Model):
 
     class Meta:
         unique_together = ("application", "priority")
+        ordering = ("application", "priority")
```
igordavydsson, line 35: I think, this should cause new migration files 🤔

a-rmz: I'm not sure about it 🤔 🤔 because that's on the query level (i.e. the SQL that django builds) rather than on the DB schema.

a-rmz: What I'm not sure about is `application` as an ordering key, since Application is a FK. Unless it would use the internal `int` id that django generates.

igordavydsson: yes, Django will use that `int`

santtul: You are correct 👍  will generate them.

santtul: Added the generated migration file. This seems to work in SQL level in a way that it generates following order clause: `ORDER BY "applications_winterstorageareachoice"."application_id" ASC, "applications_winterstorageareachoice"."priority" ASC` which makes sense.

PR #228, opened 2020-07-02
VEN-587 | CRUD mutation for Boats

a-rmz:
## Description :sparkles:
* Split the `customers/schema.py` into a `customers/schema/` directory with split files
* Add mutations to create/update/delete `Boats`
* Remove the mutations to CRUD `BoatCertificates` and replace it with fields on the `Boat` mutations
```
createBoat.addBoatCertificates = graphene.List(BoatCertificateInput)
updateBoat.addBoatCertificates = graphene.List(BoatCertificateInput)
updateBoat.removeBoatCertificates = graphene.List(graphene.ID)
```

### Closes :no_good_woman:
**[VEN-587](https://helsinkisolutionoffice.atlassian.net/browse/VEN-587):** Add CRUD mutations for Boats

### Automated tests :gear:️
```shell
$ pytest customers/tests/test_customers_mutations.py
```

To keep the PR short, this is one example. For more examples, check the tests (`customers/tests/test_customers_mutations.py`)
```graphql
mutation CREATE_BOAT($input: CreateBoatMutationInput!) {
    createBoat(input: $input) {
        boat {
            owner {
                id
            }
            boatType {
                id
            }
            name
            length
            width
            certificates {
                id
                certificateType
                checkedAt
                checkedBy
                validUntil
                file
            }
        }
    }
}
```

```json
{
  "input":{
    "ownerId":"UHJvZmlsZU5vZGU6ZDE3ZDgzNmMtZTJiNS00NjBjLTg2NzUtMjdmOTE0ZWRiMTJl",
    "boatTypeId":"1",
    "name":"Flying Dutchman",
    "length":"15.0",
    "width":"33.3",
    "registrationNumber":"123",
    "model":"Pirate ship",
    "addBoatCertificates":[
      {
        "certificateType":"INSURANCE"
      }
    ]
  }
}
```
image

## Screenshots :camera_flash:
**Create Boat:**
image

**Update Boat:**
image

**Delete Boat:**
image


resources/tests/test_resources_mutations.py
```diff
@@ -497,7 +498,7 @@ def test_update_berth_not_enough_permissions(api_client, berth, pier, berth_type
     }
     assert Berth.objects.count() == 1
 
-    executed = api_client.execute(CREATE_BERTH_MUTATION, input=variables)
+    executed = api_client.execute(UPDATE_BERTH_MUTATION, input=variables)
```
igordavydsson, line 501: :derp:

a-rmz: He he he.

customers/schema/mutations.py
```diff
@@ -1,110 +1,167 @@
...
     valid_until = graphene.Date()
     checked_at = graphene.Date()
     checked_by = graphene.String()
 
 
-class CreateBoatCertificateMutation(graphene.ClientIDMutation):
-    class Input(BoatCertificateInput):
-        boat_id = graphene.ID(required=True)
-        certificate_type = BoatCertificateTypeEnum(required=True)
+class BoatInput:
```
igordavydsson, line 45: not that it has much effect, but should this also extend `graphene.InputObjectType`?

customers/schema/mutations.py
```diff
@@ -1,110 +1,167 @@
...
-        return CreateBoatCertificateMutation(boat_certificate=certificate)
+        return CreateBoatMutation(boat=boat)
 
 
-class UpdateBoatCertificateMutation(graphene.ClientIDMutation):
-    class Input(BoatCertificateInput):
+class UpdateBoatMutation(graphene.ClientIDMutation):
+    class Input(BoatInput):
         id = graphene.ID(required=True)
+        remove_boat_certificates = graphene.List(graphene.ID)
```
igordavydsson, line 98: we also need to cover the `update_boat_certificates` use case. when updating, `certificate_type` is non-required, but when adding certs, it should be required. so, might have to split `BoatCertificateInput` in 2: `CreateBoatCertificateInput` and `UpdateBoatCertificateInput` (this latter one could extend the former one and just set `required=True` for the `certificate_type`. just a quick suggestion, there might be other ways to implement this.

customers/tests/conftest.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 
 from berth_reservations.tests.conftest import *  # noqa
+from resources.tests.conftest import boat_type  # noqa
```
igordavydsson, line 4: I think `# noqa` is not needed, since the line above has it, because default isort config complains about wildcard imports

customers/tests/test_customers_mutations.py
```diff
@@ -38,48 +48,66 @@
...
+    boat_type_id = str(boat_type.id)
 
     variables = {
-        "boatId": boat_id,
-        "file": SimpleUploadedFile(
-            name=file_name, content=None, content_type="application/pdf"
-        ),
-        "certificateType": random.choice(list(BoatCertificateType)).name,
-        "validUntil": check_date,
-        "checkedAt": check_date,
```
igordavydsson, line 54: how did it work like this before 🤯 🤔

customers/tests/test_customers_mutations.py
```diff
@@ -38,48 +48,66 @@
...
-        "checkedBy": "John Wick",
+        "ownerId": owner_id,
+        "boatTypeId": boat_type_id,
+        "name": "Flying Dutchman",
+        "length": Decimal(random.randrange(1, 99)),
```
igordavydsson, line 61: we do `round(Decimal(random.randrange(1, 99)), 1)` or `round(Decimal(random.randrange(1, 99)), 2)` in other places 🤔 should we be more consistent 🤔 should we wrap this into a util 🤔

customers/schema/mutations.py
a-rmz: Nope. The `mutation.Input` automatically extends `InputObjectType`, so `BoatInput` doesn't have to.
```python
class CreateBoatMutation(graphene.ClientIDMutation):
    class Input(BoatInput):
        pass
```
If you extend `InputObjectType` both on `BoatInput` and then `mutation.Input`, you get a resolution problem because python doesn't know the hierarchy of the extends.

```
TypeError: Cannot create a consistent method resolution
order (MRO) for bases InputObjectType, Input
```

`BoatCertificateInput` requires extending the `InputObjectType` because it's used for input but it doesn't directly extend the input superclass.

customers/tests/conftest.py
a-rmz: It is, because of the `unused import` error.

customers/tests/test_customers_mutations.py
a-rmz: Huh? 🤔

customers/tests/test_customers_mutations.py
a-rmz: True, would be nicer to have a single util for all this. There's `payments.utils.round_price`, but definitely worth extracting. Good catch. 👍

customers/schema/mutations.py
igordavydsson: ahhh, gotcha! 👍

customers/tests/conftest.py
igordavydsson: ah, right-right-right! 👍

customers/tests/test_customers_mutations.py
igordavydsson: these are date objects. how did they equal to strings like `"2020-07-02"`? this is what we faced now when fixing the tests

a-rmz: Aaahh, yeah. It wasn't working before, it was _another_ of the failures that you helped me fix yesterday.

igordavydsson: nono, it was working when it was a test for `createBoatCertificate` 🤔  which is why I am wondering, how did it work

utils/numbers.py
```diff
@@ -0,0 +1,32 @@
...
+    :param as_string: Whether the value should be returned as Decimal or string.
+    For testing, it's useful to be able to parse the Decimal value directly to string.
+    :return: A rounded Decimal number
+    """
+    number = round(Decimal(value), decimals)
+
+    return str(number) if as_string else number
+
+
+def random_int(min: int = 0, max: int = 100, as_string: bool = False) -> DecimalString:
```
igordavydsson, line 25: `random_int` that returns either `Decimal` or `string` :big-hm:

igordavydsson: if it's not needed yet, maybe we can skip it for now

a-rmz: Well, the API tests return `Decimal` values as string, that's why I thought it would be more convenient to have it wrapped in the same function to simplify testing.

igordavydsson: yeah, but that's why you wrote `random_decimal` 🙂  I mean this util is called `random_int` but does not return an int of really, a bit misleading

a-rmz: Well yeah, returns a `Decimal` object which wraps an _integer_ number (i.e. no decimals and no point, e.g. `Decimal("42")`).

If you think it's misleading, we can just leave `random_decimal` and pass `decimals=0` when we'd need an "int".

igordavydsson: I think we do not need any Decimals in random int generation logic at all. just the following would suffice:
```python
def random_int(min: int = 0, max: int = 100, as_string: bool = False) -> Union[Int, String]:
    integer = randint(min, max)
    return str(integer) if as_string else integer
```

a-rmz: In that case, yes. In any case, it's not being used at the moment 😛 so we can just pretend this conversation never happened.

PR #230, opened 2020-07-13
VEN-674 | Add django-ilmoitin queries to API

a-rmz:
## Description :sparkles:
* Update the `django-ilmoitin` version to the latest
* Add the `notificationTemplates` query to the API

### Closes :no_good_woman:
**[VEN-674](https://helsinkisolutionoffice.atlassian.net/browse/VEN-XXX):** Existing notification templates could be queried via GQL API

```graphql
query Notifications {
  notificationTemplates {
    edges {
      node {
        id
        preview
        translations {
          subject
          languageCode
          preview
          bodyText
        }
      }
    }
  }
}
```

### DIFF
--- a/berth_reservations/new_schema.py
+++ b/berth_reservations/new_schema.py
@@ -1,4 +1,6 @@
+import django_ilmoitin.api.schema as django_ilmoitin_schema
 import graphene
+from django_ilmoitin.models import NotificationTemplate
 from graphene_federation import build_schema
 
 import applications.new_schema
@@ -6,6 +8,7 @@ import customers.schema
 import leases.schema
 import payments.schema
 import resources.schema
+from users.decorators import view_permission_required
 
 # =====================================================
 # New GraphQL API for the new 'resources' app
@@ -21,9 +24,15 @@ class Query(
     leases.schema.Query,
     resources.schema.Query,
     payments.schema.Query,
+    django_ilmoitin_schema.Query,
     graphene.ObjectType,
 ):
-    pass
+    @staticmethod
+    @view_permission_required(NotificationTemplate)
+    def resolve_notification_templates(parent, info, **kwargs):
+        return django_ilmoitin_schema.Query.resolve_notification_templates(
+            parent, info, **kwargs
+        )
 
 
 class Mutation(
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ django-filter==2.2.0      # via -r requirements.in
 django-graphql-geojson==0.1.4  # via -r requirements.in
 django-graphql-jwt==0.2.2  # via -r requirements.in
 django-helusers==0.5.2    # via -r requirements.in
-django-ilmoitin==0.2.0    # via -r requirements.in
+django-ilmoitin==0.5.2    # via -r requirements.in
 django-js-asset==1.2.2    # via django-mptt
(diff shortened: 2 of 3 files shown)


igordavydsson:
what about authentication? is this query now open to non-authenticated users?

a-rmz:
> what about authentication? is this query now open to non-authenticated users?

As per `django-ilmoitin` it is, it doesn't have any permission checking, but we can try to find a way to restrict those queries for admins only.

igordavydsson:
yeah, kinda forgot about that part. could you investigate possible solutions for this? i think ideallly this would follow the same permission checks as other queries in our backend. I think only the people with the permission to `Send messages` should be allowed to fetch the templates.

a-rmz:
One first approach could be:

```python
class ProtectedIlmoitinQuery(django_ilmoitin_schema.Query):
    @view_permission_required(NotificationTemplate)
    def resolve_notification_templates(self, info, **kwargs):
        return NotificationTemplate.objects.all()

class Query(
    # ...
    ProtectedIlmoitinQuery,
    graphene.ObjectType,
):
    pass
```

So basically, manually overriding/extending the resolvers that need to be protected.

igordavydsson:
hm, I think this looks. like a good approach. you kinda have to know what queries there are if you are hooking permission checks into them, so this is ok IMO. if no other neater / DRYer solutions come to mind, we can go with this 👍

a-rmz:
Another option would be to somehow pass the required permissions to some kind of `Meta` class and enforce those permissions on the `ilmoitin` level, but that would require:
* more research/work
* add the permission checking functions to the ilmoitin/city utils package
* good luck because I'm not 100% sure it would be easy or straightforward

a-rmz:
Regarding the first approach, I'm not sure about completely overwriting the return value of the `resolve_xxxx`, since the (query) fields are using `DjangoConnectionField`, and by overwriting them we are throwing that off the board.

Just thinking out loud now.

igordavydsson:
🤔  maybe we can also call `super`:
```python
class ProtectedIlmoitinQuery(django_ilmoitin_schema.Query):
    @view_permission_required(NotificationTemplate)
    def resolve_notification_templates(self, info, **kwargs):
        return super().resolve_notification_templates(info, **kwargs)
```
this way we ensure that return value stays the same

igordavydsson:
yeah, would be neat to have that somehow in `Meta`, but not sure it after all it would seem pythonic. I think that extending and hooking permission check is quite ok 👍

a-rmz:
> 🤔 maybe we can also call `super`:
> 
> ```python
> class ProtectedIlmoitinQuery(django_ilmoitin_schema.Query):
>     @view_permission_required(NotificationTemplate)
>     def resolve_notification_templates(self, info, **kwargs):
>         return super().resolve_notification_templates(info, **kwargs)
> ```
> 
> this way we ensure that return value stays the same

I tried this too, but the resolver is not directly implemented on the `super` class (some graphene magic in between) so `super().resolve_notification_templates()` is not defined.

igordavydsson:
oh, yeah, there is that dynamic resolver, `super` won't work right away, only if you add the resolver first [here](https://github.com/City-of-Helsinki/django-ilmoitin/blob/develop/django_ilmoitin/api/schema/queries.py) 🤔

PR #234, opened 2020-07-16
VEN-759 | Default season dates for winter leases

a-rmz:
## Description :sparkles:
* Add a default date for "winter" season start-end

### Closes :no_good_woman:
**[VEN-759](https://helsinkisolutionoffice.atlassian.net/browse/VEN-759):** add season logic to winter storage leases

### Related :handshake:
**[VEN-671](https://helsinkisolutionoffice.atlassian.net/browse/VEN-671):**  schema for Winter Storage Leases

### Automated tests :gear:️
```shell
pytest leases/tests/test_lease_models.py::test_winter_storage_lease_model
```


leases/models.py
```diff
@@ -183,6 +188,12 @@ class WinterStorageLease(AbstractLease):
         null=True,
         related_name="lease",
     )
+    start_date = models.DateField(
```
igordavydsson, line 191: we could also maybe delete the fields from the abstract base model? now that every actual model has its own implementation

leases/utils.py
```diff
@@ -63,3 +63,30 @@ def calculate_berth_lease_end_date() -> date:
...
+def calculate_winter_storage_lease_start_date() -> date:
+    """
+    Return the date when the "winter" lease season should start
+
+    The season should start by default on 15.9 (one day after the "summer" season ends)
+    """
+
+    # Otherwise, return the latest date between the default start date or today
+    today = date.today()
+    default = max(date(day=15, month=9, year=today.year), today)
```
igordavydsson, line 77: this is good logic for the period from 10.6 till 15.9, i.e. during the summer season. however from 1.1 till 10.6 it should still use today, I think. otherwise, you might end up with a ws lease that start e.g. 15.9.2020, but ends 10.6.2020.

leases/utils.py
```diff
@@ -63,3 +63,30 @@ def calculate_berth_lease_end_date() -> date:
...
+
+def calculate_winter_storage_lease_end_date() -> date:
+    """Return the date when the lease season should end
+
+    The season should end by default on 10.6 on the following year
+    """
+    today = date.today()
+    default = date(day=10, month=6, year=today.year + 1)
+
+    if today.month < default.month:
```
igordavydsson, line 89: and here we probably should also compare the days? otherwise leases generated from 1st till 10th of June will go to the next year 🤔

leases/utils.py
a-rmz: So in this case it would be
```python
if today < date(10, 6, today.year):
    return min(date(day=15, month=9, year=today.year), today)
```

leases/utils.py
a-rmz: 🤔 Tru, good catch

igordavydsson: extreme edge case, but since we noticed, we can account for it already now 😅

a-rmz: Yup yup, better safe than sorry 😛

leases/utils.py
```diff
@@ -63,3 +63,34 @@ def calculate_berth_lease_end_date() -> date:
...
+
+def calculate_winter_storage_lease_end_date() -> date:
+    """Return the date when the lease season should end
+
+    The season should end by default on 10.6 on the following year
+    """
+    today = date.today()
+    default = date(day=10, month=6, year=today.year + 1)
+
+    if today.month < default.month and today.day < default.day:
```
igordavydsson, line 93: hm, should it be
```python
if today.month <= default.month and ...
```
🤔

a-rmz: Right! 😓

leases/utils.py
```diff
@@ -63,3 +63,34 @@ def calculate_berth_lease_end_date() -> date:
...
+
+    The season should start by default on 15.9 (one day after the "summer" season ends)
+    """
+
+    # Return the earliest date between the default start date or today
+    today = date.today()
+    default = max(date(day=15, month=9, year=today.year), today)
+
+    if today < date(10, 6, today.year):
+        return min(default, today)
```
igordavydsson, line 80: hm, `min()` here is kinda redundant, since `default` cannot be `< date(15, 9, today.year)`. not that it affects anything 🤔

a-rmz: 🤔 :derp: Good catch actually, doesn't affect, but it's entirely unnecessary.

PR #241, opened 2020-07-22
VEN-766 | Fix tax percentage calculation

a-rmz:
## Description :sparkles:
* The tax percentage is now calculated properly and is rounded to the nearest 0.05-decimal
* Add a "round to nearest decimal" option to `utils.numbers.rounded`
* Add a test to explicitly test for order's tax percentage
* Tidy up rounding functions and reuse code

### Closes :no_good_woman:
**[VEN-766](https://helsinkisolutionoffice.atlassian.net/browse/VEN-766):** total tax percentage on orders gets wrong value

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_models.py::test_order_tax_percentage
```

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -25,10 +25,10 @@ from payments.utils import (
     calculate_product_partial_year_price,
     calculate_product_percentage_price,
     convert_aftertax_to_pretax,
-    round_price,
     rounded,
 )
 from utils.models import TimeStampedModel, UUIDModel
+from utils.numbers import rounded as rounded_decimal
 
 PLACE_PRODUCT_TAX_PERCENTAGES = [Decimal(x) for x in ("24.00",)]
 ADDITIONAL_PRODUCT_TAX_PERCENTAGES = [Decimal(x) for x in ("24.00", "10.00")]
@@ -268,9 +268,12 @@ class Order(UUIDModel, TimeStampedModel):
         )
 
     @property
-    @rounded
     def total_tax_percentage(self):
-        return ((self.total_price - self.total_pretax_price) / self.total_price) * 100
+        return rounded_decimal(
+            ((self.total_price - self.total_pretax_price) / self.total_pretax_price)
+            * 100,
+            round_to_nearest=0.05,
+        )
 
     def _check_valid_products(self) -> None:
         if self.product and not isinstance(
@@ -434,7 +437,7 @@ class Order(UUIDModel, TimeStampedModel):
                     )
                     price = price * place_sqm
 
-            self.price = round_price(self.price or price)
+            self.price = rounded_decimal(self.price or price)
             self.tax_percentage = self.tax_percentage or tax_percentage
 
         old_instance = Order.objects.filter(id=self.id).first()
--- a/payments/tests/factories.py
+++ b/payments/tests/factories.py
@@ -1,5 +1,4 @@
 import random
-from decimal import Decimal
 
 import factory
 
@@ -25,10 +24,12 @@ from .utils import random_bool, random_price
 
 
 class AbstractBaseProductFactory(factory.django.DjangoModelFactory):
-    price_value = factory.LazyFunction(
-        lambda: round(Decimal(random.uniform(1, 999)), 2)
-    )
     price_unit = factory.Faker("random_element", elements=list(PriceUnits))
+    price_value = factory.LazyAttribute(
+        lambda o: random_price()
+        if o.price_unit == PriceUnits.AMOUNT
+        else random_price(0, 100, decimals=0)
+    )
 
     class Meta:
         model = AbstractBaseProduct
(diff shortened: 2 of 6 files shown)


utils/numbers.py
```diff
@@ -15,9 +18,20 @@ def rounded(
     :param decimals: The amount of decimals to round value
     :param as_string: Whether the value should be returned as Decimal or string.
     For testing, it's useful to be able to parse the Decimal value directly to string.
+    :param round_to_nearest: Rounds the value to the nearest decimal passed.
+    I.e. the nearest 0.05 of 3.1416 -> 3.15
     :return: A rounded Decimal number
     """
-    number = round(Decimal(value), decimals)
+    number = value
```
igordavydsson, line 25: I think we could skip this assignment and just use `value` everywhere. or rename the parameter to `number`

a-rmz: I thought so too but I went the other way to avoid overwriting parameters, but I guess it won't make much of a difference here.

payments/models.py
```diff
@@ -270,7 +270,12 @@ def total_pretax_price(self):
     @property
     @rounded
     def total_tax_percentage(self):
-        return ((self.total_price - self.total_pretax_price) / self.total_price) * 100
+        return rounded_decimal(
```
igordavydsson, line 273: we probably don't need the decorator then?

payments/utils.py
```diff
@@ -22,13 +19,13 @@ def rounded(func):
     @wraps(func)
     def wrapped(*args, **kwargs):
         value = func(*args, **kwargs)
-        return round_price(value)
+        return util_rounded(value, decimals=2)
```
igordavydsson, line 22: are you passing `decimals` explicitly to stress that we will have 2 decimal points in the result? since `2` is default for `decimals` anyway

payments/utils.py
```diff
@@ -1,17 +1,14 @@
...
 
 from dateutil.relativedelta import relativedelta
 from dateutil.rrule import MONTHLY, rrule
 
 from leases.utils import calculate_season_end_date, calculate_season_start_date
-
-
-def round_price(price):
-    return price.quantize(Decimal("0.01"), rounding=ROUND_HALF_UP)
+from utils.numbers import rounded as util_rounded
```
igordavydsson, line 11: should this be `as rounded_decimal` for consistency. and also we are in the `utils.py` file and there is `rounded` decorator function here

payments/models.py
a-rmz: True, not needed indeed 🔥

payments/utils.py
a-rmz: 🤔 good point

payments/utils.py
a-rmz: Makes more sense 👍

payments/tests/test_payments_models.py
```diff
@@ -807,3 +807,7 @@ def test_order_line_pretax_price(order_line):
     assert order_line.pretax_price == round(
         order_line.price / (1 + (order_line.tax_percentage / 100)), 2
     )
+
+
+def test_order_tax_percentage(order):
+    assert order.tax_percentage == Decimal("24.00")
```
igordavydsson, line 813: `order.total_tax_percentage`? but also, we could use a bit more logic in this test, let me work on my idea. brb

igordavydsson: ok, this is what I had in mind, both as a test fore most important parts of our business logic related to orders as well as. a sort of explanations of what we expect:
```python
def test_order_tax_percentage():
    # Create a product for an optional service
    ap = AdditionalProductFactory(
        service=ProductServiceType.PARKING_PERMIT,
        price_unit=PriceUnits.AMOUNT,
    )
    # Ensure tax percentage is 10%, as factory randomly chooses between 24 and 10
    ap.tax_percentage = Decimal("10.00")
    ap.save()

    # Hardcode the base price. Base tax percentage currently is always 24%
    o = OrderFactory(price=Decimal("100.00"))
    # Hardcode the price of additional product to override model's save method
    OrderLineFactory(order=o, product=ap, price=Decimal("88.70"))
    
    # Calculations:
    # base product: pretax = 80.65, tax = 24.0%, tax amount = 19.35
    # order line: pretax = 80.64, tax = 10.0%, tax amount = 8.06
    # total tax should be: (24.0 + 10.0) / 2 == 17.0

    assert o.total_tax_percentage == Decimal("17.00")
```

payments/tests/factories.py
```diff
@@ -68,15 +65,19 @@ class AdditionalProductFactory(AbstractBaseProductFactory):
     period = factory.LazyFunction(lambda: PeriodType.SEASON)
     tax_percentage = factory.LazyFunction(lambda: DEFAULT_TAX_PERCENTAGE)
 
-    # Because of the FIXED_SERVICE restrictions (only allowed to have 24% VAT)
-    # the actual assignment of a random Tax value is done once the service has
-    # been assigned to the model.
     @factory.post_generation
-    def service_taxes(self, created, extracted, **kwargs):
+    def price_value_correction(self, created, extracted, **kwargs):
```
igordavydsson, line 69: should this be just `price_value`? should it also be aware of the `extracted`?

a-rmz: The problem with naming it `price_value` is that it ignores the value assigned to the actual `price_value` by the factory/generator since it's a `post_generation` hook, so when the object is being created, `self.price_value == None` which causes the DB insertion to fail.

igordavydsson: UPD: we went with a `LazyAttribute` instead 🤗

PR #243, opened 2020-07-23
Hotfix | Pretax price breaking Admin page

a-rmz:
## Description :sparkles:
* Check if the value being formatted exists, otherwise return a placeholder value
* Add checks for `total_price` and `total_tax_percentage` on the `payments/admin` to avoid `None` values passed

Go to `admin/payments/order/{order_id}/change/` and `admin/payments/order/add/`, both should work fine

### DIFF
--- a/payments/admin.py
+++ b/payments/admin.py
@@ -103,7 +103,7 @@ class OrderAdmin(admin.ModelAdmin):
 
     @currency
     def total_price(self, obj):
-        return obj.total_price
+        return obj.total_price if obj.price and obj.tax_percentage else None
 
     @currency
     def total_pretax_price(self, obj):
@@ -111,7 +111,7 @@ class OrderAdmin(admin.ModelAdmin):
 
     @percentage
     def total_tax_percentage(self, obj):
-        return obj.total_tax_percentage
+        return obj.total_tax_percentage if obj.price and obj.tax_percentage else None
 
     pretax_price.short_description = _("Pretax price")
     pretax_price.admin_order_field = "pretax_price"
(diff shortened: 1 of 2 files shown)


payments/admin.py
```diff
@@ -103,15 +103,15 @@ def pretax_price(self, obj):
 
     @currency
     def total_price(self, obj):
-        return obj.total_price
+        return obj.total_price if obj.price and obj.tax_percentage else None
```
igordavydsson, line 106: The problem we have in Django admin is that Django generate an empty order line inline (even though we specify `extra = 0`), which of course has no prices. Once that inline tries to populate `pretax_price` (lines 68-70 in this file), decorator `currency` fails. I think that the changes you added now to the `currency_format` and `percentage_format` are good and what we need here. However, all these `if obj.price and obj.tax_percentage`, I feel, should not be here. We should ensure that all the order and order line properties do not fail and either return the value or return `None`, but do not end up in errors.

a-rmz: The reason for those `if obj.price` is different than the one that was making `@currency` fail.
```
Exception Value: unsupported operand type(s) for /: 'NoneType' and 'int'
Exception Location: /berth-reservations/payments/utils.py in convert_aftertax_to_pretax, line 55
```

Since `total_price` is calculated by `sum([ol.price for ol in self.order_lines.all()], self.price)`, the `self.price == None` for those empty orders being generated on the admin. With those checks we ensure that the `total_price` is not being calculated _if_ the order doesn't have a price.

There are two alternatives:
1. Leave it as-is, since those checks only concern the admin, which is the only place where "empty" orders can be created without a price
2. Add a some kind of validation to
```python\

igordavydsson: yep, you are right, after further investigation also seemed like these were needed, let's leave them then 👍

PR #245, opened 2020-07-24
VEN-730 | Profile mutations

a-rmz:
## Description :sparkles:
* Add `CUD` mutations for `ProfileNode`
* Minor fixes for some errors left on `BoatNode` mutations

### Closes :no_good_woman:
**[VEN-730](https://helsinkisolutionoffice.atlassian.net/browse/VEN-730):** add CRUD mutations for CustomerProfile

### Automated tests :gear:️
```shell
$ pytest customers/tests/test_customers_mutations.py
```

You can find examples on `customers/tests/test_customers_mutations.py`

## Screenshots :camera_flash:
**createBerthServicesProfile**
image

**updateBerthServicesProfile**
image

**deleteBerthServicesProfile**
image

The `deleteBerthServicesProfile` mutation takes a `deleteOrganization: boolean` input. The mutation will raise an error if it's passed as `true` and the `profile` does not have an organization or if both `deleteOrganization: true` and `organization: {}` inputs are passed.

### DIFF
--- a/customers/schema/mutations.py
+++ b/customers/schema/mutations.py
@@ -1,6 +1,7 @@
 import graphene
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
+from django.utils.translation import ugettext_lazy as _
 from graphene_file_upload.scalars import Upload
 
 from berth_reservations.exceptions import VenepaikkaGraphQLError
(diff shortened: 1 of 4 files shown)


customers/schema/mutations.py
```diff
@@ -164,6 +172,57 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+class CreateBerthServicesProfileMutation(graphene.ClientIDMutation):
+    class Input(BerthServicesInput):
+        id = graphene.GlobalID(
+            required=True,
+            description="`GlobalID` associated to the Open City Profile customer profile",
+        )
+        organization = OrganizationInput()
+
+    profile = graphene.Field(ProfileNode)
+    organization = graphene.Field(OrganizationNode)
```
igordavydsson, line 199: not sure this is needed, as `organization` is a field of `PRofileNode` anyway

a-rmz: 🤔 true... I was following the example of [payments/schema/mutations.py#L415-L420](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/payments/schema/mutations.py#L415-L420)
```python
class UpdateOrderLineMutation(graphene.ClientIDMutation):
    class Input(OrderLineInput):

    order_line = graphene.Field(OrderLineNode)
    order = graphene.Field(OrderNode)
```
But now that I think more about it, it makes more sense to not do it, since profile contains organization and not the other way around.

igordavydsson: yeah, order can have multiple order lines, so it would require extra effort for the client to find the one that changed, but here only one organization for a profile, so it is rather straightforward.

customers/schema/mutations.py
```diff
@@ -223,6 +224,64 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+                    _("The passed Profile is not associated with an Organization")
+                )
+            profile.organization.delete()
+            profile.refresh_from_db()
+        elif "organization" in input:
+            organization_input = input.pop("organization")
+            if getattr(profile, "organization", None):
+                update_object(profile.organization, organization_input)
+            else:
+                Organization.objects.create(customer=profile, **organization_input)
```
igordavydsson, line 268: how about:
```
Organization.objects.update_or_create(customer=profile, defaults=organization_input)
```

customers/schema/mutations.py
a-rmz: Yup yup, makes sense. 👍

PR #246, opened 2020-07-24
Hotfix | Fix winter storage product season and price calculation

a-rmz:
## Description :sparkles:
There was a miscalculation of the winter season by 1 day. With this new calculation, it's also not necessary to check if the order lasts for the full season (`order.is_full_season()`), since it will always yield the correct result, even when "partially" calculating

### Related :handshake:
**[VEN-759](https://helsinkisolutionoffice.atlassian.net/browse/VEN-759):** add season logic to winter storage leases

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_models.py::test_order_winter_storage_lease_right_price_for_full_season
```

### DIFF
--- a/leases/utils.py
+++ b/leases/utils.py
@@ -1,5 +1,7 @@
 from datetime import date
 
+from dateutil.relativedelta import relativedelta
+
 
 def calculate_season_start_date(lease_start: date = None) -> date:
     """Return the date when the summer season starts
@@ -43,6 +45,58 @@ def calculate_season_end_date(lease_end: date = None) -> date:
     return default
 
 
+def calculate_winter_season_start_date(lease_start: date = None) -> date:
+    """Return the date when the winter season starts
+
+    If the current date is after the season end date,
+    it returns next year's default start date.
+
+    Winter Storage Leases always start on 15.9 the earliest
+    """
+    today = date.today()
+    default = date(day=15, month=9, year=today.year)
+
+    if lease_start:
+        lease_date = default.replace(year=lease_start.year)
+        # If the lease started between 1.1 and 10.6, that means
+        # the season began the previous year
+        if lease_start <= date(day=10, month=6, year=lease_start.year):
+            lease_date -= relativedelta(years=1)
+        return lease_date
+
+    # If the current day is before the winter season ends,
+    # the start day will be on the previous year
+    if today <= date(day=10, month=6, year=today.year):
+        default -= relativedelta(years=1)
+
+    return default
+
+
+def calculate_winter_season_end_date(lease_end: date = None) -> date:
+    """Return the date when the winter season ends
+
+    Winter Storage Leases always end on 10.6, the year depends on the current year
+    """
+    today = date.today()
+    default = date(day=10, month=6, year=today.year)
+
+    if lease_end:
+        lease_date = default.replace(year=lease_end.year)
+        # If the lease ended between 15.9 and 31.12, that means
+        # the season ends the following year.
+        if lease_end >= date(day=15, month=9, year=lease_end.year):
+            lease_date = default.replace(year=lease_end.year + 1)
+        return lease_date
+
+    # If today is gte than the day when all leases end,
+    # return the default end date for the next year
+    if today > default:
+        return default.replace(year=today.year + 1)
+
+    # Otherwise, return the default end date for the current year
+    return default
+
+
 def calculate_berth_lease_start_date() -> date:
     """
     Return the date when the lease season should start
@@ -73,13 +127,7 @@ def calculate_winter_storage_lease_start_date() -> date:
     """
 
     # Return the earliest date between the default start date or today
-    today = date.today()
-    default = max(date(day=15, month=9, year=today.year), today)
-
-    if today < date(day=10, month=6, year=today.year):
-        return today
-
-    return default
+    return max(calculate_winter_season_start_date(), date.today())
 
 
 def calculate_winter_storage_lease_end_date() -> date:
@@ -87,10 +135,4 @@ def calculate_winter_storage_lease_end_date() -> date:
 
     The season should end by default on 10.6 on the following year
     """
-    today = date.today()
-    default = date(day=10, month=6, year=today.year + 1)
-
-    if today.month <= default.month and today.day < default.day:
-        default = default.replace(year=today.year)
-
-    return default
+    return calculate_winter_season_end_date()
--- a/payments/models.py
+++ b/payments/models.py
@@ -10,7 +10,6 @@ from django.utils.translation import ugettext_lazy as _
 from enumfields import EnumField
 
 from leases.models import BerthLease, WinterStorageLease
-from leases.utils import calculate_season_end_date, calculate_season_start_date
 from payments.enums import (
     AdditionalProductType,
     OrderStatus,
@@ -344,13 +343,6 @@ class Order(UUIDModel, TimeStampedModel):
             # If the lease is being changed
             self._check_same_lease(old_instance)
 
-    def is_full_season(self):
-        return self.lease.start_date == calculate_season_start_date(
-            lease_start=self.lease.start_date
-        ) and self.lease.end_date == calculate_season_end_date(
-            lease_end=self.lease.end_date
-        )
-
     def _create_order_lines(self, section):
         for service in ProductServiceType.FIXED_SERVICES():
             # If the section has the service (prop) and it's valid (True)
@@ -416,14 +408,12 @@ class Order(UUIDModel, TimeStampedModel):
             tax_percentage = self.product.tax_percentage
 
             if self.lease:
-                # If the lease doesn't last for the whole season
-                if not self.is_full_season():
-                    price = calculate_product_partial_season_price(
-                        price,
-                        self.lease.start_date,
-                        self.lease.end_date,
-                        summer_season=isinstance(self.lease, BerthLease),
-                    )
+                price = calculate_product_partial_season_price(
+                    price,
+                    self.lease.start_date,
+                    self.lease.end_date,
+                    summer_season=isinstance(self.lease, BerthLease),
+                )
 
                 # If the order is for a winter product with a lease, the price
                 # has to be calculated based on the dimensions of the place associated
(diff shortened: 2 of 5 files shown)


payments/models.py
```diff
@@ -416,14 +408,12 @@ def save(self, *args, **kwargs):
...
             if self.lease:
-                # If the lease doesn't last for the whole season
-                if not self.is_full_season():
-                    price = calculate_product_partial_season_price(
-                        price,
-                        self.lease.start_date,
-                        self.lease.end_date,
-                        summer_season=isinstance(self.lease, BerthLease),
-                    )
+                price = calculate_product_partial_season_price(
```
igordavydsson, line 411: what if we rename this function to `calculate_product_season_price` since it is not necessarily `partial` anymore?

payments/tests/test_payments_models.py
```diff
@@ -387,6 +389,62 @@ def test_order_berth_lease_right_price_for_full_season(harbor):
...
+        if create:
+            additional_product = AdditionalProduct.objects.filter(
+                service=ProductServiceType(service),
+                price_unit=PriceUnits.AMOUNT,
+                period=PeriodType.SEASON,
+            ).first()
+            OrderLine.objects.create(order=order, product=additional_product)
+
+    assert order.lease.start_date == calculate_winter_storage_lease_start_date()
+    assert order.lease.end_date == calculate_winter_storage_lease_end_date()
```
igordavydsson, line 421: if you are not freezing time, you might get partial season leases, if you run the test between 15.9 and 9.6.

payments/tests/test_payments_models.py
```diff
@@ -387,6 +389,62 @@ def test_order_berth_lease_right_price_for_full_season(harbor):
...
+    sqm = order.lease.place.place_type.width * order.lease.place.place_type.length
+
+    expected_price = calculate_product_partial_season_price(
+        product.price_value,
+        calculate_winter_storage_lease_start_date(),
+        calculate_winter_storage_lease_end_date(),
+        summer_season=False,
+    )
+    expected_price = rounded(expected_price * sqm, decimals=2)
+    assert order.price == expected_price
```
igordavydsson, line 431: and doing this you just test that `calculate_product_partial_season_price` is used to calculate the price. a slightly more explicit approach to testing, I think, would be to: (1) freeze time and (2) compare `order.price` against some hardcoded `Decimal` value

payments/tests/test_payments_models.py
igordavydsson: same problem would happen not just in tests :derp: we should cover those cases as well in the `calculate_product_partial_season_price`

payments/models.py
a-rmz: 🤔 I have my doubts because it would still calculate the "partial" price as well as the "full" season price, and also to keep consistency with the other helper functions (partial month, and partial year)

payments/tests/test_payments_models.py
a-rmz: > same problem would happen not just in tests :derp: we should cover those cases as well in the calculate_product_partial_season_price

What do you mean? 🤨

igordavydsson: if you run the `calculate_product_partial_season_price` on 8.6, you would get only 2 days in the season

payments/models.py
igordavydsson: fair enough, consistency argument makes sense 👍

payments/tests/test_payments_models.py
a-rmz: But wouldn't that be the case? I mean, if a lease were to be generated on 8.6, then the season end would be 10.6, unless a different date is specified..? Still a bit confused 😅

igordavydsson: yes, lease should be 2 days indeed in that case. but season should always start on 15.9 for the proper price calculation. with this approach any leas (even if it is a 2 day lease) will be charged full season price. in other words: we need something similar to [this](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/leases/utils.py#L57)

a-rmz: But wouldn't [this](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/leases/utils.py#L77-L80) be enough then?
```python
def calculate_winter_storage_lease_start_date() -> date:
    """
    Return the date when the "winter" lease season should start
    The season should start by default on 15.9 (one day after the "summer" season ends)
    """

    # Return the earliest date between the default start date or today
    today = date.today()
    default = max(date(day=15, month=9, year=today.year), today)

    if today < date(day=10, month=6, year=today.year):
        return today

    return default
```

a-rmz: Scratch that, I see the problem now. I'll do some local testing and work on the fix 🦖

igordavydsson: a good unit test for this would be
```
create a berth product with period type SEASON and cost of e.g. 1 eur per day (268 eur per winter season on non-leap years)
create a lease for 15 days
assert order.price == Decimal("15.00")
```

leases/utils.py
```diff
@@ -43,6 +45,52 @@ def calculate_season_end_date(lease_end: date = None) -> date:
...
+    If the current date is after the season end date,
+    it returns next year's default start date.
+
+    Winter Storage Leases always start on 15.9 the earliest
+    """
+    today = date.today()
+    default = date(day=15, month=9, year=today.year)
+
+    if lease_start:
+        return default.replace(year=lease_start.year)
```
igordavydsson, line 60: this is a bit tricky, since WS lease can start between 1.1 and 10.6, in which case we should do `default.replace(year=lease_start.year - 1)`

leases/utils.py
```diff
@@ -43,6 +45,52 @@ def calculate_season_end_date(lease_end: date = None) -> date:
...
+
+    # If the current day is before the winter season ends,
+    # the start day will be on the previous year
+    if today < date(day=10, month=6, year=today.year):
```
igordavydsson, line 64: I think this should be `<=` 🤔

leases/utils.py
```diff
@@ -43,6 +45,52 @@ def calculate_season_end_date(lease_end: date = None) -> date:
...
+        default -= relativedelta(years=1)
+    # Otherwise, if the current day is after the season has started,
+    # the default start date will be on the following year
+    elif today >= date(day=15, month=9, year=today.year):
+        default += relativedelta(years=1)
```
igordavydsson, line 69: I don't thinkwe need this at all. 
- this function returns the start date for the current season, if season is going on right now
- it returns next season's startdate, if the season is already over

for this we need to:
- from 1.1 till 10.6 return 15.9 of the previous year
- from 10.6 return 15.9 of the current year

🤔

igordavydsson: so, if you just skip these lines, it should be ok, _I think_

leases/utils.py
```diff
@@ -43,6 +45,52 @@ def calculate_season_end_date(lease_end: date = None) -> date:
...
+def calculate_winter_season_end_date(lease_end: date = None) -> date:
+    """Return the date when the winter season ends
+
+    Winter Storage Leases always end on 10.6, the year depends on the current year
+    """
+    today = date.today()
+    default = date(day=10, month=6, year=today.year)
+
+    if lease_end:
+        return default.replace(year=lease_end.year)
```
igordavydsson, line 83: and here if the winter storage ends from 15.9 till 31.12, we should return 10.6 of the next year

payments/tests/test_payments_models.py
```diff
@@ -828,3 +884,34 @@ def test_order_tax_percentage():
...
+
+def test_berth_season_price(berth):
+    start_date = today()
+    end_date = start_date + relativedelta(days=15)
+
+    days_in_season = (calculate_season_end_date() - calculate_season_start_date()).days
+    product = BerthProductFactory(price_value=Decimal(days_in_season))
+    lease = BerthLeaseFactory(berth=berth, start_date=start_date, end_date=end_date)
+    order = Order.objects.create(product=product, lease=lease, customer=lease.customer)
+    assert order.price == Decimal("15.00")
```
igordavydsson, line 917: these are. amazing 😻 :pepelove:

payments/tests/test_payments_models.py
```diff
@@ -387,6 +395,54 @@ def test_order_berth_lease_right_price_for_full_season(harbor):
             assert product_price == order_price
 
 
+@freeze_time("2020-06-11T08:00:00Z")
+def test_order_winter_storage_lease_right_price_for_full_season(winter_storage_area):
```
igordavydsson, line 399: very nice job with this test as well 🎖️

leases/utils.py
a-rmz: :hm: right, it's not as straightforward as the summer season one.

leases/utils.py
a-rmz: So on 10.6.2020 the start date will be 10.6.2021?

igordavydsson: on 10.6.2020 the start will be 15.9.2019

leases/utils.py
a-rmz: So on `10.10.2020` the start date would be `15.9.2020`, right? Then we can get rid of those lines.

igordavydsson: >So on 10.10.2020 the start date would be 15.9.2020, right? 

yes, I think that is the logic we need

igordavydsson: this will be consistent with the logic for summer seasons, where we also have these rules:
> - this function returns the start date for the current season, if season is going on right now
> - it returns next season's start date, if the season is already over

leases/utils.py
a-rmz: Ohhh yeah, yeah, okay.

leases/utils.py
a-rmz: Right, makes sense returning the beginning of the ongoing season. :thu

leases/utils.py
```diff
@@ -43,6 +45,58 @@ def calculate_season_end_date(lease_end: date = None) -> date:
...
+    """
+    today = date.today()
+    default = date(day=15, month=9, year=today.year)
+
+    if lease_start:
+        lease_date = default.replace(year=lease_start.year)
+        # If the lease started between 1.1 and 10.6, that means
+        # the season began the previous year
+        if lease_start <= date(day=10, month=6, year=lease_start.year):
+            lease_date = default.replace(year=lease_start.year - 1)
```
igordavydsson, line 64: total nitpick, but would `lease_date = -= relativedelta(years=1)` work here and make `calculate_winter_season_start_date` more consistent? 😅

PR #248, opened 2020-07-28
VEN-695 | Generate order with lease

a-rmz:
## Description :sparkles:
Generate an `Order` whenever a `winter storage/berth lease` is generated _only_ through the `createXXXLease` mutations.

It will automatically look for a `Berth/WinterStorageProduct` that can be associated to the lease based on the `Harbor/Area` to which the lease place belongs, and a `BerthPriceGroup` for `BerthLease`s.

For `BerthLease`s, it will `get_or_create` the `BerthPriceGroup` every time. In case the price group wouldn't exist, it will be generated, but the `order/lease` creation will fail because of the lack of `BerthProduct`, since there will be no products related to the new price group.

### Closes :no_good_woman:
**[VEN-695](https://helsinkisolutionoffice.atlassian.net/browse/VEN-695):** generate an order (a.k.a "invoice") when generating Lease

### Automated tests :gear:️
```shell
pytest -k "test_create_berth_lease_with_order or test_create_winter_storage_lease_with_order"
```

Before executing the mutation, make sure you have an existing `BerthProduct` for the `berth.pier.harbor` you're gonna use.

```graphql
mutation CREATE_LEASE($input: CreateBerthLeaseMutationInput!) {
  createBerthLease(input:$input){
    berthLease {
      id
      startDate
      endDate
      customer {
        id
      }
      boat {
        id
      }
      status
      comment
      berth {
        id
      }
      application {
        id
      }
    }
  }
}
```

**input:**
```json
{
  "input": {
    "applicationId": "",
    "berthId": ""
  }
}
```
## Screenshots :camera_flash:
**Create Berth lease:**
image

**Create Winter Storage lease:**
image

### DIFF
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -8,6 +8,7 @@ from applications.models import BerthApplication, WinterStorageApplication
 from applications.new_schema import BerthApplicationNode, WinterStorageApplicationNode
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.models import CustomerProfile
+from payments.models import BerthPriceGroup, BerthProduct, Order, WinterStorageProduct
 from resources.schema import BerthNode, WinterStoragePlaceNode
 from users.decorators import (
     add_permission_required,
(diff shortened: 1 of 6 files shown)


leases/schema/mutations.py
```diff
@@ -78,6 +79,17 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+            price_group = BerthPriceGroup.objects.get_or_create_for_width(
+                berth.berth_type.width
+            )
+            product = BerthProduct.objects.get(
+                harbor=berth.pier.harbor, price_group=price_group
+            )
+            Order.objects.create(
+                customer=input["customer"], lease=lease, product=product
+            )
+        except (BerthPriceGroup.DoesNotExist, BerthProduct.DoesNotExist) as e:
```
igordavydsson, line 91: hm, since you do `BerthPriceGroup.objects.get_or_create_for_width` you will not encounter `BerthPriceGroup.DoesNotExist` 🤔  so, either switch to `BerthPriceGroup.objects.get` or skip this exception

a-rmz: Right!! I forgot to remove it from my testings. 😅

PR #253, opened 2020-08-03
[VEN-712] Upgrade dependencies

a-rmz:
## Description :sparkles:
* Upgrade Django to v3.x
  * Remove `django-enumfields` and change to [Django TextChoices](https://docs.djangoproject.com/en/3.0/ref/models/fields/#enumeration-types)
* Upgrade [`django-ilmoitin`](https://github.com/City-of-Helsinki/django-ilmoitin/releases/tag/v0.6.0)
* Upgrade [`django-munigeo`](https://github.com/a-rmz/django-munigeo/releases/tag/0.4.0-beta) (WIP)

### Closes :no_good_woman:
**[VEN-712](https://helsinkisolutionoffice.atlassian.net/browse/VEN-712):** upgrade dependencies

### Automated tests :gear:
```shell
$ pytest
```

We are using a [forked version](https://github.com/a-rmz/django-munigeo) of `django-munigeo` with the changes we require since we cannot merge it right away. Once the changes are updated to the [upstream repo](https://github.com/City-of-Helsinki/django-munigeo), we'll replace the requirement.

We are also bumping the minor version of python from `3.6 -> 3.8`, so we need to coordinate with Kanslia people to make sure it runs smoothly.


harbors/schema.py
```diff
@@ -89,35 +86,28 @@ class Meta:
...
-            "summer_storage_for_trailers",
-            "summer_storage_for_boats",
-            "availability_level",
-            "number_of_marked_places",
-            "max_width",
-            "max_length",
-            "number_of_section_spaces",
-            "max_length_of_section_spaces",
-            "number_of_unmarked_spaces",
-            "location",
```
igordavydsson, line 114: looked at the linked release notes and PRs and cannot see why this change is needed 🤔  these fields are model fields, graphene should be able to convert them automatically. linked PRs were about fields, that are listed in graphene class' `Meta` but are not model fields 🤔

applications/utils.py
```diff
@@ -155,8 +156,12 @@ def export_winter_storage_applications_as_xlsx(applications):
                     value = getattr(application, attr_name)
                     if isinstance(value, bool):
                         value = "Yes" if value else ""
-                    elif isinstance(value, Enum):
-                        value = str(value)
+                    elif value in ApplicationStatus:
+                        status = ApplicationStatus(value)
+                        value = str(getattr(status, "label", str(status)))
```
igordavydsson, line 161: application status field is not used when generating excels, you can skip these lines, only `storage_method` change was needed

payments/migrations/0001_initial.py
```diff
@@ -53,8 +52,8 @@ class Migration(migrations.Migration):
                 ),
                 (
                     "price_unit",
-                    enumfields.fields.EnumField(
-                        default="amount", enum=payments.enums.PriceUnits, max_length=10
+                    models.CharField(
+                        choices=[("amount", "Amount")], default="amount", max_length=10,
```
igordavydsson, line 56: probably `AdditionalProduct` can be amount or percentage

payments/migrations/0001_initial.py
```diff
@@ -154,8 +175,10 @@ class Migration(migrations.Migration):
                 ),
                 (
                     "price_unit",
-                    enumfields.fields.EnumField(
-                        default="amount", enum=payments.enums.PriceUnits, max_length=10
+                    models.CharField(
+                        choices=[("amount", "Amount"), ("percentage", "Percentage")],
```
igordavydsson, line 179: but `WinterStorageProduct` only `amount`

harbors/schema.py
a-rmz: Yup, the reason why I went on with this is because, otherwise, it raises the following error:

```
UserWarning: Field name "number_of_places" matches an attribute on Django model "harbors.Harbor" but it's not a model field so Graphene cannot determine what type it should be. Either define the type of the field on DjangoObjectType "HarborType" or remove it from the "fields" list.
```
with all the fields inherited from `AbstractArea`. The curious thing is that this only happens with resources on the `harbors` app, so I guess it could have something to do on how the schema is built? Since the "old" schema is built explicitly by `graphene.Schema`, but the "new" one is built through Federation... 🤔 

Without the explicit declaration, _it works_, but it throws a ton of warnings.

image

igordavydsson: ah, interesting, they are all passed as attributes from the abstract base model. well, then the change is ok 🙆‍♂️

setup.cfg
```diff
@@ -41,6 +30,7 @@ include_trailing_comma = True
 force_grid_wrap = 0
 use_parentheses = True
 line_length = 88
+src_paths=.
```
igordavydsson, line 33: if it works without this setting, we can maybbe skip it?

payments/schema/types.py
```diff
@@ -5,7 +5,6 @@
 from leases.schema import BerthLeaseNode, WinterStorageLeaseNode
 from resources.schema import HarborNode, WinterStorageAreaNode
 from users.decorators import view_permission_required
-from utils.enum import graphene_enum
```
igordavydsson, line 8: do we still use this anywhere after all the changes? should we delete it?

setup.cfg
a-rmz: Skipped 👍

payments/schema/types.py
a-rmz: We're still using it in one place for:

[`resources/schema.py#L47`](https://github.com/City-of-Helsinki/berth-reservations/pull/253/files#diff-09f1721fc193a5a442fee1cf24f5bd49R47)
```python
BerthMooringTypeEnum = graphene_enum(BerthMooringType)
```

payments/models.py
```diff
@@ -127,14 +128,21 @@ class WinterStorageProduct(AbstractPlaceProduct):
         related_name="product",
         on_delete=models.CASCADE,
     )
+    price_unit = models.CharField(
+        choices=[(PriceUnits.AMOUNT, PriceUnits.AMOUNT.label)],
+        default=PriceUnits.AMOUNT,
+        max_length=10,
+    )
```
igordavydsson, line 135: why is this needed, if it's already defined on the `AbstractPlaceProduct`? 🤔

payments/models.py
```diff
@@ -145,9 +153,9 @@ class AdditionalProduct(AbstractBaseProduct):
 
     @property
     def product_type(self):
-        if self.service.is_fixed_service():
+        if self.service in ProductServiceType.FIXED_SERVICES():
             return AdditionalProductType.FIXED_SERVICE
-        elif self.service.is_optional_service():
+        elif self.service in ProductServiceType.OPTIONAL_SERVICES():
```
igordavydsson, line 158: I guess, these methods (`is_fixed_service` and `is_optional_service`) do not work anymore on an instance of `AdditionalProduct`? i.e. doing `self.service.is_fixed_service()` would return `None` or something? maybe we should get rid of those static methods then? they are used in some places that you did not update during this commit by the way ([example](https://github.com/City-of-Helsinki/berth-reservations/blob/c7d9fce97f38eba8181556834fccfe54e2ffd854/payments/schema/types.py#L151))

igordavydsson: ah, doing this `ProductServiceType(self.service).is_fixed_service()` works. I see. still, double check that we don't do `self.service.is_fixed_service()` anywhere

payments/models.py
a-rmz: Excellent question :derp: removed now.

payments/models.py
a-rmz: Good catch, I forgot to check all places, but now I got rid of it, we can get around perfectly without it 👍

PR #254, opened 2020-08-05
VEN-739 | Add area to winter storage leases

a-rmz:
## Description :sparkles:
* Add an `area` field to `WinterStorageLease` in case a `WinterStorageArea` does not have marked places, so it can be assigned to the whole area instead of only `place`s

### Closes :no_good_woman:
**[VEN-739](https://helsinkisolutionoffice.atlassian.net/browse/VEN-739):** link WS leases simply to WS area if it has no marked places

### Automated tests :gear:️
```shell
$ pytest leases/tests/test_lease_models.py::test_get_winter_storage_area_with_leases
$ pytest leases/tests/test_lease_models.py::test_get_winter_storage_area_with_leases_not_enough_permissions
$ pytest leases/tests/test_lease_queries.py::test_query_winter_storage_leases
```

```graphql
query WSL {
  winterStorageAreas {
    edges {
      node {
        properties {
          leases {
            edges {
              node {
                id
              }
            }
          }
        }
      }
    }
  }
}
```


leases/models.py
```diff
@@ -218,8 +228,12 @@ def clean(self):
             existing_leases = existing_leases.exclude(pk=self.pk)
         if existing_leases.exists():
             raise ValidationError(_("WinterStoragePlace already has a lease"))
-        if not self.place.is_active and creating:
+        if creating and self.place and not self.place.is_active:
             raise ValidationError(_("Selected place is not active"))
+
+        if self.place and self.area:
+            raise ValidationError(_("Lease cannot have both place and area assigned"))
+
```
santtul, line 236: Should we also check that either place or area is set?

leases/models.py
```diff
@@ -218,8 +228,12 @@ def clean(self):
             existing_leases = existing_leases.exclude(pk=self.pk)
         if existing_leases.exists():
```
santtul, line 229: With the current implementation there can be only one "winterstorage area type lease", if you try to add another one, you get "WinterStoragePlace already has a lease" because of this check.

leases/models.py
a-rmz: Yes! Good catch. 👍

leases/models.py
a-rmz: Ohhhhh sneaky one. 😅

santtul: There seems to be still problem with this :)
````
        elif self.area:
            existing_leases = existing_leases.exclude(area=self.area)
````
With the check above you get "WinterStoragePlace already has a lease" if some *other* WS Area has leases.

File: leases/models.py [hunk 3]
```diff
@@ -216,10 +230,21 @@ def clean(self):
...
             existing_leases = existing_leases.exclude(pk=self.pk)
+
+        # WSPlaces can only have one lease associated, but WSAreas can have many,
+        # so we don't check if there are existing leases associated to the area.
+        if self.place:
+            existing_leases = existing_leases.filter(place=self.place)
+        elif self.area:
+            existing_leases = existing_leases.exclude(area=self.area)
+
+        # print(existing_leases.all())
```
santtul, line 241: I think we can remove this debug line

a-rmz: :derp: completely missed it

leases/models.py
a-rmz: Right 🤔 then if the lease is associated with an area, we don't need to check fro `existing` leases since there can be many associated to the it.

PR #255, opened 2020-08-13
[VEN-327] | Bambora payments

a-rmz:
## Description :sparkles:
* Add Bambora payment provider and the required flow to send payments to Bambora
* Add extra fields to models to properly work with Bambora:
  * `Order.order_number`
  * `OrderStatus.CANCELLED`
* Change `OrderLogEntry` to support more detailed satus transitions
  * `OrderLogEntry.status` -> `OrderLogEntry.to_status`
  * `OrderLogEntry.from_status`

### Closes :no_good_woman:
**[VEN-327](https://helsinkisolutionoffice.atlassian.net/browse/VEN-327):** send order to Bambora

### Automated tests :gear:️
```shell
$ python manage.py payments/tests/test_bambora_payform
```

0. Set the required env variables
```
VENE_PAYMENTS_BAMBORA_API_KEY={API_KEY}
VENE_PAYMENTS_BAMBORA_API_SECRET={PRIVATE_KEY}
VENE_PAYMENTS_BAMBORA_PAYMENT_METHODS=creditcards,banks
VENE_UI_RETURN_URL=https://venepaikat.hel.fi/payment
```
1. Create a new `Order` with a `Lease` and `Application`
2. Go to the "old" api `/graphql`, and run the following mutation to get the URL:
```graphql
mutation Confirm {
  confirmPayment(input: {orderNumber: ""}) {
    url
  }
}
```
3. The mutation will return a Bambora PayForm URL, follow that URL and pay with the credit card
(if you want to try other payment methods or other cards, you can find more details on the [Bambora docs](https://payform.bambora.com/docs/web_payments/?page=testing#Test_card_numbers))
```
number = 4012888888881881
expiration = any date
CVV =  any number (3 digits)
```
4. The payment should be successful and redirect you to `https://venepaikat.hel.fi/payment?payment_status=success`

* A cron job (or something of that sort) needs to be set up to check for expired orders and change the status, running the `python manage.py expire_too_old_unpaid_orders`
* Documentation about the payment configuration and flow can be found on `docs/payments.md`
* The following environment variables have to be added:
```env
VENE_PAYMENTS_BAMBORA_API_KEY=VENE_PAYMENTS_BAMBORA_API_KEY
VENE_PAYMENTS_BAMBORA_API_SECRET=VENE_PAYMENTS_BAMBORA_API_SECRET
VENE_PAYMENTS_BAMBORA_PAYMENT_METHODS=creditcards,banks
VENE_UI_RETURN_URL=https://venepaikat.hel.fi/payments
```


payments/schema/mutations.py
```diff
@@ -379,6 +381,26 @@ def mutate_and_get_payload(cls, root, info, **input):
         return DeleteBerthProductMutation()
 
 
+class ConfirmPaymentMutation(graphene.ClientIDMutation):
```
a-rmz, line 384: Should we also send some kind of `one-time-token` to the UI and then expect it on the `confirm` mutation just to have some kind of check that the request is not just coming from a random client?

santtul: I think that the `order_number` already is our "one-time-token". If we are concerned that it's too easy to guess, then we should make `order_number` stronger instead. Are we concerned or has "respa" been concerned about it? :)

payments/models.py
```diff
@@ -450,6 +478,58 @@ def save(self, *args, **kwargs):
...
+            self.lease.status = LeaseStatus.PAID
+        elif new_status in (OrderStatus.REJECTED, OrderStatus.CANCELLED):
+            self.lease.status = LeaseStatus.REFUSED
+        elif new_status == OrderStatus.EXPIRED:
+            self.lease.status = LeaseStatus.EXPIRED
+        elif new_status == OrderStatus.WAITING:
+            self.lease = LeaseStatus.OFFERED
+
+        self.lease.save()
+        self.save()
```
igordavydsson, line 515: maybe add `update_fields=["status"]` for both of these? to be on the safe side

payments/models.py
```diff
@@ -558,8 +638,14 @@ class OrderLogEntry(UUIDModel, TimeStampedModel):
         related_name="log_entries",
         on_delete=models.CASCADE,
     )
-    status = EnumField(OrderStatus)
+    from_status = EnumField(OrderStatus, blank=True, null=True)
```
igordavydsson, line 641: not sure if it can be blank 🤔

berth_reservations/settings.py
```diff
@@ -47,6 +47,9 @@
     TOKEN_AUTH_AUTHSERVER_URL=(str, ""),
     TOKEN_AUTH_FIELD_FOR_CONSENTS=(str, "https://api.hel.fi/auth"),
     TOKEN_AUTH_REQUIRE_SCOPE_PREFIX=(bool, True),
+    VENE_PAYMENTS_ENABLED=(bool, True),
```
igordavydsson, line 50: respa had this setting because it is used by multiple cities, and some do not have payable resources. respa can be used without payments, but vene can't. I think we could skip this setting in our backend.

a-rmz: True! I thought I had removed it everywhere but I missed this occurrence. 👍

payments/models.py
a-rmz: I kept it like that in order to be able to migrate the existing `status -> to_status`, so if there's any existing orders, there wouldn't be any "from" status.

Another option would be to set `WAITING` as default `from_status` on the migration and make it non-nullable.

docs/payment_flow.txt
```diff
@@ -0,0 +1,30 @@
+# Generated with https://sequencediagram.org/
+
+title Venepaikka payment flow
+
+note over Admin: Confirm order
+Admin->Back-end: confirmOrderMuation
+Back-end->User: Send mail to Customer UI with order number\nhttps://venepaikat.hel.fi/{LANG}/payments?order_number={str}
```
santtul, line 7: path will be `payment`

docs/payment_flow.txt
```diff
@@ -0,0 +1,30 @@
...
+
+
```
santtul, line 9: I think we should add "Fetch order details" here

docs/payment_flow.txt
```diff
@@ -0,0 +1,30 @@
...
+Back-end-->>User: PayForm URL
+
+note over User: Customer UI is responsible for handling\nthe response url and redirecting the user
+User->Bambora PayForm: Redirect
+
+note over Bambora PayForm: Handle payment
+
+Bambora PayForm->Back-end: Redirect to success/failure
+note over Back-end: Update order details
+Back-end-->>User: Redirect to Customer UI success/failure page\nhttps://venepaikat.hel.fi/payments?payment_status={success|failure}&order_number={str}
```
santtul, line 28: Do we need `order_number` param here? Could the URL be:
`venepaikat.hel.fi/payment-result?status={success|failure}` ?

payments/schema/mutations.py
a-rmz: Should be safe enough, just thinking on how to make it a bit more secure, but the ID generation is not a concern.

docs/payment_flow.txt
a-rmz: 🤔 not unless we would fetch details from the order (which I guess we're not)

santtul: yeah, there are no plans to fetch them now.

payments/providers/base.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, **kwargs):
+        if "config" in kwargs:
+            self.config = kwargs.get("config")
+        self.request = kwargs.get("request")
+        self.ui_return_url = kwargs.get("ui_return_url")
+
+    def initiate_payment(self, order: Order) -> str:
+        """Create a payment to the provider.
+        Implement this in your subclass. Should return a URL to which the user
+        is redirected to actually pay the order."""
```
igordavydsson, line 24: should this `raise NotImplementedError`?

igordavydsson: well, Tuomas left it out, maybe there was a reason 😅 🤷‍♂️ let's follow respa

a-rmz: I had the exact same train of thought. But I guess we can just add it as a failsafe?

igordavydsson: let's add it, I think it should be ok

payments/providers/bambora_payform.py
```diff
@@ -0,0 +1,313 @@
...
+            return self.ui_redirect_failure(order)
+
+    def handle_notify_request(self):
+        """Handle the asynchronous part of payform response
+
+        Arrives some time after user has completed the payment flow or stopped it abruptly.
+        Skips changing order status if it has been previously set. Although, according to
+        Bambora's documentation, there are some cases where payment status might change
+        from failed to successful, the reservation has probably been soft-cleaned up by then.
+        TODO Maybe try recreating the reservation if the time slot is still available
```
igordavydsson, line 275: I think we don't soft-clean our orders. I guess, if first API call to Bambora said that transaction did not go through, but this notify call says so, we can just mark the order as paid. 🤔

igordavydsson: ... which is what this function does. so, just delete these couple lines from this docstring?

docs/payment_flow.txt
igordavydsson: for the sake of clarity, I would maybe say here `Send email with the following link to Customer UI` 🤔

payments/providers/bambora_payform.py
a-rmz: Alright 🤷

a-rmz: So, should we allow transition from cancelled to paid? In case the notify would mark that transition.

igordavydsson: I was thinking about that, but did not fully get the flow. Can it happen so, that first Bambora says "Failed" and we mark it as "Rejected" or "Canceled" but then a couple seconds later it sends us a notification that "Hey, actually it went ok"?
If that is the case, we should probably allow going from "Canceled" to "Paid"

payments/schema/mutations.py
```diff
@@ -551,3 +573,5 @@ class Mutation:
         "\n\nErrors:"
         "\n* The passed `OrderLine` does not exist"
     )
+
+    confirm_payment = ConfirmPaymentMutation.Field()
```
santtul, line 577: Let's move this under `/graphql` as discussed.

payments/providers/bambora_payform.py
a-rmz: According to Bambora docs, it would be rare but possible:

> _In some rare cases it's possible that the status of the transaction changes from failed to successful later. In such cases the notify request is done again._

payments/models.py
a-rmz: Ping @igordavydsson

payments/schema/mutations.py
igordavydsson: good thing, that you thought about it 👍  I think `order_number` is strong enough for now and in the future (fopefully 😅 ), the customer will be paying these after logging into her profile, so there will be an auth layer

payments/models.py
igordavydsson: we don't have any orders in prod, so we can maybe set a default on the migration, that would be ok for our test instance 👍

payments/schema/mutations.py
santtul: good discussion. one clarification: there will be no customer login in the first version we are about to release.

payments/providers/bambora_payform.py
igordavydsson: ok, then we should allow going from `REJECTED` to `PAID` for the sake of this notify call 👍

payments/schema/mutations.py
igordavydsson: yes, but that is a feature mentioned on the roadmap of the project, so maybe hopefully potentially one day 🙂

payments/providers/bambora_payform.py
a-rmz: There would also be the case when a payment was rejected because something went wrong with the payment itself. For that I think we should allow moving from `REJECTED -> WAITING` so the user can try and pay again, just not sure how we could "enable" this somehow for the user.

igordavydsson: hmm, yeah, if that can happen, going `REJECTED -> WAITING` should be possible

payments/models.py
igordavydsson: 👀   🤔

igordavydsson: ah, it's in the next commit, ok

payments/providers/bambora_payform.py
```diff
@@ -0,0 +1,317 @@
...
+        order_lines: [OrderLine] = OrderLine.objects.filter(order=order.id)
+        items: [dict] = []
+
+        product = order.product
+        int_tax = int(order.tax_percentage)
+        assert int_tax == product.tax_percentage  # make sure the tax is a whole number
+        items.append(
+            {
+                "id": str(product.id),
+                "title": product.name,
```
igordavydsson, line 121: do we need to activate some language for this?

a-rmz: Somehow yes, but `product.name` calls a getter that does this, so I'm not sure how should we handle it.
```python
@property
def name(self):
    return f"{_('Berth product')}: {self}"

@property
def name(self):
    return f"{self.service} - {self.period}"
```

igordavydsson: maybe we should do something like:
```python
with activate(order.application.language):
    product_name = product.name
```
🤔

payments/providers/base.py
```diff
@@ -0,0 +1,118 @@
...
+    def _get_final_return_url(self, vene_return_url):
+        query_params = urlencode({self.ui_return_url_param_name: self.ui_return_url})
+        return "{}?{}".format(vene_return_url, query_params)
+
+    @classmethod
+    def _redirect_to_ui(cls, return_url: str, status: str, order: Order = None):
+        params = {"payment_status": status}
+        if order:
+            params["order_number"] = order.order_number
+        return redirect("{}?{}".format(return_url, urlencode(params)))
```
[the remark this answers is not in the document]
santtul: Could the URL be:
`https://venepaikat.hel.fi/{LANG_CODE}/payment-result?payment_status={success|failure}` ?

santtul (changes requested):
Tested bambora payment locally and it works 🥳  Some comments there.

docs/payments.md
```diff
@@ -0,0 +1,25 @@
...
+
+The Venepaikka payments rely mostly on the [Respa](https://github.com/City-of-Helsinki/respa/blob/develop/docs/payments.md) implementation, however, changes were made to adapt it to the requirements of the project.
+
+## Configuring Payments
+
+There are a couple of required configuration keys that need to be set in order to enable the payments and the third party provider currently in use:
+
+- `VENE_PAYMENTS_PROVIDER_CLASS`: Dotted path to the active provider class as a string. The default value is `payments.providers.BamboraPayformProvider`.
+
+`./manage.py expire_too_old_unpaid_orders` runs the order/reservation cleanup for current orders. You'll probably want to run it periodically at least in production. [Cron](https://en.wikipedia.org/wiki/Cron) is one candidate for doing that.
```
santtul, line 11: One thing to consider: Cron is not ideal solution with Docker. It would be great if we could aim for solution that would play well with KuVa infra (Docker).

a-rmz: We can set a [Kubernetes CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/) that runs the command.

payments/providers/bambora_payform.py
a-rmz: Done :D

locale/en/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,1311 @@
+# Berth Reservations API
+# Copyright (C) 2020 City of Helsinki
+# This file is distributed under the same license as the berth-reservations package.
```
igordavydsson, line 3: why do we need a file for `en`? 🤔

a-rmz: My intuition tells me because `fi` is the default language 🤔  but not acutally sure, I just created the file for the three languages we support.

igordavydsson: When you do `_("My string")` you tell Django that this is a translatable string with id `My string`. Then Django checks for active language and checks its .mo/.po files. If no file is given it will use this id as string itself. Since we agreed beforehand that all our ids are in english, we can basically skip locale files for `en`.

a-rmz: Ahaa! Okay, okay, makes sense then 👍

santtul (approved):
Great work! 👍

igordavydsson (approved):
splendid

PR #256, opened 2020-08-19
[VEN-771] | Add is_active field to Winter Storage leases

a-rmz:
## Description :sparkles:
* Add a `is_active` field to WSLeases

### Closes :no_good_woman:
**[VEN-771](https://helsinkisolutionoffice.atlassian.net/browse/VEN-771):**  add "isActive" field to WinterStorageLeaseNode

### Automated tests :gear:️
```shell
$ pytest leases/tests/test_lease_models.py
```

## Screenshots :camera_flash:
image

### DIFF
--- a/leases/models.py
+++ b/leases/models.py
@@ -169,6 +169,29 @@ class BerthLease(AbstractLease):
         )
 
 
+class WinterStorageLeaseManager(models.Manager):
+    def get_queryset(self):
+        current_season_start = calculate_winter_storage_lease_start_date()
+        today = date.today()
+
+        is_paid = Q(status=LeaseStatus.PAID)
+
+        if today < current_season_start:
+            in_current_season = Q(start_date=current_season_start)
+        else:
+            in_current_season = Q(start_date__lte=today) & Q(end_date__gte=today)
+
+        return (
+            super()
+            .get_queryset()
+            .annotate(
+                is_active=ExpressionWrapper(
+                    in_current_season & is_paid, output_field=models.BooleanField(),
+                )
+            )
+        )
+
+
 class WinterStorageLease(AbstractLease):
     place = models.ForeignKey(
         WinterStoragePlace,
@@ -200,6 +223,7 @@ class WinterStorageLease(AbstractLease):
     end_date = models.DateField(
         verbose_name=_("end date"), default=calculate_winter_storage_lease_end_date
     )
+    objects = WinterStorageLeaseManager()
 
     class Meta:
         verbose_name = _("winter storage lease")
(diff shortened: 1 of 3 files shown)


leases/tests/test_lease_models.py
```diff
@@ -349,6 +350,46 @@ def test_berth_lease_is_not_active_after_today():
...
+    )
+
+    assert not WinterStorageLease.objects.get(id=lease.id).is_active
+
+
+@freeze_time("2020-07-10T08:00:00Z")
+def test_winter_storage_lease_is_not_active_after_today():
+    start_date = date.today()
+    lease = WinterStorageLeaseFactory(status=LeaseStatus.PAID, start_date=start_date)
+    assert not WinterStorageLease.objects.get(id=lease.id).is_active
```
igordavydsson, line 390: hm, interesting catch, I think this case should be `true` 😅 since if the lease starts today, then yeah, it is active today and untill the day it ends.

leases/models.py
```diff
@@ -169,6 +169,30 @@ def __str__(self):
         )
 
 
+class WinterStorageLeaseManager(models.Manager):
+    def get_queryset(self):
+        current_season_start = calculate_winter_storage_lease_start_date()
+        today = date.today()
+
+        active_current_status = Q(status=LeaseStatus.PAID)
```
santtul, line 177: Could this var be `is_paid` ?

santtul (changes requested):
Small comment left. Nice tests btw 👍

leases/tests/test_lease_models.py
a-rmz: 🤔 fair point

leases/models.py
```diff
@@ -169,6 +169,29 @@ def __str__(self):
         )
 
 
+class WinterStorageLeaseManager(models.Manager):
+    def get_queryset(self):
+        current_season_start = calculate_winter_storage_lease_start_date()
```
igordavydsson, line 174: should this use `calculate_winter_season_start_date`?

igordavydsson: hm, seems like it does not matter for the logic we have 🤔

igordavydsson (approved):
Had to think about the point I commented on, but it seems like the solution is good. 👍  Maybe renaming `current_season_start` var would help with the clarity, but it is ok like it is as well. 🙂 

make it happen

leases/models.py
a-rmz: Yeah, I think for this case it would be the same.

santtul (approved):
LGTM!

PR #259, opened 2020-08-24
Delete obsolete "data" field from application models

igordavydsson:
## Description :sparkles:

The field was used  during the first stages of the project and
has since become obsolete.


a-rmz (approved):
🤩

PR #267, opened 2020-08-28
VEN-799 | Customer UI schema updates

santtul:
## Description :sparkles:

- [x] Rename orderStatus query to orderDetails
- [x] Remove unnecessary optionals from customer ui schema

### Related :handshake:
[VEN-799](https://helsinkisolutionoffice.atlassian.net/browse/VEN-799)


harbors/schema.py
```diff
@@ -96,18 +96,18 @@ class Meta:
     municipality = graphene.String()
     image_file = graphene.String()
     servicemap_id = graphene.String()
-    zip_code = graphene.String()
+    zip_code = graphene.String(required=True)
```
igordavydsson, line 99: also `phone` and `email`, they are non-null too

igordavydsson: and `image_link`

harbors/schema.py
```diff
@@ -45,17 +45,17 @@ class Meta:
...
     image_file = graphene.String()
     servicemap_id = graphene.String()
-    zip_code = graphene.String()
-    phone = graphene.String()
-    email = graphene.String()
-    www_url = graphene.String()
+    zip_code = graphene.String(required=True)
+    phone = graphene.String(required=True)
+    email = graphene.String(required=True)
+    www_url = graphene.String(required=True)
```
igordavydsson, line 51: and `image_link` for `HarborType` should be `required=True`

harbors/schema.py
a-rmz: How bout `suitable_boat_types`?

igordavydsson: yeah, `suitable_boat_types` should be also required, as it can be empty list, but not `null`

harbors/schema.py
```diff
@@ -45,17 +45,17 @@ class Meta:
...
-    waste_collection = graphene.Boolean()
-    gate = graphene.Boolean()
-    lighting = graphene.Boolean()
+    mooring = graphene.Boolean(required=True)
+    electricity = graphene.Boolean(required=True)
+    water = graphene.Boolean(required=True)
+    waste_collection = graphene.Boolean(required=True)
+    gate = graphene.Boolean(required=True)
+    lighting = graphene.Boolean(required=True)
     suitable_boat_types = graphene.List(BoatTypeType)
```
a-rmz, line 59: Same for `suitable_boat_types` here

harbors/schema.py
santtul: Good point gentlemen! Will fix these.

harbors/schema.py
```diff
@@ -45,18 +45,20 @@ class Meta:
...
+    image_link = graphene.String(required=True)
+    mooring = graphene.Boolean(required=True)
+    electricity = graphene.Boolean(required=True)
+    water = graphene.Boolean(required=True)
+    waste_collection = graphene.Boolean(required=True)
+    gate = graphene.Boolean(required=True)
+    lighting = graphene.Boolean(required=True)
+    suitable_boat_types = graphene.NonNull(
+        graphene.List(graphene.NonNull(BoatTypeType))
+    )
```
igordavydsson, line 61: this says that there is guaranteed to be a suitable boat type, which is not really technically true, although _usually_ true in terms of business logic. just `suitable_boat_types = graphene.List(BoatTypeType, required=True)` should be enough

santtul: It depends how we represent the "no suitable boat types", if that is represented as empty array then I think it's correct because it generates the following contract:
`suitableBoatTypes: BoatTypesBerthsQuery_harbors_edges_node_properties_suitableBoatTypes[];`

if I apply `suitable_boat_types = graphene.List(BoatTypeType, required=True)` we get contract:
`suitableBoatTypes: (BoatTypesBerthsQuery_harbors_edges_node_properties_suitableBoatTypes | null)[];`
So it means that the array it self can contain null values, which is not nice I think.

What do you think?

santtul: We btw get the same result with:
`suitable_boat_types = graphene.List(graphene.NonNull(BoatTypeType), required=True)`
which actually looks better.

igordavydsson: looking at it from the generated GQL schema:
- your current code generates `suitableBoatTypes: [BoatTypeType!]!` which is just not true, as we can't guarantee that there will always be items in the list
- my suggestion generates: `suitableBoatTypes: [BoatTypeType]!` which means it could be empty list or list with null values, as GQL does not differentiate unfortunately

igordavydsson: yeah, no-need to wrap `graphene.List` in `NonNull`, since you can pass `required=True`

santtul: ok, fair enough! 👍  will fix it.

a-rmz: As per the [Graphene docs](https://docs.graphene-python.org/en/latest/types/list-and-nonnull/#nonnull)

```python
import graphene

class Character(graphene.ObjectType):
    name = graphene.NonNull(graphene.String)
```

The previous NonNull code snippet is also equivalent to:
```python
import graphene

class Character(graphene.ObjectType):
    name = graphene.String(required=True)
```

---
The second one keeps our code consistent since we're using it everywhere and there's no real advantage on using `NonNull` over `required`.

santtul: and fixed, you can have a look.

a-rmz (approved):

PR #272, opened 2020-09-03
Create forgotten helusers migration

santtul:
## Description :sparkles:
Create migration that was caused by `django-helusers` dependency update earlier.


igordavydsson (approved):
yes

PR #274, opened 2020-09-04
VEN-547 | Add a mutation to approve orders and send emails

a-rmz:
## Description :sparkles:
* Add a mutation to approve orders

The mutation takes a list of order `graphene.ID` and a `dueDate`. If no due date is passed, it takes the one already set on the `order`.

### Closes :no_good_woman:
**[VEN-547](https://helsinkisolutionoffice.atlassian.net/browse/VEN-547):**  mutation to finalize lease offer

### Automated tests :gear:️
```shell
```

Create an order, with a lease, and application. Then execute the following mutation:
```graphql
mutation ApproveOrders($input: ApproveOrderMutationInput!) {
  approveOrders(input: $input) {
    failedOrders {
      id
      error
    }
  }
}
```

```json
{
  "input": {
    "orders": [
      {
        "orderId": "",
        "email": "foo@bar.com"
      }
    ]
  }
}
```

## Screenshots :camera_flash:
image

### DIFF
--- a/payments/schema/mutations.py
+++ b/payments/schema/mutations.py
@@ -1,11 +1,18 @@
 import graphene
+from anymail.exceptions import AnymailError
+from dateutil.relativedelta import relativedelta
+from dateutil.utils import today
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError, transaction
 from django.utils.translation import gettext_lazy as _
+from django_ilmoitin.utils import send_notification
 
+from applications.enums import ApplicationStatus
+from applications.models import BerthApplication, WinterStorageApplication
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.schema import ProfileNode
+from leases.enums import LeaseStatus
 from leases.models import BerthLease, WinterStorageLease
 from leases.schema import BerthLeaseNode, WinterStorageLeaseNode
 from resources.schema import HarborNode, WinterStorageAreaNode
@@ -32,6 +39,7 @@ from .types import (
     AdditionalProductTaxEnum,
     BerthPriceGroupNode,
     BerthProductNode,
+    FailedOrderType,
     OrderLineNode,
     OrderNode,
     OrderStatusEnum,
@@ -479,6 +487,78 @@ class DeleteOrderLineMutation(graphene.ClientIDMutation):
         return DeleteOrderLineMutation()
 
 
+class OrderApprovalInput(graphene.InputObjectType):
+    order_id = graphene.ID(required=True)
+    email = graphene.String(required=True)
+
+
+class ApproveOrderMutation(graphene.ClientIDMutation):
+    class Input:
+        orders = graphene.List(OrderApprovalInput, required=True)
+        due_date = graphene.Date(description="Defaults to the Order due date")
+
+    failed_orders = graphene.List(FailedOrderType, required=True)
+
+    @classmethod
+    @change_permission_required(
+        Order,
+        BerthLease,
+        WinterStorageLease,
+        BerthApplication,
+        WinterStorageApplication,
+    )
+    def mutate_and_get_payload(cls, root, info, **input):
+        from ..notifications import NotificationType
+
+        failed_orders = []
+        due_date = input.get("due_date", today().date() + relativedelta(weeks=2))
+
+        for order_input in input.get("orders"):
+            order_id = order_input.get("order_id")
+            try:
+                with transaction.atomic():
+                    order = get_node_from_global_id(
+                        info, order_id, only_type=OrderNode, nullable=False,
+                    )
+
+                    # Update due date
+                    order.due_date = due_date
+                    order.save()
+
+                    # Update application status
+                    order.lease.application.status = ApplicationStatus.OFFER_SENT
+                    order.lease.application.save()
+
+                    # Update lease status
+                    order.lease.status = LeaseStatus.OFFERED
+                    order.lease.save()
+
+                    # Get payment URL
+                    language = (
+                        order.lease.application.language or settings.LANGUAGE_CODE
+                    )
+                    payment_url = get_payment_provider(
+                        info.context, ui_return_url=settings.VENE_UI_RETURN_URL
+                    ).get_payment_email_url(order, lang=language)
+
+                    # Send email
+                    email = order_input.get("email")
+                    context = {"order": order, "payment_url": payment_url}
+                    send_notification(
+                        email, NotificationType.ORDER_APPROVED.value, context, language
+                    )
+            except (
+                AnymailError,
+                OSError,
+                Order.DoesNotExist,
+                ValidationError,
+                VenepaikkaGraphQLError,
+            ) as e:
+                failed_orders.append(FailedOrderType(id=order_id, error=str(e)))
+
+        return ApproveOrderMutation(failed_orders=failed_orders)
+
+
 class Mutation:
     create_berth_product = CreateBerthProductMutation.Field(
         description="Creates a `BerthProduct` object."
@@ -578,6 +658,19 @@ class Mutation:
         "\n\n**Requires permissions** to edit payments."
         "\n\nErrors:"
         "\n* The passed `OrderLine` does not exist"
+    )
+
+    approve_orders = ApproveOrderMutation.Field(
+        description="Approves a list of Orders."
+        "\n\nIt receives a list of `GlobalID`s with the email to which the notification will be sent. "
+        "It also receives a `dueDate` for the orders. If no due date is passed, "
+        "it will take the due date set on the order."
+        "\n\nIt returns a list of failed order ids and the reason why they failed."
+        "\n\n**Requires permissions** to update orders, leases, and applications."
+        "\n\nErrors:"
+        "\n* The passed `order` does not exist"
+        "\n* An `order.lease` or `order.lease.application` have an invalid status transition"
+        "\n* There's an error when sending the email"
     )
 
 
--- a/payments/tests/conftest.py
+++ b/payments/tests/conftest.py
@@ -1,4 +1,6 @@
 import pytest
+from django.conf import settings
+from django_ilmoitin.models import NotificationTemplate
 from factory.random import randgen
 from requests import RequestException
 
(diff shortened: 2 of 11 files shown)


payments/schema/mutations.py
```diff
@@ -479,6 +485,77 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        BerthLease,
+        WinterStorageLease,
+        BerthApplication,
+        WinterStorageApplication,
+    )
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **input):
+        from ..notifications import NotificationType
+
+        failed_orders = []
```
santtul, line 512: Question(s): since the function is marked as `@transaction.atomic` all orders fail or all orders pass? can there be a case that 3 out of 5 orders will pass with data committed succesfully to db? if not, then I returning the `failed_orders` is useless.

If we would like to keep the `failed_orders` should we then adjust the transaction boundaries so that each order is processed in own separate transaction?

payments/tests/test_payments_mutations.py
```diff
@@ -1340,3 +1351,133 @@ def test_confirm_payment_invalid_status(old_schema_api_client, status):
     mock_call.assert_not_called()
     assert_in_errors("The order is not valid anymore", executed)
     assert_in_errors(status.label, executed)
+
+
+APPROVE_ORDER_MUTATION = """
```
santtul, line 1356: Would it make sense to create own test file for these e.g. `test_payments_mutations_approve_order.py` since the file is already huge?

payments/schema/mutations.py
igordavydsson: very good point, we should wrap the try/except below in the atomic transaction, not whole resolver

a-rmz: :big-hm: good catch. It makes sense to move the transaction to the `try/except` level.

However, after running a test in which one of them fails (raising `DoesNotExist`) and the other actually passes, the result is as expected. I.e. one `failedOrder` and one on the mailbox.

payments/tests/test_payments_mutations.py
a-rmz: On one hand I agree that it would be nice to split them somehow to avoid long extremely long files.
On the other, we don't have any other test file split by behaviour (?), but it could be a good start with the right approach.

payments/schema/mutations.py
igordavydsson: yes, `failedOrder` will be populated, and yes mailbox will get the email, but the tables for orders, leases and applications will not get the changes for the successful ones.

a-rmz: The test I added checks for
```python
assert order.due_date == due_date
assert order.lease.status == LeaseStatus.OFFERED
assert order.lease.application.status == ApplicationStatus.OFFER_SENT
```
and all of them passed 🤔 but I already moved the atomic inside, since it makes sense that it goes there.

payments/tests/test_payments_mutations.py
santtul: yeah, I think it would be a step to the right direction.

payments/schema/mutations.py
igordavydsson: and yes, even the python objects in memory will be with the changes, but if you refresh them from the DB, you will see that they did not get commited 😅

igordavydsson: ah, no, you had tests that either had all failed or all succeed, that's why

a-rmz: No no, it checks that one specific order is updated and one specific order fails. This is the full code test (without the setup)
```python
order = Order.objects.get(id=order.id)
assert len(executed["data"]["approveOrders"]["failedOrders"]) == 1
assert executed["data"]["approveOrders"]["failedOrders"][0] == {
    "id": failure_order_id,
    "error": "Order matching query does not exist.",
}
assert order.due_date == due_date
assert order.lease.status == LeaseStatus.OFFERED
assert order.lease.application.status == ApplicationStatus.OFFER_SENT
assert len(mail.outbox) == 1
assert (
    mail.outbox[0].subject
    == f"test order approved subject, event: {order.order_number}!"
)
assert mail.outbox[0].body == f"{ order.order_number } { payment_url }"
assert mail.outbox[0].to == [order.lease.application.email]
assert mail.outbox[0].alternatives == [
    (f"<b>{ order.order_number } { payment_url }</b>", "text/html")
]
```

igordavydsson: hm, don't see it here. do you mean you have it locally? and it works before you move the atomic to try/except level?

a-rmz: Yup, I added the test locally only (I'm pushing some changes now), and it worked even _before_ moving the atomic.

igordavydsson: 🤯 🙊

payments/tests/test_payments_mutations.py
a-rmz: Pushed 👍

payments/schema/mutations.py
a-rmz: Pushed the changes including the test

payments/tests/test_payments_mutations_approve_order.py
```diff
@@ -0,0 +1,214 @@
...
+        side_effect=AnymailError("Anymail error"),
+    ) as mock:
+        executed = superuser_api_client.execute(APPROVE_ORDER_MUTATION, input=variables)
+    mock.assert_called_once()
+
+    assert len(executed["data"]["approveOrders"]["failedOrders"]) == 1
+    assert executed["data"]["approveOrders"]["failedOrders"][0] == {
+        "id": order_id,
+        "error": "Anymail error",
+    }
```
santtul, line 161: Should we check here also that the order actually did not get updated?

payments/tests/test_payments_mutations_approve_order.py
```diff
@@ -0,0 +1,214 @@
...
+
+    assert len(executed["data"]["approveOrders"]["failedOrders"]) == 1
+    assert executed["data"]["approveOrders"]["failedOrders"][0] == {
+        "id": failure_order_id,
+        "error": "Order matching query does not exist.",
+    }
+
+    assert order.due_date == due_date
+    assert order.lease.status == LeaseStatus.OFFERED
+    assert order.lease.application.status == ApplicationStatus.OFFER_SENT
```
santtul, line 202: Should we check here also that the `failure_order_id ` actually did not get updated?

a-rmz: Good idea :Thu

payments/schema/mutations.py
```diff
@@ -479,6 +485,79 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        for order_input in input.get("orders"):
+            order_id = order_input.get("order_id")
+            with transaction.atomic():
+                try:
+                    order = get_node_from_global_id(
+                        info, order_id, only_type=OrderNode, nullable=False,
+                    )
+
+                    # Update due date
+                    if due_date:
```
igordavydsson, line 523: how about defaulting due date to today + 2 weeks? I am thinking that maybe some orders could be generated a while ago before they get sent, so due date originally assigned already passed, even though the order was not yet sent.

payments/tests/test_payments_mutations_approve_order.py
a-rmz: On this one we can't test because the order does not exist, but on the previous one yes

payments/schema/mutations.py
a-rmz: Makes sense, but in that case, we would never really use the `due date` that is generated with the order, it would always be overwritten by this.

igordavydsson: true, I was also thinking about that, not sure what is the best approach for this, allowing blank/null dates on the field does not feel like a good solution, so some default is needed.

a-rmz: We can follow this approach and just leave the other dates as backup

payments/tests/conftest.py
```diff
@@ -25,7 +26,13 @@
 from .utils import random_price, random_tax
 
 FAKE_BAMBORA_API_URL = "https://fake-bambora-api-url/api"
-UI_RETURN_URL = "https://front-end-url/{LANG}/"
+UI_RETURN_URL = "https://front-end-url/{LANG}"
```
igordavydsson, line 29: do we need this here, if you added it to 
```python
@pytest.fixture(autouse=True)
def force_settings(settings):
    settings.MAILER_EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"
    settings.DEFAULT_FROM_EMAIL = "noreply@foo.bar"
    settings.LANGUAGE_CODE = "en"
    settings.VENE_UI_RETURN_URL = "https://front-end-url/{LANG}"
```

igordavydsson: or is this used for testing that the URL gets built properly in different places?

a-rmz: It's being used in a couple of places, namely, for the `DummyContext` on the notifications and in some of the tests. We probably could remove it and always get the value from `settings` 🤔

igordavydsson: easier to keep track of only one var vs two vars :d0rp:

payments/tests/test_payments_mutations_approve_order.py
```diff
@@ -0,0 +1,222 @@
...
+                "orderId": to_global_id(OrderNode, order.id),
+                "email": order.lease.application.email,
+            }
+        ],
+    }
+    api_client.execute(APPROVE_ORDER_MUTATION, input=variables)
+
+    order = Order.objects.get(id=order.id)
+
+    assert order.due_date == today().date() + relativedelta(weeks=2)
```
igordavydsson, line 97: would be nice the assert that initially `order.due_date != today().date() + relativedelta(weeks=2)`, but only after the mutation it bacame that

a-rmz: Sure!

igordavydsson (approved):
Looking good now 🤩 Very good job! 👍 🎖️ 

applause

santtul (approved):
Great!

PR #276, opened 2020-09-08
Remove test reference on payments dummy context

a-rmz:
## Description :sparkles:
The config taken from `payments/tests/conftest.py` used on `payments/notifications.py` make the application crash because `conftest` imports `pytest`, which is not present on non-dev environments.

By replacing the imported conf values by hard-coded ones, we avoid problems.


santtul (approved):
👍

PR #277, opened 2020-09-09
VEN-829 | Send different email notification for umarked ws application

santtul:
## Description :sparkles:
- Send different email notification for umarked ws application
- Improve readme, add instructions for test running

### Closes :no_good_woman:
[VEN-829](https://helsinkisolutionoffice.atlassian.net/browse/VEN-829)

### Automated tests :gear:️
    pytest applications/tests/test_applications_notifications.py::test_unmarked_winter_application_created_notification_is_sent
    pytest applications/tests/test_application_models.py::test_winterstorage_application_is_not_unmarked
    pytest applications/tests/test_application_models.py::test_unmarked_winterstorage_application_is_unmarked

### DIFF
--- a/applications/constants.py
+++ b/applications/constants.py
@@ -1,0 +1,2 @@
+MARKED_WS_SENDER = "CreateMarkedWinterStorageApplication"
+UNMARKED_WS_SENDER = "CreateUnmarkedWinterStorageApplication"
--- a/applications/models.py
+++ b/applications/models.py
@@ -281,6 +281,15 @@ class WinterStorageApplication(BaseApplication):
         verbose_name=_("trailer registration number"), max_length=64, blank=True
     )
 
+    def is_unmarked_ws_application(self) -> bool:
+        first_area = self.chosen_areas.first()
+
+        return bool(
+            self.chosen_areas.count() == 1
+            and first_area.number_of_unmarked_spaces
+            and first_area.number_of_unmarked_spaces > 0
+        )
+
     def get_notification_context(self):
         return {
             "created_at": localize_datetime(self.created_at, self.language),
--- a/README.md
+++ b/README.md
@@ -115,3 +115,9 @@ Lastly, you can load the User Groups:
 And install the model permissions:
 
     ./manage.py set_group_model_permissions
+
+## Running tests
+
+    pytest
+    
+In order to successfully run tests in ```applications/tests/test_applications_notifications.py``` you need to set env variable ```NOTIFICATIONS_ENABLED=1```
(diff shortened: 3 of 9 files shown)


applications/models.py
```diff
@@ -281,6 +281,17 @@ class WinterStorageApplication(BaseApplication):
         verbose_name=_("trailer registration number"), max_length=64, blank=True
     )
 
+    def is_unmarked_ws_application(self) -> bool:
```
igordavydsson, line 284: slightly more powerful result if you mark it as `@property`. you could use it as field in gql without custom resolver then

applications/constants.py
```diff
@@ -0,0 +1,2 @@
+MARKED_WS_SENDER = "CreateMarkedWinterStorageApplication"
+UNMARKED_WS_SENDER = "CreateUnmarkedWinterStorageApplication"
```
igordavydsson, line 2: very nice solution 👍

applications/models.py
```diff
@@ -281,6 +281,17 @@ class WinterStorageApplication(BaseApplication):
...
+        first_area = self.chosen_areas.first()
+
+        if (
+            self.chosen_areas.count() == 1
+            and first_area.number_of_unmarked_spaces
+            and first_area.number_of_unmarked_spaces > 0
+        ):
+            return True
+        return False
```
a-rmz, line 293:
```suggestion
        return (
            self.chosen_areas.count() == 1
            and first_area.number_of_unmarked_spaces
            and first_area.number_of_unmarked_spaces > 0
        )
```

applications/models.py
santtul: I think we want to publish the field only when we create the new field for WS application type (the enum). So there is no reason to publish it now in gql or what do you think?

applications/models.py
igordavydsson: or maybe what Alex suggested but with `bool` = `return bool(...)`

igordavydsson: if you want it to be more explicit, but Alex' suggestion is more pythonic

applications/models.py
igordavydsson: no, setting it as property will nto automatically publish it, but it would make using it in someinstances easier. OTOH, we will probably add a new enum field, which will be published in the gql api, and not this property method, since it is more efficient that way, so not really that much use for the property except for checking the type before saving the new future enum field.

applications/models.py
santtul: 👍 done

igordavydsson (approved):
LGTM 👍 

very nice

PR #282, opened 2020-09-17
VEN-823 | Add WS application area type

santtul:
## Description :sparkles:
- Add new enum column for WS application
- Enable searching by that
- Add command to update existing applications

### Closes :no_good_woman:
**[VEN-823](https://helsinkisolutionoffice.atlassian.net/browse/VEN-823)** 

### Automated tests :gear:️
`applications/tests/test_applications_new_schema_ws_queries.py::test_query_with_area_type_filter`
`applications/tests/test_application_models.py`

    python manage.py migrate
    ./manage.py assign_application_area_type

query via `graphql_v2`
```
{
  winterStorageApplications(first: 10, areaType: "marked") {
    count
    edges {
      node {
        id
        areaType
        __typename
      }
      __typename
    }
    __typename
  }
}
```


applications/management/commands/assign_application_area_type.py
```diff
@@ -0,0 +1,32 @@
...
+            created_at__gt=date(2020, 9, 14)
+        )
+
+        for application in new_applications:
+            application.area_type = application.resolve_area_type()
+            application.save()
+
+        # Get ws applictions created before 14.9.2020
+        old_applications = WinterStorageApplication.objects.filter(
+            created_at__lt=date(2020, 9, 14)
```
igordavydsson, line 24: we probably need to use `__lte`, since otherwise applications left on 14.9 will be missed.

applications/management/commands/assign_application_area_type.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+
+        for application in old_applications:
+            # Mark all older applications as MARKED
+            application.area_type = ApplicationAreaType.MARKED
+            application.save()
```
igordavydsson, line 30: and since here you simply set the same value for every object, you could do:
```python
WinterStorageApplication.objects.filter(
    created_at__lt=date(2020, 9, 14)
).update(area_type=ApplicationAreaType.MARKED)
```

applications/new_schema/types.py
```diff
@@ -123,6 +123,9 @@ class WinterStorageApplicationFilter(django_filters.FilterSet):
     no_customer = django_filters.BooleanFilter(
         field_name="customer", lookup_expr="isnull"
     )
+    area_type = django_filters.ChoiceFilter(
+        choices=[(entry, entry.value) for entry in ApplicationAreaType]
+    )
```
igordavydsson, line 128: i think a cleaner approach if we use graphene enum here, like we do for orders with [`OrderTypeEnum`](https://github.com/City-of-Helsinki/berth-reservations/blob/6da06ad55793248cb6d822f7afa689006ac6e7c3/payments/schema/queries.py#L64)

igordavydsson: this way we ensure that the client know what values are allowed for this filter and they are parsed correctly

applications/management/commands/assign_application_area_type.py
santtul: the expressions seem to produce following SQLs:
`created_at__gt=date(2020, 9, 14) -> WHERE "applications_winterstorageapplication"."created_at" > '2020-09-14T00:00:00+03:00'`
`created_at__lt=date(2020, 9, 14) -> WHERE "applications_winterstorageapplication"."created_at" < '2020-09-14T00:00:00+03:00'`
so I think we are ok?

santtul: well it misses the application(s) that are created exactly at `2020-09-14T00:00:00+03:00` :D I'll add `lte` to second of so that it's bullet proof

igordavydsson: hmmm, seems like `__gt` is essentially `__gte`, while `__lt` is as expected `__lt`. maybe then your current solution is ok? can you double check that it catches the applications ON the day we use for these lookups?

applications/management/commands/assign_application_area_type.py
santtul: Good catch, done.

applications/new_schema/types.py
santtul: done

applications/management/commands/assign_application_area_type.py
a-rmz: I guess there's no harm in adding the `__lte` 🤷

a-rmz (approved):

applications/management/commands/assign_application_area_type.py
santtul: agree what @a-rmz said and dubbel checked that it does catch the applications that are created on that day (2020-09-14). So I think we are good to go.

igordavydsson (approved):
Looks good! 👍

PR #285, opened 2020-09-18
Improve customer ui schema tests

santtul:
## Description :sparkles:
- Improve customer ui related queries test coverage: execute the same queries that customer ui is executing
- Assert for ws application area type


harbors/tests/test_harbors_graphql.py
```diff
@@ -1,31 +1,120 @@
...
+
+
+def test_get_ws_areas(boat_type, old_schema_api_client):
+    area = WinterStorageAreaFactory()
+
+    executed = old_schema_api_client.execute(CUSTOMER_UI_WS_AREAS_QUERY)
+
+    area_node = executed["data"]["winterStorageAreas"]["edges"][0]["node"]
+    assert area_node["id"] == to_global_id(WinterStorageAreaType._meta.name, area.id)
+    assert area_node["properties"]["name"] == area.name
```
a-rmz, line 120: Since we're querying for all the fields, would it make sense to assert that all of them are correct?

You can probably reuse some of the assertions from [resources/tests/test_resources_queries.py#L301](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/resources/tests/test_resources_queries.py#L301)

a-rmz:
Could you also rename the test file to `test_harbors_queries.py` to make it consistent with the other test files? :)

santtul:
Good catch, renamed the test file.

harbors/tests/test_harbors_graphql.py
santtul: This PR was a first step that ensures that the query doesn't blow up. We can continue improving these tests when we have time.

a-rmz: Alright! Let's keep it in mind then, but good progress already. :)

a-rmz (approved):

PR #280, opened 2020-09-15
VEN-815 | Mail notification templates

a-rmz:
## Description :sparkles:
* Adapt the `Order approved` notification type for the email template
* Add the `Berth/WS Application approved` notification type (will be used in the future)
* Add a `DummyContext` for the `Order approved` notifications
* Fix a weak query for `ContentType` that was making the tests break

### Closes :no_good_woman:
**[VEN-815](https://helsinkisolutionoffice.atlassian.net/browse/VEN-815):** Insert email templates into the database

### Related :handshake:
#274: VEN-547 | Add a mutation to approve orders and send emails

### Automated tests :gear:️
```shell
$ pytest
```

## Additional notes :spiral_notepad:
This changes are required to send the invoice emails, the notification templates will later be added to the database.

PR #286, opened 2020-09-18
VEN-815 | Order notifications

a-rmz:
## Description :sparkles:
* Add an `order_type` prop to the `Order`s to determine which email should be used

The criteria to determine the `order type` is as follows:
* `NEW_BERTH_ORDER` = The order has a `berth lease` and no previous leases for that customer/berth
* `RENEW_BERTH_ORDER` = The order has a `berth lease` and has previous leases for that customer/berth
* `BERTH_SWITCH_ORDER` = The `order.lease.application` has a `berth switch`
* `WINTER_STORAGE_ORDER` = The order has a `winter storage lease`
* `UNMARKED_WINTER_STORAGE_ORDER` = The order has an winter storage application with `area_type == UNMARKED`
* `INVALID` = The order is not associated with any lease

### Closes :no_good_woman:
**[VEN-815](https://helsinkisolutionoffice.atlassian.net/browse/VEN-815):** Insert email templates into the database

### Related :handshake:
#280: Mail notification templates

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_order_types.py
```

1. Add `Notification`s on the [django admin notifications](http://localhost:8000/admin/django_ilmoitin/notificationtemplate/). You can see the required context fields on the [payments/notifications/dummy_context.py](https://github.com/City-of-Helsinki/berth-reservations/compare/ven-815/order-notifications?expand=1#diff-48a9f92debcc403210e7a307f3c98446)
2. Open the `Preview` for each of them

## Screenshots :camera_flash:
**Previewed Unmarked Winter Storage invoice**
image

The actual templates are built and compiled separately and inserted into the database manually

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -9,6 +9,8 @@ from django.db import models
 from django.db.models import Q, UniqueConstraint
 from django.utils.translation import gettext_lazy as _
 
+from applications.enums import ApplicationAreaType
+from applications.models import BerthApplication, WinterStorageApplication
 from leases.enums import LeaseStatus
 from leases.models import BerthLease, WinterStorageLease
 from utils.models import TimeStampedModel, UUIDModel
@@ -17,6 +19,7 @@ from utils.numbers import rounded as rounded_decimal
 from .enums import (
     AdditionalProductType,
     OrderStatus,
+    OrderType,
     PeriodType,
     PriceUnits,
     ProductServiceType,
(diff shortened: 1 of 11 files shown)


payments/models.py
```diff
@@ -295,6 +298,51 @@ class Meta:
     def __str__(self):
         return f"{self.product} [{self.status}]"
 
+    @property
+    def order_type(self) -> OrderType:
```
santtul, line 302: Should this be a DB enum column instead? The db column would be more future proof, performant and would enable searching by order_type in case we want it in upcoming "invoices list" view. We can of course refactor this to DB column later.

payments/schema/mutations.py
```diff
@@ -554,9 +554,27 @@ def mutate_and_get_payload(cls, root, info, **input):
                         ),
                         "payment_url": payment_url,
                     }
-                    send_notification(
-                        email, NotificationType.ORDER_APPROVED.value, context, language
-                    )
+
+                    if order.order_type == OrderType.NEW_BERTH_ORDER:
```
santtul, line 558: Could we extract function for cleaner code here:
`notification_type = get_notification_type(order.order_type)`
?

santtul (changes requested):
Looks good! some suggestions.

payments/models.py
a-rmz: I think the ideal approach would be to add it as an annotation (like we have done with available berths or active leases), but I think it will be a bit complex, so it will take some time to implement it.

santtul (approved):
Great work!

PR #289, opened 2020-09-21
VEN-630 | Hotfix | Timmi import

a-rmz:
## Description :sparkles:
* Check that lease dates don't overlap (allows creating leases on the past)
* Only create leases for paid orders (Add a check for importing leases)
* Only add leases for existing berths (Check if the berth exists, could be the case that it doesn't)
* Update the fixtures to add missing berths

### Related :handshake:
**[VEN-630](https://helsinkisolutionoffice.atlassian.net/browse/VEN-630):** import customers' data in production

### Automated tests :gear:️
```shell
$ pytest
```


santtul (approved):
Ship it! 👍

PR #300, opened 2020-09-25
VEN-684 | Add pricing rules for organization customers

santtul:
## Description :sparkles:
- Company customers pay x2 prices
- Non-billable customers do not pay at all, order should show all prices and taxes as 0 and get marked PAID upon creation

### Closes :no_good_woman:
**[VEN-684](https://helsinkisolutionoffice.atlassian.net/browse/VEN-684)** 

### Automated tests :gear:️
Several new tests in:
`leases/tests/test_lease_mutations.py`
`payments/tests/test_payments_models.py`

It's hard to test without the whole payment process in place, should be tested once it's implemented.

### DIFF
--- a/customers/models.py
+++ b/customers/models.py
@@ -258,6 +258,12 @@ class CustomerProfile(TimeStampedModel):
         verbose_name_plural = _("customer profiles")
         ordering = ("id",)
 
+    def is_non_billable_customer(self):
+        return (
+            hasattr(self, "organization")
+            and self.organization.organization_type == OrganizationType.NON_BILLABLE
+        )
+
     def __str__(self):
         if self.user:
             return "{} {} ({})".format(
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -8,6 +8,7 @@ from applications.models import BerthApplication, WinterStorageApplication
 from applications.new_schema import BerthApplicationNode, WinterStorageApplicationNode
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.models import CustomerProfile
+from payments.enums import OrderStatus
 from payments.models import BerthPriceGroup, BerthProduct, Order, WinterStorageProduct
 from resources.schema import BerthNode, WinterStoragePlaceNode, WinterStorageSectionNode
 from users.decorators import (
@@ -91,7 +92,7 @@ class CreateBerthLeaseMutation(graphene.ClientIDMutation):
                 else price_group_products.get(harbor__isnull=True)
             )
 
-            Order.objects.create(
+            order = Order.objects.create(
                 customer=input["customer"], lease=lease, product=product
             )
         except BerthProduct.DoesNotExist as e:
@@ -103,6 +104,9 @@ class CreateBerthLeaseMutation(graphene.ClientIDMutation):
 
         application.status = ApplicationStatus.OFFER_GENERATED
         application.save()
+
+        if order.customer.is_non_billable_customer():
+            order.set_status(OrderStatus.PAID, "Non-billable customer.")
 
         return CreateBerthLeaseMutation(berth_lease=lease)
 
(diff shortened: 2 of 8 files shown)


customers/models.py
```diff
@@ -258,6 +258,12 @@ class Meta:
         verbose_name_plural = _("customer profiles")
         ordering = ("id",)
 
+    def is_non_billable_customer(self):
```
a-rmz, line 261: Would it make sense to turn this into a `@prop` instead?

leases/schema/mutations.py
```diff
@@ -104,6 +105,9 @@ def mutate_and_get_payload(cls, root, info, **input):
         application.status = ApplicationStatus.OFFER_GENERATED
         application.save()
 
+        if order.customer.is_non_billable_customer():
+            order.set_status(OrderStatus.PAID, "Non-billable customer.")
```
a-rmz, line 109:
```suggestion
            order.set_status(OrderStatus.PAID, _("Non-billable customer."))
```

leases/schema/mutations.py
```diff
@@ -265,6 +269,9 @@ def mutate_and_get_payload(cls, root, info, **input):
...
```
a-rmz, line 273:
```suggestion
            order.set_status(OrderStatus.PAID, _("Non-billable customer."))
```

payments/utils.py
```diff
@@ -156,6 +157,21 @@ def calculate_product_percentage_price(base_price, percentage):
     return base_price * (percentage / 100)
 
 
+def calculate_organization_price(price, organization_type: OrganizationType) -> Decimal:
+    if organization_type == OrganizationType.COMPANY:
+        return price * 2
+    elif organization_type == OrganizationType.NON_BILLABLE:
+        return Decimal("0.00")
+    else:
+        return price
```
a-rmz, line 166: Would it make more sense to do the
```python 
return rounded_price(price)
```

instead of later having to call it? Since we already know it's gonna be a price
```python
organization_price = calculate_organization_price(
    price_value, organization_type
)

self.price = rounded_decimal(organization_price)
```

payments/models.py
```diff
@@ -676,36 +693,39 @@ def save(self, *args, **kwargs):
...
-                    elif self.product.period == PeriodType.YEAR:
-                        price = calculate_product_partial_year_price(
-                            price,
-                            self.order.lease.start_date,
-                            self.order.lease.end_date,
-                        )
+            if hasattr(self.order.customer, "organization"):
+                organization_type = self.order.customer.organization.organization_type
+
+                self.price = calculate_organization_price(price, organization_type)
```
a-rmz, line 723: Not calling the `round_price` here

payments/tests/test_payments_models.py
```diff
@@ -915,7 +915,7 @@ def test_berth_season_price(berth):
     assert order.price == Decimal("15.00")
 
 
-def test_order_winter_storage_lease_with_area_right_price(
+def test_order_winter_storage_lease_with_section_right_price(
```
a-rmz, line 918: Good catch :derp:

customers/models.py
santtul: I think there is no benefit because we are not exposing the field. what do you think / what would be the benefit?

payments/models.py
santtul: It was not called here before, should it be called?

payments/utils.py
santtul: That function is used in two places where another one (in OrderLine) does not call rounding, so that's why it doesn't do the rounding.

customers/models.py
a-rmz: 🤔 Good point. In my mind, it is a property of the customer, meaning that the customer `either is a billable customer or not`, but there's no real benefit/downside of keeping it as either,

payments/utils.py
a-rmz: Let's skip the later call and add it here instead.

payments/models.py
a-rmz: I think so, it should be `rounded` right before being set to ensure that it keeps the `two rounded decimals` rule. If we update the called function, this can stay as-is.

payments/utils.py
santtul: done

payments/models.py
santtul: done

a-rmz (approved):
Good stuff!

PR #305, opened 2020-10-05
VEN-812 | Handle repeated payment attempts

a-rmz:
## Description :sparkles:
* Add an `OrderToken` model to store the payment tokens retrieved from Bambora
* Check if a valid token exists before trying to initiate the payment flow

### Closes :no_good_woman:
**[VEN-812](https://helsinkisolutionoffice.atlassian.net/browse/VEN-812):** Handle repeated payment attempts

### Automated tests :gear:️
```shell
$ python manage.py payments/tests/test_bambora_payform.py::test_initiate_duplicated_payment
$ python manage.py payments/tests/test_bambora_payform.py::test_initiate_duplicated_payment_new_token_after_expiry
```

1. Create an order
2. Go to `https://venepaikka.test.kuva.hel.ninja/fi/payment?order_number=XXXXXX` for your order
3. Click `Jatka maksamaan`
4. Close the Bambora tab
5. Open the Customer UI page again and click `Jatka maksamaan` again
6. The Bambora order should still be payable


payments/providers/bambora_payform.py
```diff
@@ -97,7 +119,15 @@ def handle_initiate_payment(self, order: Order, response) -> str:
         if result == 0:
             # Create the URL where user is redirected to complete the payment
             # Append "?minified" to get a stripped version of the payment page
-            return self.url_payment_token.format(token=response["token"])
+            token = response["token"]
+
+            if order_token := (
+                OrderToken.objects.filter(order=order).order_by("created_at").first()
```
santtul, line 125: Would it make sense to pass the created `order_token.id` to this function and update based on id here?

santtul (commented):
Looks good, one question added.

payments/providers/bambora_payform.py
a-rmz: That would be an alternative. I tried to keep the function signatures the same (to try to decouple the `BaseProvider` with the `OrderToken` flow) to properly implement it, but we could also add it and save one call to the db 🤔

santtul: I think there is no reason to try to hide the dependency to `OrderToken`, instead we should make it more visible to make it easier to understand and follow.

a-rmz: 🤔 alrighty then

santtul (approved):
Great!

PR #313, opened 2020-10-08
VEN-891 | Fixed season prices

a-rmz:
## Description :sparkles:
* Remove partial price calculation for season products

### Closes :no_good_woman:
**[VEN-891](https://helsinkisolutionoffice.atlassian.net/browse/VEN-891):** Full price for season products

### Automated tests :gear:️
```shell
$ pytest
```

1. Create an application and lease for a berth/WS place
2. Create a berth/ws product for the place you selected
3. From the Django admin, go to create an order and assign it to the lease and product you generated
4. Hit `Save and continue editing`
5. The price should be the same as the product if you created a berth order. If you created a ws order, it would be equivalent to `boat lenght * width * product price`
6. Add an `OrderLine` for a `season` product
7. The price should be the same for the order line, the `order` price should've changed and would be `order price + order line price`

## Screenshots :camera_flash:
Working example:
image

### DIFF
--- a/leases/tests/test_lease_mutations.py
+++ b/leases/tests/test_lease_mutations.py
@@ -17,7 +17,6 @@ from berth_reservations.tests.utils import (
 from customers.schema import BoatNode, ProfileNode
 from payments.models import BerthPriceGroup, Order
 from payments.tests.factories import BerthProductFactory, WinterStorageProductFactory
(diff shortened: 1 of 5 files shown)


santtul (approved):
Great stuff! Always nice to see things getting more simple 😄

PR #315, opened 2020-10-09
Hotifx | Remove payment cancel from notify

a-rmz:
## Description :sparkles:
We disabled cancelling payments from Bambora `success` request, but forgot about the `notify` request. In some cases where the payment was cancelled, the `Order` (and lease and application) would be marked as `rejected`.

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_bambora_payform.py::test_handle_notify_request_payment_failed
```

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -581,11 +581,7 @@ class Order(UUIDModel, TimeStampedModel):
             return
 
         valid_status_changes = {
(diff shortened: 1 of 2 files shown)


santtul (approved):
Makes sense!

PR #324, opened 2020-10-13
Chore | Fix pipeline 

tuovinensanttu:
## Description :sparkles:
Rename gitlab-config file.
### Closes :no_good_woman:


tebbon (approved):
Looks good now. The actual deployment might still fail, but the pipeline is now triggered and we can proceed. Thanks for the help!

PR #326, opened 2020-10-14
VEN-883 | Add mutation to cancel an order

a-rmz:
## Description :sparkles:
* Add a mutation to cancel an order (intended to be used by the customer UI)

### Closes :no_good_woman:
**[VEN-883](https://helsinkisolutionoffice.atlassian.net/browse/VEN-883):** Mutation to cancel orders

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_mutations.py::test_cancel_order
$ pytest payments/tests/test_payments_mutations.py::test_cancel_order_invalid_status
$ pytest payments/tests/test_payments_mutations.py::test_cancel_order_does_not_exist
```

### DIFF
--- a/payments/schema/mutations.py
+++ b/payments/schema/mutations.py
@@ -416,6 +416,28 @@ class ConfirmPaymentMutation(graphene.ClientIDMutation):
         return ConfirmPaymentMutation(url=payment_url)
 
 
+class CancelOrderMutation(graphene.ClientIDMutation):
+    class Input:
+        order_number = graphene.String(required=True)
+
+    @classmethod
+    def mutate_and_get_payload(cls, root, info, **input):
+        order_number = input.get("order_number", None)
+        try:
+            order = Order.objects.get(order_number=order_number)
+            if order.status != OrderStatus.WAITING:
+                raise VenepaikkaGraphQLError(
+                    _("The order is not valid anymore")
+                    + f": {OrderStatus(order.status).label}"
+                )
+            order.set_status(OrderStatus.REJECTED, _("Order rejected by customer"))
+            order.invalidate_tokens()
+        except (Order.DoesNotExist, ValidationError) as e:
+            raise VenepaikkaGraphQLError(e)
+
+        return CancelOrderMutation()
+
+
 class OrderLineInput:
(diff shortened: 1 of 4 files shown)


payments/schema/mutations.py
```diff
@@ -416,6 +416,28 @@ def mutate_and_get_payload(cls, root, info, **input):
...
 
+class CancelOrderMutation(graphene.ClientIDMutation):
+    class Input:
+        order_number = graphene.String(required=True)
+
+    @classmethod
+    def mutate_and_get_payload(cls, root, info, **input):
+        order_number = input.get("order_number", None)
+        try:
+            order = Order.objects.get(order_number=order_number)
```
tebbon, line 427: The order numbers are sufficiently random hopefully, so that you can't just guess random orders to cancel?

tebbon (approved):
Looks good.

payments/schema/mutations.py
a-rmz: [Should be](https://github.com/City-of-Helsinki/berth-reservations/blob/develop/payments/utils.py#L183-L186)
```python
def generate_order_number() -> str:
    t = time.time() * 1000000 * random.uniform(1, 1000)
    b = base64.b32encode(struct.pack(">Q", int(t)).lstrip(b"\x00")).strip(b"=").lower()
    return b.decode("utf8")
```

payments/schema/mutations.py
```diff
@@ -416,6 +416,28 @@ def mutate_and_get_payload(cls, root, info, **input):
         return ConfirmPaymentMutation(url=payment_url)
 
 
+class CancelOrderMutation(graphene.ClientIDMutation):
+    class Input:
+        order_number = graphene.String(required=True)
+
+    @classmethod
+    def mutate_and_get_payload(cls, root, info, **input):
```
ahie, line 424: Should this be decorated with `@change_permission_required(Order)`?

a-rmz: It's a public endpoint, so we can't check if the user has permissions.

ahie: Ooh right, mixed this with the discussion about cancelling sent out orders from the admin side :)

PR #336, opened 2020-10-19
VEN-922 | Ignore empty OrderTokens when building the payment URL

a-rmz:
## Description :sparkles:
* Exclude empty tokens when handling the `initiate_payment` flow

If the response from bambora was failing (either a `PayloadValidation` or service error), the token was saved as-is, so with a blank token value. When the client tried to get the payment link again, it would find that it had a valid token and return that URL to the user, however the token was empty and that’s why it was failing.

### Closes :no_good_woman:
**[VEN-922](https://helsinkisolutionoffice.atlassian.net/browse/VEN-922)** 

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_mutations.py::test_payment_fails_doesnt_use_empty_token
```

### DIFF
--- a/payments/providers/bambora_payform.py
+++ b/payments/providers/bambora_payform.py
@@ -73,13 +73,22 @@ class BamboraPayformProvider(PaymentProvider):
             today() + relativedelta(days=6), datetime.max.time()
         )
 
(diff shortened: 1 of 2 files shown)


santtul (approved):
Great fix!

PR #339, opened 2020-10-20
Add postgresql-client to docker image

tebbon:
Add postgresql-client to docker image, needed for the staging and prod db migrations. Also upgrade local postgres docker to use the same version as in prod.


santtul (approved):
👍

PR #345, opened 2020-10-22
VEN-948 | Check migrations on initialize

santtul:
Check migrations on container start.


a-rmz (approved):

.prod/on_initialize.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/sh
+set -e
+
+# Check that migrations are in sync with models
+python /app/manage.py migrate --check
```
tebbon, line 5: What is the purpose of this? This script is only ever ran on a completely empty database, so this will always fail?

tebbon (changes requested):

.prod/on_initialize.sh
santtul: It would fail if the Python code is not in sync with the generated migrations (i.e. there are pending migrations you should create).

tebbon: I tested this locally, and it also fails if the db is empty, which is always the case when this script is ran.

tebbon: Maybe `/app/manage.py makemigrations --check` is what you mean?

a-rmz: AHA! You're right, it's `makemigrations`

tebbon: that also btw fails for this branch, but for a reason most likely, as it actually makes a migration file.
I propose `python ./manage.py makemigrations --verbosity 3 --dry-run --noinput --check`.

santtul: great feedback 👍  yes, the PR is not yet rebased, so it should still fail. will adjust the PR and see how it works.

tebbon (approved):
nice!

PR #342, opened 2020-10-20
VEN-933 | Add missing ProductServiceType

a-rmz:
## Description :sparkles:
*  Add missing ProductServiceType (`Storage on ice`) and translations

### Closes :no_good_woman:
**[VEN-933](https://helsinkisolutionoffice.atlassian.net/browse/VEN-933):** Add missing product services

### Related 🤝 
**[VEN-814](https://helsinkisolutionoffice.atlassian.net/browse/VEN-814):** Pricing page: support "Add New" additional services

### Automated tests :gear:️
N/A

### Manual testing :construction_worker_man:
1. Go to add a new `AdditionalProduct`
http://localhost:8000/admin/payments/additionalproduct/add/
2. You should see the `Storage on ice` option

## Screenshots :camera_flash:
image

PR #346, opened 2020-10-23
VEN-933 | Add a missing migration for unique services

a-rmz:
## Description :sparkles:
* Replace the constraint values from enum values to plain text values

### Related :handshake:
**[VEN-933](https://helsinkisolutionoffice.atlassian.net/browse/VEN-933):** Add missing product services
#342


santtul (approved):
👍

PR #350, opened 2020-10-29
Add production force_script_name config

tebbon:
Allows forcing the app path to the one required by api.hel.fi reverse proxy.


a-rmz (approved):

santtul (approved):
👍

PR #352, opened 2020-10-29
add production csrf config

tebbon:
Add csrf header and x-forwarded-for settings and session cookie settings to allow django admin to work under the api.hel.fi reverse proxy.

* Ship to production, and check that django admin ui works.


a-rmz (approved):

santtul (approved):
👍

PR #360, opened 2020-11-02
Update docker-compose.yml to use correct target for django service

ahie:
Without this target, dev requirements will not be installed due to changes in the Dockerfile.


santtul (approved):
Looks good, but please rebase.

ahie:
Already included in #364.

PR #363, opened 2020-11-04
VEN-925 | Add mutation for creating additional product orders

santtul:
## Description :sparkles:
- Add CreateAdditionalProductOrderMutation mutation
- Change payment models to support ordering only additional product

### Closes :no_good_woman:
**[VEN-925](https://helsinkisolutionoffice.atlassian.net/browse/VEN-925)** 

### Automated tests :gear:️
New test suite:
`payments/tests/test_payments_additional_product_order.py`
Several new test cases in:
`payments/tests/test_bambora_payform.py`
`payments/tests/test_payments_order_types.py`
`payments/tests/test_payments_queries.py`

When PR is merged we should smoke test in staging that payment process still works for normal orders.

There will be more PRs like approving/sending the additional product invoice.


leases/models.py
```diff
@@ -53,7 +54,7 @@ class Meta:
 
     @property
     def order(self):
-        return self._orders_relation.first()
+        return self._orders_relation.filter(order_type=OrderType.LEASE_ORDER).first()
```
a-rmz, line 57: Not sure if it would make sense to keep this prop now that leases can have multiple orders. Let's rename the `_orders_relation` field to just `orders` so we can do `lease.orders.first()` if we need this.

(In any case, we're not using it anywhere.)

payments/models.py
```diff
@@ -727,6 +736,18 @@ def save(self, *args, **kwargs):
 
         super().save(*args, **kwargs)
 
+    @staticmethod
+    def _calculate_percentage_price_for_additional_prod_order(
+        lease: BerthLease, percentage
+    ):
+        lease_order = lease._orders_relation.filter(
+            status=OrderStatus.PAID, order_type=OrderType.LEASE_ORDER
+        ).first()
```
a-rmz, line 745: Here we could leverage the `lease.orders` making it clearer

payments/schema/mutations.py
```diff
@@ -333,6 +335,53 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+    @classmethod
+    @add_permission_required(Order)
+    @view_permission_required(BerthLease, WinterStorageLease)
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **input):
+        customer_id = input.pop("customer_id")
+        customer = get_node_from_global_id(
+            info, customer_id, ProfileNode, nullable=False
+        )
```
a-rmz, line 354:
```suggestion
        customer = get_node_from_global_id(
            info, input.pop("customer_id"), ProfileNode, nullable=False
        )
```

payments/schema/mutations.py
```diff
@@ -333,6 +335,53 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+        additional_product_id = input.pop("additional_product_id", None)
+        additional_product = get_node_from_global_id(
+            info, additional_product_id, AdditionalProductNode, nullable=False
+        )
```
a-rmz, line 359: No need for `.pop("", None)`. Because it's a required field, it will fail if it doesn't have a value.

```suggestion
        additional_product = get_node_from_global_id(
            info, input.pop("additional_product_id"), AdditionalProductNode, nullable=False
        )
```

payments/schema/mutations.py
```diff
@@ -333,6 +335,53 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+        lease_id = input.pop("lease_id", None)
+        lease = get_node_from_global_id(info, lease_id, BerthLeaseNode, nullable=False)
```
a-rmz, line 362: Same here.

```suggestion
        lease = get_node_from_global_id(info, input.pop("lease_id"), BerthLeaseNode, nullable=False)
```

payments/tests/conftest.py
```diff
@@ -104,6 +105,15 @@ def _generate_order(order_type: str = None):
...
         )
+    elif order_type == "additional_product_order":
+        order = OrderFactory(
+            order_type=OrderType.ADDITIONAL_PRODUCT_ORDER,
+            customer=customer_profile,
+            price=random_price(),
+            tax_percentage=random_tax(),
+            product=None,
+            lease=None,
+        )
```
a-rmz, line 116: To ensure that the order will have the additional product.
```suggestion
        order = OrderFactory(
            order_type=OrderType.ADDITIONAL_PRODUCT_ORDER,
            customer=customer_profile,
            price=random_price(),
            tax_percentage=random_tax(),
            product=None,
            lease=None,
        )
        OrderLineFactory(
             order=order
         )             
```

a-rmz (changes requested):
Looking good so far

leases/models.py
santtul: The FE is using the `order` when displaying the invoice on application page.

payments/tests/conftest.py
santtul: Tried adding it but it opens kinda worm box with the model validations (starting from a lease and a customer) so IMO not worth doing.

a-rmz (commented):
The `ApproveOrderMutation` will fail on [L557-559](https://github.com/City-of-Helsinki/berth-reservations/blob/master/payments/schema/mutations.py#L557-L559) if the lease doesn't have an application (the case for imported leases)
```python
# Update application status
order.lease.application.status = ApplicationStatus.OFFER_SENT
order.lease.application.save()
```

I've already covered the case on [#365](https://github.com/City-of-Helsinki/berth-reservations/pull/365/files#diff-149de76f2f2a735d26e56250b9adb8fbb20b40d9991797d33091dca8e2eca4d0R290-R295), but just to be aware that it could happen for now.

a-rmz (changes requested):
We also need a new `NotificationType` for this additional product orders to be able to send confirmation invoice emails

a-rmz:
On [bambora_payform.py#L210](https://github.com/City-of-Helsinki/berth-reservations/blob/master/payments/providers/bambora_payform.py#L210) it again relies on having an application to get the customer data... I guess in the end we _do_ rely on having an application.

a-rmz:
_Other than that_, the flow worked all the way to the payment, the order has been marked as paid, and everything is good.

a-rmz (approved):
Good one! 🏆

PR #365, opened 2020-11-04
VEN-843 | Send invoices for existing berth leases

a-rmz:
## Description :sparkles:
* Add a semi-automatable `send invoice` task
* Add a mutation to execute that task
* Add a `ProfileService` to call the Profiili back-end and fetch all the customer profiles
* Add an `ERROR` state to leases and orders
* Rearrange some code that could be re-used in other components

### Closes :no_good_woman:
**[VEN-843](https://helsinkisolutionoffice.atlassian.net/browse/VEN-843):** Send invoice for existing leases

### Automated tests :gear:️
```shell
$ pytest customers/tests/test_profile_service.py
$ pytest leases/tests/test_send_invoice_task.py
$ pytest payments/tests/test_payments_utils.py
```

TBD

* Winter Storage leases are not implemented yet, we need to add the `renew_automatically` field and make some changes, but the base work is there already
* ~How to handle multiple attempts? I.e. avoid generating duplicate leases.~ Addressed already


santtul:
> Not sure if a plain GQL mutation is the best approach, since it could take a while to get the response back, but let's see about this
Yes, it might even hit to the timeout(s), there are many proxies along the way. Another way would be that the request creates an async task and returns a handle to that task. Then FE could poll/display with that handle. That would also probably mean that we need to store the results to the database.

customers/services/__init__.py
```diff
@@ -0,0 +1,26 @@
+import environ
+from django.conf import settings
+
+from .profile import ProfileService
+
+
+def load_services_config():
+    """Initialize the active payment provider config dict
```
santtul, line 8: Refers to payment provider

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+    """
+
+    def __init__(self, request, profile_list_threshold=None, **kwargs):
+        if "config" in kwargs:
+            self.config = kwargs.get("config")
+
+        self.profile_token = self._get_city_profile_token(request)
+        self.profile_list_threshold = (
```
santtul, line 45: `profile_list_threshold` seems to be unused?

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+        return {
+            PROFILE_API_URL: str,
+        }
+
+    def _get_city_profile_token(self, request: HttpRequest, nullable: bool = False):
+        api_tokens = request.META.get("HTTP_API_TOKENS")
+        if not api_tokens:
+            raise Exception("Missing tokens")
+
+        api_tokens = json.loads(api_tokens)
```
santtul, line 61: Rename to `parsed_api_tokens `? And in general it's good to avoid reassignment.

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        self.api_url = self.config.get(PROFILE_API_URL)
+
+    @staticmethod
+    def get_config_template():
+        return {
+            PROFILE_API_URL: str,
+        }
+
+    def _get_city_profile_token(self, request: HttpRequest, nullable: bool = False):
```
santtul, line 56: Could we remove the `nullable` switch?

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+        return profile_token
+
+    def get_all_profiles(self) -> Dict[str, CustomerProfile]:
+        query = self.ALL_PROFILES_QUERY
+
+        response = self.query(query)
+        edges = response.get("profiles", {}).get("edges", [])
+
+        # Parse the users received from the Profile Service
+        users = {}
```
santtul, line 75: Rename to `hki_profiles` ?

customers/tests/test_profile_service.py
```diff
@@ -0,0 +1,87 @@
+from unittest import mock
```
santtul, line 1: Running of this test is giving:
`django.core.exceptions.ImproperlyConfigured: Set the PROFILE_API_URL environment variable`

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+                        email
+                    }}
+                }}
+            }}
+        """
+
+        response = self.query(query)
+        user = self.parse_user(response.pop("profile"))
+        profile = CustomerProfile.objects.get(id=id)
+        profile.user = user
```
santtul, line 113: AFAIK `profile.user` is the "Django user" why we are overriding that with the object from Hki profile?

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # Associate the customer profiles with the fetched from the Profile Service
+        for profile in customer_profiles:
+            user = users[profile.id]
+            profile.user = user
+            users[profile.id] = profile
+
+        return users
+
+    def get_profile(self, id: UUID) -> CustomerProfile:
```
santtul, line 92: Could this function just return the profile data from Hki profile API (id, firs namet, last name, email) instead of our CustomerProfile?

leases/schema/mutations.py
```diff
@@ -411,6 +413,47 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+    class Input(SendExistingInvoicesInput):
+        pass
+
+    result = graphene.Field(SendExistingInvoicesType)
+
+    @classmethod
+    @view_permission_required(CustomerProfile)
+    @change_permission_required(BerthLease, WinterStorageLease, Order)
+    def mutate_and_get_payload(cls, root, info, **input):
+        from payments.schema import OrderNode
```
santtul, line 432: Can this be moved to start of the file with other imports?

santtul (changes requested):
Looks good so far! Great job with the `send_invoices.py` task, it was easy to follow because it was wisely splitted to meaningful functions 💪  Some comments there.

customers/services/profile.py
a-rmz: Good catch, leftover from my previous tests. 👍

leases/tasks/send_invoices.py
```diff
@@ -0,0 +1,231 @@
...
+def get_valid_winter_storage_leases(season_start: date) -> QuerySet:
+    raise NotImplementedError
+
+
+def get_valid_berth_leases(season_start: date) -> QuerySet:
+    """
+    Get the leases that were active last year
+    If today is:
+      (1) before season: leases from last year
+      (2) during or after season: leases from this year
```
ahie, line 102: It's possible to renew leases after the season has ended? It might be clearer for the admins if this action is only performed on currently active leases? I feel the already expired ones should be manually renewed.

customers/tests/test_profile_service.py
a-rmz: Have you added the variable to your `.env`?

leases/schema/mutations.py
a-rmz: It causes some circular import errors, that's why I added the import here.

.gitlab-ci.yml
```diff
@@ -52,6 +52,7 @@ staging:
     K8S_SECRET_VENE_PAYMENTS_BAMBORA_API_KEY: "$GL_QA_BAMBORA_API_KEY"
     K8S_SECRET_VENE_PAYMENTS_BAMBORA_API_SECRET: "$GL_QA_BAMBORA_API_SECRET"
     K8S_SECRET_VENE_PAYMENTS_BAMBORA_PAYMENT_METHODS: "$GL_QA_BAMBORA_PAYMENT_METHODS"
+    K8S_SECRET_PROFILE_API_URL: "https://profiili-api.test.kuva.hel.ninja/graphql/"
```
ahie, line 55: Should add this to the review env as well?

ahie (commented):
LGTM so far, a couple comments added.

Could we test how long this operation will approximately take in production? If it's a very heavy operation, this might affect how the UI should be built.

customers/services/profile.py
a-rmz: My reasoning for this was to have a common concise User model witht the fields that we need rather than just dicts. By setting the user to the `CustomerProfile`, we would have access to all the customer-related stuff when needed, like leases, orders, etc... and also the personal information through the `user` object.

customers/services/profile.py
a-rmz: (See the previous comment)

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+from ..models import CustomerProfile
+
+User = get_user_model()
+
+PROFILE_API_URL = "PROFILE_API_URL"
+
+
+class ProfileService:
+    profile_token: str
+    api_url = "https://profiili-api.prod.kuva.hel.ninja/graphql/"
```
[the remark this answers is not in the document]
a-rmz: Yup, just added this change in the upcoming batch of fixes.

leases/tasks/send_invoices.py
a-rmz: I'm not sure about the business requirements for this, I'll bring it up on Slack. 👍

leases/tasks/send_invoices.py
```diff
@@ -0,0 +1,231 @@
...
+            try:
+                approve_order(order, email, due_date, request)
+                successful_orders.append(order.id)
+            except (
+                AnymailError,
+                OSError,
+                ValidationError,
+                VenepaikkaGraphQLError,
+            ) as e:
+                fail_order(order, str(e), failed_orders)
```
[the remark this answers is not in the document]
a-rmz: For this first MVP version of the invoicing, we (dev/technical team) would manually check the logs to see which ones failed, we would only give them a count of how many succeeded and how many failed.

customers/services/profile.py
santtul: I think it's dangerous to override a field with something else that it's supposed to be. Could we create a new class called `HkiProfileUser` that holds only the fields form hki profiili (`id, first name, last name, email`). Then this service would expose only those kind of objects. What do you think? Would also like to hear opinion(s) from @ahie and @haavikko ?

customers/services/profile.py
```diff
@@ -0,0 +1,128 @@
...
+            PROFILE_API_URL: str,
+        }
+
+    def _get_city_profile_token(self, request: HttpRequest, nullable: bool = False):
+        api_tokens = request.META.get("HTTP_API_TOKENS")
+        if not api_tokens:
+            raise Exception("Missing tokens")
+
+        api_tokens = json.loads(api_tokens)
+        profile_token = api_tokens.get("https://api.hel.fi/auth/helsinkiprofile")
```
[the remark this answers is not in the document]
a-rmz: Sounds good

leases/tasks/send_invoices.py
a-rmz: Rsolved over slack:

> Lets go on with the case 2 that doesn’t limit the invoicing timetable and covers all the leases from previous season.
>
> Recurring invoicing will most probably always happen straight after the season before the year changes, but I don’t see any reason to limit the the invoicing from the system side.

customers/services/profile.py
```diff
@@ -0,0 +1,133 @@
...
+PROFILE_TOKEN_SERVICE = "PROFILE_TOKEN_SERVICE"
+
+
+@dataclass
+class HelsinkiProfileUser:
+    id: UUID
+    first_name: str = ""
+    last_name: str = ""
+    email: str = ""
+    customer: Optional[CustomerProfile] = None
```
santtul, line 23: Looks now great with `dataclass` 👍  One nit: having `customer` as field is not needed. By dropping it we could avoid the heavy customer table scan/iteration on line 89. what do you think?

a-rmz: I agree, it's not necessary but I was thinking on a generic use case where you would need access to the customer's leases, invoices, or something else customer-related. As it's not required for now, we can drop it and add it back if it's needed at some point. 👍

santtul (approved):
Great stuff!

leases/tasks/send_invoices.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def send_winter_storage_invoices(
+    request: HttpRequest, due_date: date = None
+) -> Dict[str, List]:
+    raise NotImplementedError
+
+
+@transaction.atomic
+def send_invoices(
```
haavikko, line 166: Error handling/risk management suggestion: would it make sense to have a "maximum_failures" parameter? If too many leases fail, then throw an exception, which results in the transaction being cancelled and no changes made. That way the databse will not accidentally end up with possibly 1000s of leases in LeaseStatus.ERROR, which probably would require manual cleanup? Maybe set the maximum_failures to something like 50 - that is probably an amount the admins can sort out manually? Or are there some classes of lease failures that are normal and expected?

leases/tasks/send_invoices.py
```diff
@@ -0,0 +1,231 @@
...
+
+    for lease in leases:
+        lease = create_new_lease(lease, season_start, season_end)
+
+        # Get the associated product
+        product = get_product(lease)
+
+        # If no product is found, the billing can't be done
+        if not product:
+            fail_lease(lease, _("No suitable product found"), failed_leases)
```
haavikko, line 193: If no product is found, then I understand the lease is left in LeaseStatus.DRAFTED state. What happens to these leases afterwards? Is the draft lease in this case useful for admins, or alternatively should the draft be deleted? If leaving the draft lease in db is needed, maybe set also the comment field in lease to indicate the error?

leases/tasks/send_invoices.py
```diff
@@ -0,0 +1,231 @@
...
+            continue
+
+        try:
+            order = Order.objects.create(
+                customer=lease.customer, lease=lease, product=product
+            )
+            orders.append(order)
+        except ValidationError as e:
+            fail_lease(lease, str(e), failed_leases)
```
haavikko, line 202: Same as above, what are the admins going to do with the draft lease if this fails? If a draft is left in the database and the send_invoices() is executed again, do I understand correctly that get_valid_berth_leases will exclude this lease from renewal?

leases/tasks/send_invoices.py
```diff
@@ -0,0 +1,231 @@
...
+        # If the profile doesn't have an associated email or if it has an example.com address, set it as failed
+        email = profiles.get(order.customer.id).email
+        if email and "example.com" not in email:
+            try:
+                approve_order(order, email, due_date, request)
+                successful_orders.append(order.id)
+            except (
+                AnymailError,
+                OSError,
+                ValidationError,
```
haavikko, line 220: Are ValidationError, VenepaikkaGraphQLError and OSError here "expected" as part of situation-as-usual, or are they more like an indication that something is horribly wrong and someone should take a look before the process can continue?
Maybe log the stacktrace to ease debugging?

leases/tasks/send_invoices.py
haavikko: Maybe log the stacktrace to ease debugging?

leases/tasks/send_invoices.py
a-rmz: > then I understand the lease is left in LeaseStatus.DRAFTED state

The lease is set to `ERROR` status.

> maybe set also the comment field in lease to indicate the error

Yes! Great minds think alike, that's part of the improvements on the next PR

leases/tasks/send_invoices.py
a-rmz: If `send_invoices` is run again, it ignores _all_ the renewed leases for the upcoming season. Meaning that if the lease is in `ERROR`, it won't try to generate it again.

> do I understand correctly that get_valid_berth_leases will exclude this lease from renewal? Or is this the desired result?

Yes to both. On the `Kausilaskut` section on the Admin UI they will be able to check which leases are in error state and handle them. As you pointed out on the previous one, we'll store the error message on the `comment` field so they/we know what was the issue.

leases/tasks/send_invoices.py
a-rmz: `VenepaikkaGraphQLError` would be a more "situation-as-usual" error handling (something went wrong with the instance.

`ValidationError` would be a bit more concerning, since it's raised by the `get_order_notification_type` meaning that an order doesn't have a valid type, but shouldn't be the case very often.

IIRC, `OSError` can be raised by `anymail`/`django-mailer`/`django-ilmoitin`, so probably not critical (systems down-critical), but also worth checking.

In any case, definitely worth logging this more "serious" errors. 👍

leases/tasks/send_invoices.py
a-rmz: Good point 🤔 worth exploring the idea. I'll see if I can reach a simple version that works for this MVP.

PR #366, opened 2020-11-05
Cleanup after migration

tebbon:
## Description :sparkles:
Cleanup leftovers from the migration work:
* Reduce uwsgi log spam by ignoring errors about clients dropping connections
* Remove postgres-cmd-line tools from prod image
* Remove sticker assignment management command
* Disable db initialization fixture as it does not work properly, and staging and prod already have data in the db.


santtul (approved):
👍

PR #371, opened 2020-11-12
Improve vene-specific GraphQL exceptions

ahie:
Adds a `type` field to the extensions that graphql emits in the error response. This way we can differentiate between actually "unexpected" errors and errors that we caught but specifically forward to the frontend.


santtul (approved):
LGTM!

PR #372, opened 2020-11-12
Add translation compilation to Dockerfile

ahie:
Added translation compilation step to docker script. This step is needed for translations to be used by django.

Additionally removes location details from translation files, to remove unnecessary changes from git diffs in the future.


docker-entrypoint.sh
```diff
@@ -29,6 +29,11 @@ if [[ "$CREATE_SUPERUSER" = "1" ]]; then
   echo "Admin user created with credentials admin:adminpass (email: admin@example.com)"
 fi
 
+if [[ "$COMPILE_MESSAGES" = "1" ]]; then
```
santtul, line 32: I think entrypoint (starting of the service) is not a good place for compilation. Could this be done when the docker image is built?

ahie: I opted to add it here to keep the build script cleaner, and because other manage scripts are also run here. The operation is actually very quick so it shouldn't noticeably increase startup time.

santtul: Ok, could you rebase this? @a-rmz can you also take a look please?

santtul: @ahie In general it's a good practice to do any compilation and preparation work on docker build phase, so that we catch the problems early as possible. But in our case it doesn't matter that much since building is done within deployment. It would be great if the docker image would be build just once and used in all envs, there it would matter more.

santtul (dismissed):
Approved, but would like to @a-rmz to check it also.

docker-entrypoint.sh
a-rmz: I agree that it would make more sense to have it on the build stage. The commands on the entrypoint are there because they require a database connection, but since this is a code-only change, we could keep it on build.

ahie: Moved to build phase and rebased.

a-rmz (approved):

santtul:
@ahie seems that the msg compilation is failing: https://gitlab.com/City-of-Helsinki/KuVa/github-mirrors/berth-reservations/-/jobs/853613115

santtul (changes requested):
seems that the msg compilation is failing: https://gitlab.com/City-of-Helsinki/KuVa/github-mirrors/berth-reservations/-/jobs/853613115

ahie:
> seems that the msg compilation is failing: https://gitlab.com/City-of-Helsinki/KuVa/github-mirrors/berth-reservations/-/jobs/853613115

At least locally docker-compose builds correctly with the recent changes. Lets see what the pipeline thinks.

santtul (approved):
👍

PR #380, opened 2020-11-25
VEN-1016 | Additional product order adjustments

santtul:
## Description :sparkles:
Approve order changes for AP order
- Do not update lease and applications status
- Send notification

Other
- Resolve lease order type as INVALID for AP order
- Extend hki profile with address info
- Minor fixes

## Issues :bug:
**[VEN-1016](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1016)**

### Automated tests :gear:️
Several new tests in `payments/tests`

### DIFF
--- a/customers/services/profile.py
+++ b/customers/services/profile.py
@@ -145,10 +145,10 @@ class ProfileService:
         user_id = from_global_id(profile.pop("id"))
         email = (profile.pop("primary_email") or {}).get("email")
 
-        primary_address = profile.pop("primary_address", None)
-        address = (primary_address or {}).get("address")
-        postal_code = (primary_address or {}).get("postal_code")
-        city = (primary_address or {}).get("city")
+        primary_address = profile.pop("primary_address", {})
+        address = primary_address.get("address")
+        postal_code = primary_address.get("postal_code")
+        city = primary_address.get("city")
 
         return HelsinkiProfileUser(
             UUID(user_id),
--- a/customers/services/profile.py
+++ b/customers/services/profile.py
@@ -18,6 +18,9 @@ class HelsinkiProfileUser:
     first_name: str = ""
     last_name: str = ""
     email: str = ""
+    address: str = ""
+    postal_code: str = ""
+    city: str = ""
 
 
 class ProfileService:
@@ -39,6 +42,11 @@ class ProfileService:
                         last_name: lastName
                         primary_email: primaryEmail {
                             email
+                        }
+                        primary_address: primaryAddress {
+                            address
+                            postal_code: postalCode
+                            city
                         }
                     }
                 }
@@ -117,6 +125,11 @@ class ProfileService:
                     primary_email: primaryEmail {{
                         email
                     }}
+                    primary_address: primaryAddress {{
+                        address
+                        postal_code: postalCode
+                        city
+                    }}
                 }}
             }}
         """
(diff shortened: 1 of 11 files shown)


payments/notifications/dummy_context.py
```diff
@@ -101,6 +101,26 @@ def _get_winter_storage_order_context():
...
 
 
+def _get_additional_product_order_context():
+    customer = CustomerProfileFactory.build()
+    order = OrderFactory.build(
+        customer=customer,
+        lease=BerthLeaseFactory.build(customer=customer),
+        product=None,
+        price=Decimal("0.00"),
+        tax_percentage=Decimal("0.00"),
```
a-rmz, line 111: One thing to consider if we're gonna be setting the tax percentage to zero, is to update this to avoid division errors. Or, just stick to the predefined `PLACE_PRODUCT_TAX_PERCENTAGES`.

```python
    @property
    def total_tax_percentage(self):
        return (
            rounded_decimal(
                ((self.total_price - self.total_pretax_price) / self.total_pretax_price)
                * 100,
                round_to_nearest=0.05,
            )
            # Add this check
            if self.total_pretax_price != 0
            else 0
        )
```

customers/services/profile.py
```diff
@@ -131,7 +144,20 @@ def parse_user_edge(self, gql_edge: Dict[str, dict]) -> HelsinkiProfileUser:
     def parse_user(self, profile: Dict[str, dict]) -> HelsinkiProfileUser:
         user_id = from_global_id(profile.pop("id"))
         email = (profile.pop("primary_email") or {}).get("email")
-        return HelsinkiProfileUser(UUID(user_id), email=email, **profile)
+
+        primary_address = profile.pop("primary_address")
+        address = (primary_address or {}).get("address")
+        postal_code = (primary_address or {}).get("postal_code")
+        city = (primary_address or {}).get("city")
```
a-rmz, line 151: Would it be a bit simpler to do
```suggestion
        primary_address = profile.pop("primary_address") or {}
        address = primary_address.get("address", "")
        postal_code = primary_address.get("postal_code", "")
        city = primary_address.get("city", "")
```

payments/schema/mutations.py
```diff
@@ -434,7 +434,7 @@ class Input:
         id = graphene.ID(required=True)
 
     @classmethod
-    @delete_permission_required(BerthProduct)
+    @delete_permission_required(Order)
```
a-rmz, line 437: Good catch 😅

payments/tests/test_payments_mutations_approve_ap_order.py
```diff
@@ -0,0 +1,63 @@
...
+    executed = api_client.execute(APPROVE_ORDER_MUTATION, input=variables)
+    payment_url = payment_provider.get_payment_email_url(order, lang="en")
+
+    order = Order.objects.get(id=order.id)
+
+    assert order.due_date == due_date
+
+    # approving additional product order should not touch the lease
+    assert order.lease.status != LeaseStatus.OFFERED
+    assert order.lease.application.status != ApplicationStatus.OFFER_SENT
```
a-rmz, line 50: Should we check that the new order is in the right status?
```suggestion
    assert order.lease.status != LeaseStatus.OFFERED
    assert order.lease.application.status != ApplicationStatus.OFFER_SENT
    assert order.status == OrderStatus.DRAFTED
```

santtul: Approve order doesn't touch the order.status so no reason to test.

customers/services/profile.py
santtul: `profile.pop("primary_address")` without default value throws error. cleaned this thing up a bit.

a-rmz (approved):

PR #382, opened 2020-11-27
VEN-843 | Berth Invoicing improvements

a-rmz:
## Description :sparkles:
* Replace the `send_invoices` task with a service, easier to manage, easier to extend
* Add a `statuses` and `start_year` filters to the `berthLeases` query to allow showing:
  * `ERROR` leases on the `Kausilaskut` page
  * Successful leases (`status=DRAFTED and start_date__year=next season year`)
* Add a `statuses` filter to the `orders` query to allow showing `ERROR` orders on the `Kausilaskut` page
* Add the error message to the comments for failed leases and orders (more detailed too)
* Add a limit for max failures: if more than 50 leases/orders fail on a single batch, fail the whole thing so we can take a look

### Related :handshake:
[VEN-843](https://helsinkisolutionoffice.atlassian.net/browse/VEN-843)
#365 

### Automated tests :gear:️
```shell
$ pytest
```


santtul (approved):
Great work! 👍

leases/services/invoice.py
```diff
@@ -0,0 +1,242 @@
...
+                            customer=customer, lease=new_lease, product=product
+                        )
+
+                        self.send_email(
+                            order, profiles.get(customer.id).email,
+                        )
+                    except ValidationError as e:
+                        self.fail_lease(new_lease, str(e))
+            except DatabaseError as e:
+                lease.refresh_from_db()
```
haavikko, line 167: What is the expected case where DatabaseError would be thrown? Would it make sense to catch a more specific type of DatabaseException, like IntegrityError? DatabaseError could be something that is not recoverable.

leases/services/invoice.py
```diff
@@ -0,0 +1,242 @@
...
+            try:
+                approve_order(order, email, self.due_date, self.request)
+                self.successful_orders.append(order.id)
+            except (
+                AnymailError,
+                OSError,
+                ValidationError,
+                VenepaikkaGraphQLError,
+            ) as e:
+                self.fail_order(order, str(e))
```
haavikko, line 116: Would it help debugging if the full stacktrace is logged?

leases/services/invoice.py
```diff
@@ -0,0 +1,242 @@
...
+        if current_date < season_start and current_date.year == season_start.year:
+            lease_year = current_date.year - 1
+        else:  # (2)
+            lease_year = current_date.year
+
+        # Filter leases from the upcoming season
+        future_leases = BerthLease.objects.filter(
+            start_date__year__gt=lease_year,
+            berth=OuterRef("berth"),
+            customer=OuterRef("customer"),
```
haavikko, line 228: I haven't used this technique before, good to learn new things!

leases/tests/test_send_berth_invoice_service.py
```diff
@@ -663,3 +671,52 @@ def test_send_berth_invoices_only_not_renewed(notification_template_orders_appro
...
+        "first_name": user.first_name,
+        "last_name": user.last_name,
+        "primary_email": {"email": "something@example.com"},
+    }
+
+    assert Order.objects.count() == 0
+
+    with mock.patch(
+        "customers.services.profile.requests.post",
+        side_effect=mocked_response_profile(count=1, data=data),
```
haavikko, line 716: Another new technique I learnt from this code :-)

haavikko (approved):
Great work!

leases/services/invoice.py
a-rmz: Not really sure. I was just trying to catch all the possible errors, but you make a good point that it could be not recoverable. Let me take a look at that.

haavikko: Well if it is a serious database error (connection lost or whatever), then it would most likely throw another exception in refresh_from_db, so I think the code is ok as-is, maybe log error stacktrace?

PR #385, opened 2020-12-02
VEN-964 | New berth pricing

a-rmz:
## Description :sparkles:
The full description of the changes and the new schema can be found on the [Confluence page](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/DD/pages/951124156/New+payment+schema+2021-2022)

* Remove the `BerthPriceGroup` model
* Add the tier prices and width ranges to the `BerthProduct` model
* Remove the `price` field from the `BerthProduct`
  * Add a `price_for_tier(tier: PriceTier)` method to the product
* Update the `Order` creation so that it takes the price according to the width range and `pier.price_tier`
* Add a `PriceTier` enum
* Add a `price_tier = PositiveSmallIntegerField(choices=PriceTier.choices)` field on Pier to assign it a price tier
* Remove the `price_group` from the `BerthType` model
* Add product fixtures that can be loaded easily
* Add a `berthProductForWidth` query
* Update all the mutations and methods that would be affected by this changes

### Closes :no_good_woman:
**[VEN-964](https://helsinkisolutionoffice.atlassian.net/browse/VEN-964):** Update the pricing models

### Automated tests :gear:️
```shell
$ pytest
```

You can use the review environment:
https://berth-reservations-qa-r-ven-964-new-vkz6x4.test.kuva.hel.ninja/graphql_v2/
```
admin:adminpass
```

**berthProducts:**
```graphql
query BerthProducts {
  berthProducts {
    edges {
      node {
        minWidth
        maxWidth
        tier1Price
        tier2Price
        tier3Price
      }
    }
  }
}
```
```json
{
  "data": {
    "berthProducts": {
      "edges": [
        {
          "node": {
            "minWidth": "0.00",
            "maxWidth": "2.00",
            "tier1Price": "200.00",
            "tier2Price": "170.00",
            "tier3Price": "150.00"
          }
        },
       // ..
      ]
    }
  }
}
```

**berthProductForWidth:**
```graphql
query BerthProducts {
  berthProductForWidth(width: "3.33") {
    minWidth
    maxWidth
    tier1Price
    tier2Price
    tier3Price
  }
}
```
```json
{
  "data": {
    "berthProductForWidth": {
      "minWidth": "3.00",
      "maxWidth": "4.00",
      "tier1Price": "440.00",
      "tier2Price": "410.00",
      "tier3Price": "380.00"
    }
  }
}
```

Load the fixtures to have the real products loaded on the environment

```shell
$ python manage.py loaddata berth_products.json
```


payments/models.py
```diff
@@ -512,8 +533,12 @@ def save(self, *args, **kwargs):
         # If the product instance is being passed
         # Price has to be assigned before saving but only if creating
         if creating and self.product:
-            # Assign the price and tax from the product
-            price = self.product.price_value
+            # Assign the tax from the product
```
santtul, line 536: This comment points now to wrong line, can you move it lower

payments/models.py
```diff
@@ -512,8 +533,12 @@ def save(self, *args, **kwargs):
...
+            price = (
+                self.product.price_value
+                if self.product and hasattr(self.product, "price_value")
+                else 0
```
santtul, line 540: Would it be safer to fallback to `None` in this case? Maybe the configuration error would be more visible then and something would hopefully blow up in payment process? 😄

santtul:
Looks very good so far! 👍

payments/models.py
a-rmz: Then we would need to assign it to `price_value` somewhere else, for winter storage orders this is the place where it's set.

santtul (approved):
Great job! 👍

.prod/on_deploy.sh
```diff
@@ -3,3 +3,4 @@
 python /app/manage.py migrate --noinput
 
 # Admin commands that need to be ran for each env can be added here
+python /app/manage.py loaddata berth_products.json
```
ahie, line 6: Is this needed in both `on_deploy` and `on_initialize`? The other `loaddata` files are not loaded here.

templates/email/messages/berth_invoice_en.html
```diff
@@ -17,7 +17,7 @@ <h3>Price of the boat berth</h3>
 <table style="width: 100%">
     <tr>
         <td colspan="2">
-            Basic price ({{ order.product.price_group.name }}):
+            Basic price ({{ order.product.display_dimensions }}):
```
ahie, line 20: The text here seems a bit weird, but I guess it's intentional.

payments/models.py
```diff
@@ -105,46 +81,95 @@ class Meta:
...
-    price_group = models.ForeignKey(
-        BerthPriceGroup,
-        on_delete=models.CASCADE,
-        verbose_name=_("price group"),
-        related_name="products",
+class BerthProductManager(models.Manager):
+    def get_in_range(self, width: Union[Decimal, float, int]):
+        return (
+            self.get_queryset()
+            .filter(min_width__lt=width, max_width__gte=width)
```
ahie, line 88:
```suggestion
            .filter(min_width__lte=width, max_width__gt=width)
```

Should the equality check be moved to the minimum width instead?

payments/models.py
```diff
@@ -105,46 +81,95 @@ class Meta:
...
 
-class WinterStorageProduct(AbstractPlaceProduct):
+class WinterStorageProduct(AbstractPlaceProduct, AbstractBaseProduct):
     winter_storage_area = models.OneToOneField(
         "resources.WinterStorageArea",
         verbose_name=_("winter storage area"),
         related_name="product",
         on_delete=models.CASCADE,
     )
+    price_unit = models.CharField(
```
ahie, line 168: Should `price_unit` to one of the abstact base classes? This field is duplicated in `WinterStorageProduct` and `BerthProduct`.

ahie (commented):
LGTM! A couple small comments added.

Also, just to make sure. When these changes go in will they have an effect on the additional services pricing on already billed customers which are % based?

payments/models.py
ahie: Judging by the table I think the way it's implemented is the intention. The implementation doesn't cover the case where `width == 0` though

a-rmz:
@ahie
> Also, just to make sure. When these changes go in will they have an effect on the additional services pricing on already billed customers which are % based?

Not retroactively. The orders with additional services that have already been generated have a fixed price on the `Order`, so even if the price from the product changes, the one on the order will still be the same. This will only affect additional orders once the changes roll out.

.prod/on_deploy.sh
a-rmz: Yes. The `on_initalize` is only ran on the `review` environments, since both `staging` and `prod` have already been initalized. However, for fresh review environments it's useful to have some data loaded.

Once this is merged and the command runs on staging and prod, we can remove it from here.

templates/email/messages/berth_invoice_en.html
a-rmz: You mean the `Basic price` itself?

a-rmz: Right, I guess we could go with `Base price`, would be a bit clearer.

payments/models.py
a-rmz: It was on purpose to have the `min < width <= max` 

**Original:**
| Min | Max |
| --- | --- |
| 0.0 | 2.0 |
| 2.1 | 2.75 |
| 2.76 | 3.0 |

**_Our_ version:**
| Min | Max |
| --- | --- |
| 0.0 | 2.0 |
| 2.0 | 2.75 |
| 2.75 | 3.0 |

So, to be able to keep the same values on the same ranges, we need to do it like that.
```
--|(---------]|(-------]|(------]|
  0           2        2.75      3
```

templates/email/messages/berth_invoice_en.html
ahie: The `order.product.display_dimensions` gives the range in meters and the customer isn't really aware of the pricing scheme being used. That's why I think it might be a bit confusing.

payments/models.py
a-rmz: I tried adding it to the `AbstractPlaceProduct` class but it has some conflict, I guess because also `AbstractBaseProduct` implements it and it doesn't know which one has precedence.

templates/email/messages/berth_invoice_en.html
a-rmz: Gotcha. We are going to inform the customers about the pricing changes, so in my mind it would make sense to display the price so they can corroborate that it's the right one.

(The full answer:)
> Yes, the idea is that customers (in the mailing list) know about the new pricing and about the need to check the invoice before they pay it before the actual invoices are sent. And luckily that's what people usually do, check the amount before they pay

payments/models.py
```diff
@@ -105,46 +81,95 @@ class Meta:
...
-        BerthPriceGroup,
-        on_delete=models.CASCADE,
-        verbose_name=_("price group"),
-        related_name="products",
+class BerthProductManager(models.Manager):
+    def get_in_range(self, width: Union[Decimal, float, int]):
+        return (
+            self.get_queryset()
+            .filter(min_width__lt=width, max_width__gte=width)
+            .first()
```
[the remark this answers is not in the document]
a-rmz: Hmm good point. It _shouldn't_ happen, but if that were the case, it would take the best guess. But if that happens, then we have bigger problems than selecting the correct product. 😅

payments/tests/test_payments_queries.py
```diff
@@ -284,6 +136,46 @@ def test_get_berth_product_not_enough_permissions(api_client, berth_product):
...
+"""
+
+
+@pytest.mark.parametrize(
+    "api_client",
+    ["berth_supervisor", "berth_handler", "berth_services"],
+    indirect=True,
+)
+def test_get_berth_product_for_width(api_client):
+    berth_product = BerthProductFactory(min_width="3.00", max_width="4.00")
```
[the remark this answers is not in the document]
a-rmz: Great idea!

payments/migrations/0018_add_price_tiers.py
```diff
@@ -0,0 +1,110 @@
...
+        migrations.AlterModelOptions(
+            name="berthproduct", options={"ordering": ["min_width"]},
+        ),
+        migrations.RemoveField(model_name="berthproduct", name="harbor",),
+        migrations.RemoveField(model_name="berthproduct", name="price_group",),
+        migrations.RemoveField(model_name="berthproduct", name="price_value",),
+        migrations.AddConstraint(
+            model_name="berthproduct",
+            constraint=models.UniqueConstraint(
+                fields=("min_width", "max_width"), name="unique_width_range"
```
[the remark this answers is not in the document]
a-rmz: As documented on the [new pricing schema](https://helsinkisolutionoffice.atlassian.net/wiki/spaces/DD/pages/951124156/New+payment+schema+2021-2022), there would never be the case when products overlap. As I already said, if this happened, we'd have bigger problems.

However, we could add some validation if it's not too complicated. 👍

leases/services/invoice.py
```diff
@@ -212,21 +212,7 @@ def get_product(lease: BerthLease) -> BerthProduct:
         # width based on the berth_type.
         width = lease.boat.width if lease.boat else lease.berth.berth_type.width
```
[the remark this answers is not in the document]
ahie: This is a good point. AFAIK the price should be tied to the berth, i.e. the price tier of the pier it is on and the width of the berth. But as the admins will be manually handling these it’s really easy (with how the frontend is currently implemented) to select a berth that’s needlessly wide for the customer.

This is a separate issue, but should we filter the frontend to only show berths for the range that makes most sense for a customer?

payments/models.py
```diff
@@ -105,46 +81,95 @@ class Meta:
...
 
 
-class BerthProduct(AbstractPlaceProduct):
-    price_group = models.ForeignKey(
-        BerthPriceGroup,
-        on_delete=models.CASCADE,
-        verbose_name=_("price group"),
-        related_name="products",
+class BerthProductManager(models.Manager):
+    def get_in_range(self, width: Union[Decimal, float, int]):
```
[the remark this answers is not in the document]
a-rmz: Hmm probably not raise an error but good idea for the logging 👍

leases/services/invoice.py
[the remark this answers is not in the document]
haavikko: If price is tied to the berth, then lease.boat.width should not be used?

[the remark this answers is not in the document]
haavikko: I just asked clarification from Joonas on Slack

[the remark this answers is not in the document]
a-rmz: If the berth is 3m, then the customer will be billed for the 3m berth. For berth products, the billing is pre-defined, so it doesn't matter how big is the boat, as long as the berth is of this or that dimension.

For winter storage products, those _are_ billed by m2, but that's a different case.

payments/migrations/0018_add_price_tiers.py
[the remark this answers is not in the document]
a-rmz: Will be addressed on [VEN-1046](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1046)

PR #386, opened 2020-12-03
VEN-1016 | Attach customer contact data to order

santtul:
## Description :sparkles:
We cannot fetch the hki profile customer data in customer ui because there are no admin tokes, so we will attach that data to order on approve order phase in admin so that we can use that data later in the customer ui confirm payment step.

Some other fixes bundled to this PR as well (see commit messages).

### Related :handshake:
**[VEN-1016](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1016)** 

### Automated tests :gear:️
Extended approve order automatic tests.


payments/providers/bambora_payform.py
```diff
@@ -204,24 +203,35 @@ def payload_add_products(self, payload: dict, order: Order, language: str):
...
+                    "customer": {
+                        "firstname": application.first_name.capitalize(),
+                        "lastname": application.last_name.capitalize(),
+                        "email": application.email.strip(),
+                        "address_street": application.address,
+                        "address_zip": application.zip_code,
+                        "address_city": application.municipality.capitalize(),
+                    },
+                }
+            )
```
a-rmz, line 220: Should we update the fields on the `order` so that the information matches?

santtul: Agreed that keeping it like it is.

a-rmz (approved):

PR #388, opened 2020-12-03
VEN-843 | Profile token

a-rmz:
## Description :sparkles:
* Since Federation strips the profile token before it reaches our backend, we need to receive it as an input for the mutations that require it.
  * The same way [Youth Membership](https://github.com/City-of-Helsinki/youth-membership/blob/develop/youths/schema/mutations.py#L187-L189) does it

### Related :handshake:
**[VEN-843](https://helsinkisolutionoffice.atlassian.net/browse/VEN-843)**: Send invoice for existing berth leases 

### Automated tests :gear:️
```shell
$ pytest customers/tests/test_profile_service.py
$ pytest leases/tests/test_lease_mutations_send_invoices.py
$ pytest leases/tests/test_send_berth_invoice_service.py
```

PR #390, opened 2020-12-04
VEN-843 | Fail the lease if it's not connected to a customer

a-rmz:
## Description :sparkles:
* Handle the case when a lease's customer does not have a profile on the Profile backend

### Related :handshake:
#365 
#382 
#388


santtul (approved):
👍

PR #396, opened 2020-12-09
VEN-843 | Set the correct lease status on creation

a-rmz:
## Description :sparkles:
When creating the lease, we are just copying the previous one which will
be on paid status. For proper bookkeeping, we should set it as drafted
to follow the correct flow from drafted -> offered/error -> paid/rejected.

The lease would in any case be marked as offered/error by the script,
but better to follow the right steps.

After running the script, check the `Lease changes` on the new lease. You should see the transition from `drafted -> offered`

## Screenshots :camera_flash:
**For example:**
image

**From local testing:**
image

### DIFF
--- a/leases/services/invoice.py
+++ b/leases/services/invoice.py
@@ -73,6 +73,9 @@ class BaseInvoicingService:
         lease.start_date = start_date
         lease.end_date = end_date
 
+        # Set correct status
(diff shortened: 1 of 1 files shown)


santtul (approved):
LGTM 👍

PR #398, opened 2020-12-10
Make VismaSign fulfill_contract use test SSN if present in env

ahie:
## Description :sparkles:

This is to workaround the issue where VismaSign test server
requires a certain SSN for the contract fulfillment call,
whereas in prod this is not required and we do not want to
force the user to submit an ssn if unnecessary.


a-rmz:
Would it be possible that, instead of adding the workaround, we just pass a dummy SSN on staging?

ahie:
> Would it be possible that, instead of adding the workaround, we just pass a dummy SSN on staging?

Unfortunately not. We need to pass the ssn generated for the test account here. Tried with a dummy ssn and it hangs in the test servers auth page. With no ssn provided the server just crashes with status `500`.

ahie:
As our "prod" of VismaSign has been switched to testing mode as well, lets not merge this yet to verify other authentication methods also work without providing an ssn in this step. I've only been able to test with nordea.

a-rmz:
Do we still need this?

ahie:
Yes

a-rmz (approved):
Let's take it for a ride!

PR #399, opened 2020-12-10
Use Github Actions for CI/CD instead of Gitlab

ruriky:
### Closes :no_good_woman:

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -1,0 +1,63 @@
+name: Build & Review
+on: [pull_request]
+
+env:
+  CONTAINER_REGISTRY: ghcr.io
+  CONTAINER_REGISTRY_USER: ${{ secrets.GHCR_CONTAINER_REGISTRY_USER }}
+  CONTAINER_REGISTRY_PASSWORD: ${{ secrets.GHCR_TOKEN }}
+  CONTAINER_REGISTRY_REPO: ghcr.io/city-of-helsinki/${{ github.event.repository.name }}
+  REPO_NAME: ${{ github.event.repository.name }}
+  KUBECONFIG_RAW: ${{ secrets.KUBECONFIG_RAW }}
+  BUILD_ARTIFACT_FOLDER: "build_artifacts"
+  SERVICE_ARTIFACT_FOLDER: "service_artifacts"
+  BASE_DOMAIN: ${{ secrets.BASE_DOMAIN_STAGING }}
+  DATABASE_USER: user
+  DATABASE_PASSWORD: testing-password
+  APP_MIGRATE_COMMAND: /app/.prod/on_deploy.sh
+  APP_INITIALIZE_COMMAND: /app/.prod/on_initialize.sh
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        uses: andersinno/kolga-build-action@v2
+
+  review:
+    runs-on: ubuntu-latest
+    needs: build
+    name: Review
+    steps:
+      - uses: actions/checkout@v2
+      - uses: andersinno/kolga-setup-action@v2
+
+      - name: Review-Services
+        uses: andersinno/kolga-deploy-service-action@v2
+        with:
+          projects: BERTH_RESERVATIONS
+        env:
+          POSTGRES_IMAGE: "docker.io/andersinnovations/postgis:11-bitnami"
+
+      - name: Deploy
+        uses: andersinno/kolga-deploy-action@v2
+        env:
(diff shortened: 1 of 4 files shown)


.github/workflows/review.yml
```diff
@@ -0,0 +1,60 @@
...
+  CONTAINER_REGISTRY: ghcr.io
+  CONTAINER_REGISTRY_USER: ${{ secrets.GHCR_CONTAINER_REGISTRY_USER }}
+  CONTAINER_REGISTRY_PASSWORD: ${{ secrets.GHCR_TOKEN }}
+  CONTAINER_REGISTRY_REPO: ghcr.io/city-of-helsinki/${{ github.event.repository.name }}
+  REPO_NAME: ${{ github.event.repository.name }}
+  KUBECONFIG_RAW: ${{ secrets.KUBECONFIG_RAW }}
+  BUILD_ARTIFACT_FOLDER: "build_artifacts"
+  SERVICE_ARTIFACT_FOLDER: "service_artifacts"
+  BASE_DOMAIN: ${{ secrets.BASE_DOMAIN_STAGING }}
+  APP_MIGRATE_COMMAND: /app/.prod/on_deploy.sh
```
a-rmz, line 14: Can you also add the initialize script? (only for review)
```python
APP_INITIALIZE_COMMAND: /app/.prod/on_initialize.sh
```

a-rmz (commented):
Looks good so far 👍

.github/workflows/review.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: andersinno/kolga-setup-action@v2
+
+      - name: Review-Services
+        uses: andersinno/kolga-deploy-service-action@v2
+        with:
+          projects: BERTH_RESERVATIONS
+        env:
+          POSTGRES_IMAGE: "docker.io/andersinnovations/postgis:11-bitnami"
```
a-rmz, line 39: What's the default user it generates? We should probably also add the user/pass to the env variables on the `Deploy` action so it can connect to the db.

joonvena: You can use add DATABASE_USER & DATABASE_PASSWORD env variables if you want to use your own custom credentials. Otherwise it will auto-generate these and inject them to review deployment.

a-rmz: Hmmm... there might be some misconfiguration with the variable names then. The review pipeline failed since it didn't get the right user/password.

https://github.com/City-of-Helsinki/berth-reservations/pull/399/checks?check_run_id=1536161086
```log
django.db.utils.OperationalError: FATAL:  password authentication failed for user "user"
```

joonvena: Oh, yeah actually we found bug there. When you first deploy the review env it works fine with the autogenerated user/password but as you make another push to your PR it triggers the pipeline again and redeploys the application but the updated credentials doesn't get reflected to postgresql pod and it still uses the old credentials. So the preferred way is to use those variables until we get this fixed on Kolga side.

.github/workflows/stop_review.yml
```diff
@@ -0,0 +1,19 @@
...
+  stop_review:
+    runs-on: ubuntu-latest
+    name: Stop Review
+    steps:
+      - uses: andersinno/kolga-setup-action@v2
+
+      - name: Stop Review
+        uses: andersinno/kolga-review-cleanup-action@v2
+        with:
+          namespace: ${{ env.K8S_NAMESPACE }}
```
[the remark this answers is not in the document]
joonvena: Added

.github/workflows/review.yml
a-rmz: Sneaky bug! Haha but good find. So do we add manually the variables/credentials?

a-rmz:
Another thing: the review now passes, but it doesn't log the full review url 😅 
```
 ☸️  📄 Deployment can be accessed via https://berth-reservations-536-dev-github-actions.***
```

.github/workflows/review.yml
joonvena: Yes manually. I added them to the review workflow. They can be adjusted to your liking.

joonvena:
Yeah this is known "issue". As now we are configuring the repository settings through Terraform using Github provider. We are injecting the BASE_DOMAIN env as a secret. This is because we are using subdomain for the updated cluster and when we flip the switch and use the updated cluster as a main we can update the address easily through Terraform to all repositories. Secrets are masked from logs and that's why the output is ***. I really would like to see feature added where you can define normal environment variables to repository settings or have a toggle for the secret that allows it to be visible if needed.

a-rmz:
Ahh alright! I guess as long as we know what's the full domain once this is rolled out, that would be enough. 👍

joonvena (approved):
LGTM

PR #400, opened 2020-12-11
VEN-964 | Remove automatic creation of order lines

a-rmz:
## Description :sparkles:
On the previous pricing schema, the fixed additional services
(electricity, gate, lighting, mooring, waste collection, and water) were
billed separately but the order lines were generated automatically when
creating the order, based on the services available on the pier/harbor.

The new pricing schema already includes those services on the berth
prices, so we don't need to bill them separately.

* Remove the `create_order_lines` from `Order` and the related tests.

### Related :handshake:
**[VEN-964](https://helsinkisolutionoffice.atlassian.net/browse/VEN-964):** Update the pricing models
**#385** 

### Automated tests :gear:️
```shell
```

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -500,21 +500,6 @@ class Order(UUIDModel, TimeStampedModel):
             # If the lease is being changed
             self._check_same_lease(old_instance)
 
-    def _create_order_lines(self, section):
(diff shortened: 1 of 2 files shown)


haavikko (approved):
OK great!

PR #402, opened 2020-12-11
VEN-1043 | Update payment application status

a-rmz:
## Description :sparkles:
When a payment is made for a lease which doesn't have an application associated (as is the case with the Timmi leases), it will raise a 500-Server Error because it fails to set the application to handled status.

### Closes :no_good_woman:
**[VEN-1043](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1043):** Updating applications on payment success

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_models.py::test_order_set_status_no_application
```

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -668,17 +668,18 @@ class Order(UUIDModel, TimeStampedModel):
         self.lease.save(update_fields=["status", "comment"])
 
         if new_status == OrderStatus.PAID:
-            application = self.lease.application
-            application.status = ApplicationStatus.HANDLED
-            self.lease.application.save(update_fields=["status"])
+            if self.lease.application:
+                application = self.lease.application
+                application.status = ApplicationStatus.HANDLED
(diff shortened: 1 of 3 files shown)


payments/models.py
```diff
@@ -668,17 +668,18 @@ def update_lease_and_application(self, new_status):
...
-            self.lease.application.save(update_fields=["status"])
-
-            if (
-                isinstance(self.lease, WinterStorageLease)
-                and application.area_type == ApplicationAreaType.UNMARKED
-            ):
-                sticker_number = get_next_sticker_number(self.lease.start_date)
-                self.lease.sticker_number = sticker_number
-                self.lease.save(update_fields=["sticker_number"])
+            if self.lease.application:
```
haavikko, line 671: If you like, add a comment that missing application might happen due to TIMMI imported data (and any other situations where an order might exist without application?)

haavikko (approved):
great stuff!

PR #404, opened 2020-12-14
VEN-843 | Invoicing threaded

a-rmz:
### Related :handshake:
**[VEN-843](https://helsinkisolutionoffice.atlassian.net/browse/VEN-843)** 

### Automated tests :gear:️
TBD

## Screenshots 📸 
image

The tests still have to be fixed

### DIFF
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -1,3 +1,5 @@
+import threading
+
 import graphene
 from django.core.exceptions import ValidationError
 from django.db import transaction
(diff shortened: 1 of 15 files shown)


leases/schema/mutations.py
```diff
@@ -419,42 +419,21 @@ class Input(SendExistingInvoicesInput):
...
-        result_response = SendExistingInvoicesType(
-            successful_orders=successful_orders,
-            failed_orders=failed_orders,
-            failed_leases=failed_leases,
+        service = BerthInvoicingService(
+            request=info.context,
+            profile_token=profile_token,
+            due_date=input.get("due_date"),
         )
+        t1 = threading.Thread(target=service.send_invoices, args=[])
```
haavikko, line 433: If I remember correctly, the currently active language in Django is thread-specific, so if translations are needed, a translation.activate() call is needed in send_invoices. How is the language for "email to admins" determined? Does it use currently active translation or does it get it from elsewhere?

haavikko (approved):
Good! lets merge it so testing can be done, need to redo with the 4000 emails

leases/schema/mutations.py
a-rmz: It takes the default language from the settings, which is Finnish. Since the admins are only Finnish-speaking, that's not a concern. 👍

PR #406, opened 2020-12-15
Fixes for invoicing

a-rmz:
## Description :sparkles:
* Add the plain text [en] and [sv] for berth renew email
* Remove `cancel_url` from berth offer emails
* Remove `required` price tier for pier input (make `TIER_1` as default for piers)


resources/schema/mutations.py
```diff
@@ -466,7 +466,6 @@ class PierInput(AbstractAreaSectionInput):
 class CreatePierMutation(graphene.ClientIDMutation):
     class Input(PierInput):
         harbor_id = graphene.ID(required=True)
-        price_tier = graphene.Field("payments.schema.PriceTierEnum", required=True)
```
ahie, line 469: Should this instead be kept and only the `required` part removed?

templates/email/messages/berth_offer_new_en.html
```diff
@@ -192,7 +192,7 @@ <h3>Price of the boat berth</h3>
             <p>1. Notify us via this link:</p>
 
             <a style="display: block; padding: 16px; color: #0000BF; background-color: transparent; border: solid 2px #0000BF; text-align: center; text-decoration: none;"
-               href="{{ cancel_url }}">No thank you, I do not wish to reserve
```
ahie, line 195: For these I would remove the whole link (or comment them out, but I'm not sure if all emails clients can handle commented html). If invoices are sent out with broken cancellation links customers will be confused. 

Alternatively replace with text that says something along the lines of "If you do not want to reserve this berth, please contact <vene office email>"

ahie (commented):
A couple comments, but otherwise LGTM!

resources/schema/mutations.py
a-rmz: The `price_tier` is already defined on the `PierInput` as non-required.

ahie: 👍

templates/email/messages/berth_offer_new_en.html
a-rmz: Ok ok, I was thinking that I wouldn't touch the template much since it's only a temporary fix until we get the cancel page, but I guess we can comment out the whole section. 👍

haavikko: agree

haavikko (approved):
I think we can push this to testing and fix cancel links later.

templates/email/messages/berth_offer_new_en.html
a-rmz: @ahie fixed 👍

ahie: 👍

PR #408, opened 2020-12-15
HOTFIX | Remove invalid lease contract update

a-rmz:
## Description :sparkles:
* `lease` does not have an updatable `contract` field
* Catch for errors when creating the new lease to add them to the failed leases

### DIFF
--- a/leases/services/invoice.py
+++ b/leases/services/invoice.py
@@ -89,7 +89,7 @@ class BaseInvoicingService:
             lease.contract.pk = None
             lease.contract.save()
 
-        lease.save(update_fields=["start_date", "end_date", "application", "contract"])
+        lease.save(update_fields=["start_date", "end_date", "application"])
 
         return lease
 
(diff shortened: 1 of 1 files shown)


leases/services/invoice.py
```diff
@@ -89,7 +89,7 @@ def create_new_lease(
             lease.contract.pk = None
             lease.contract.save()
 
-        lease.save(update_fields=["start_date", "end_date", "application", "contract"])
+        lease.save(update_fields=["start_date", "end_date", "application"])
```
ahie, line 92: The newly created lease will still get a copy of the contract as well?

a-rmz: Yup. That's created by the
```python
lease.contract.pk = None
lease.contract.save()
```

ahie: 👍 just thought to double check.

ahie (approved):

PR #411, opened 2020-12-16
Hotfix | disable Merihaka

a-rmz:
## Description :sparkles:
* Add a filter to only invoice leases with active berths
* Update the fixtures


ahie (approved):
👍

leases/models.py
```diff
@@ -136,6 +136,7 @@ def get_renewable_leases(self, season_start: date = None) -> QuerySet:
         # Exclude leases that have already been assigned to the same customer and berth on the future
         leases: QuerySet = qs.exclude(Exists(future_leases.values("pk"))).filter(
             # Only allow leases that are auto-renewing and have been paid
+            berth__is_active=True,
```
haavikko, line 139: Looks good&simple. Currently no berth in test env have is_active=False, so this should not have unintended effects
>>> from resources.models import Berth
>>> Berth.objects.all().count()
4101
>>> Berth.objects.filter(is_active=False)
<QuerySet []>

haavikko (commented):
OK! looks simple

leases/models.py
a-rmz: Indeed. When we load the pier with the disabled Merihaka piers, those should be the only ones excluded under that critera. 👍

PR #413, opened 2020-12-16
Add check for berth switch emails

a-rmz:
## Description :sparkles:
* Check if the `berth_switch` has a `reason` to avoid failing calling `.title` if `reason is None`

### Closes :no_good_woman:
https://sentry.hel.ninja/aok/berth-api/issues/12382/


haavikko (approved):
OK!

PR #414, opened 2020-12-17
Fix test failure when generated berth lease start and end year are different

haavikko:
Tests started failing this morning when generated berth lease end occurs in 2021. Fixed by @freeze_time.

### Closes :no_good_woman:


a-rmz (approved):
👍

PR #416, opened 2020-12-17
HOTFIX | Only invoice leases with contracts

a-rmz:
## Description :sparkles:
* Add a check to only renew leases with contracts


haavikko (approved):
Let's roll

PR #418, opened 2020-12-17
HOTFIX | Invoicing fixes

a-rmz:
## Description :sparkles:
* Increase the error limit to 100
* Ignore counting errors if they're due to a customer is missing email
* Send the email to the admins also when the proccess fails because it hits the error limit

## Screenshots :camera_flash:
Email notice when the process fails
image


haavikko (approved):
That was really quick! Lets push to staging and run tests.

PR #419, opened 2020-12-17
HOTFIX | Check that the fields exist when creating an order

a-rmz:
## Description :sparkles:
* When taking the fields form the order, make sure that they're not empty to avoid failing


ahie (approved):

haavikko (approved):
OK go for it!

PR #420, opened 2020-12-17
HOTFIX | Fix broken copying of berth contracts when renewing leases

ahie:


leases/services/invoice.py
```diff
@@ -90,9 +90,9 @@ def create_new_lease(
         # Make copy of attached contract, if one exists
         if hasattr(lease, "contract") and lease.contract is not None:
             lease.contract.pk = None
-            lease.contract.save()
 
         lease.save(update_fields=["start_date", "end_date", "application"])
+        lease.contract.save()
```
a-rmz, line 95: Let's get rid of this one for now

a-rmz (approved):
Let's do it!

PR #424, opened 2020-12-21
VEN-1007 profiling tools

haavikko:
## Description :sparkles:

Enables profiling output from dev server.

BERTH_PROFILING_ENABLED=1 python manage.py runserver

What do you think, should we enable something like this in the staging by default?

### DIFF
--- a/utils/profiling.py
+++ b/utils/profiling.py
@@ -1,0 +1,117 @@
+import cProfile
+import functools
+import io
+import os
+import pstats
+import sys
+import threading
+import time
+
+__all__ = [
+    "profiling_support",
+    "run_with_profiler",
+    "profile_call",
+    "ProfilingMiddleware",
+]
+
+
+def profiling_support(func):
+    """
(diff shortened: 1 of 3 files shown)


utils/profiling.py
```diff
@@ -0,0 +1,120 @@
...
+        ps = pstats.Stats(profile, stream=s).sort_stats(sort_val)
+        ps.print_stats(num_results)
+
+        if os.getenv("PROFILE_CALLEES") == "1":
+            ps.print_callees(num_results)
+
+        if os.getenv("PROFILE_CALLERS") == "1":
+            ps.print_callers(num_results)
+
+    sys.stderr.write(s.getvalue())
```
a-rmz, line 67: Instead of directly printing to `stderr`, we could use [python logging](https://docs.python.org/3/library/logging.html) and set up a logger on the settings specific for the profiling.

Then, we can easily configure whether the results are printed to the console or to a file or both.

PR #427, opened 2020-12-22
Order paid_at

a-rmz:
## Description :sparkles:
* Add a `paid_at` prop to the `Order` model
  * Add the prop to the `OrderNode` type
* Add details about `OrderLogEntry` to Django Admin

```graphql
query Leases {
  orders {
    edges {
      node {
        id
        paidAt
      }
    }
  }
}
```

## Screenshots :camera_flash:
image

### DIFF
--- a/payments/admin.py
+++ b/payments/admin.py
@@ -1,5 +1,5 @@
 from django.contrib import admin
-from django.forms import ChoiceField, UUIDField
+from django.forms import TypedChoiceField, UUIDField
 from django.utils.translation import gettext_lazy as _
 
 from leases.models import BerthLease, WinterStorageLease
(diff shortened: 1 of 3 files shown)


payments/models.py
```diff
@@ -813,7 +821,7 @@ def __str__(self):
 class OrderLogEntry(UUIDModel, TimeStampedModel):
     order = models.ForeignKey(
         Order,
-        verbose_name=_("order log entry"),
```
ahie, line 816: Not sure why this is changed.

payments/admin.py
```diff
@@ -193,8 +202,19 @@ class OrderInline(admin.StackedInline):
     exclude = ("_product_content_type", "_lease_content_type")
 
 
-admin.site.register([WinterStorageProduct, OrderLine, OrderLogEntry])
+class OrderLogEntryAdmin(admin.ModelAdmin):
+    list_display = ("id", "order_id", "from_status", "to_status", "created_at")
+    list_filter = ("from_status", "to_status")
+    search_fields = ("order__id",)
```
ahie, line 208: A customer name search field could probably come in handy here.

ahie (approved):
LGTM, couple small comments left.

payments/models.py
a-rmz: I realised it was the wrong value, since it refers to `order`, not the log itself.

ahie: Oh right, missed that as well.

payments/admin.py
a-rmz: I'll add it to the `OrderAdmin` one 👍  then you can see the log entry as inlines on the `Order` page

a-rmz: Done
image

ahie: 👍

ahie (approved):

PR #428, opened 2020-12-23
Add custom logger for requests

a-rmz:
## Description :sparkles:
For better visibility on logs, the new logger dumps the request-related logs as json so it can be parsed by the log aggregation services. It also adds more details to the log.

### DIFF
--- a/berth_reservations/logging.py
+++ b/berth_reservations/logging.py
@@ -1,0 +1,52 @@
+import logging
+import time
+
+logger = logging.getLogger("requests")
+
+
+__all__ = ["LogRequestFilter", "RequestLogger"]
+
+
+class LogRequestFilter(logging.Filter):
+    """Since the requests are being logged by the RequestLogger,
+    we filter the django.server logs to avoid having duplicate logs"""
+
+    def filter(self, record: logging.LogRecord):
+        return record.name != "django.server"
+
+
+class RequestLogger:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        exec_time = time.time()
+        response = self.get_response(request)
+        exec_time = int((time.time() - exec_time) * 1000)
+
+        try:
+            path = request.get_full_path()
+            status = response.status_code
+            method = request.method
+
+            message = f"{method} {path} {status}"
+            context = {
+                "host": request.get_host(),
+                "method": method,
+                "agent": request.headers.get("USER_AGENT", ""),
+                "body": request.body.decode(request.encoding or "utf-8"),
+                "path": path,
+                "status": status,
+                "exec_time": exec_time,
+            }
+            level = logging.INFO
+            if 400 >= status > 500:
+                level = logging.WARNING
+            elif status >= 500:
+                level = logging.ERROR
+
+            logger.log(level, message, extra=context)
+        except Exception as e:
+            logger.exception("Failed logging", exc_info=e)
+
+        return response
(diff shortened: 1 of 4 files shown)


berth_reservations/logging.py
```diff
@@ -0,0 +1,49 @@
...
+        path = request.get_full_path()
+        status = response.status_code
+        method = request.method
+
+        message = f"{method} {path} {status}"
+        context = {
+            "host": request.get_host(),
+            "method": method,
+            "agent": request.headers["USER_AGENT"],
+            "body": request.body.decode(request.encoding or "utf-8"),
```
ahie, line 36: Is it possible for `request.body` to be `None`? For example `GET` requests?

berth_reservations/logging.py
```diff
@@ -0,0 +1,49 @@
...
+    def __call__(self, request):
+        exec_time = time.time()
+        response = self.get_response(request)
+        exec_time = int((time.time() - exec_time) * 1000)
+
+        path = request.get_full_path()
+        status = response.status_code
+        method = request.method
+
+        message = f"{method} {path} {status}"
```
ahie, line 31: Should the message be more descriptive, e.g. something like `f"Logged request: {method} {path} {status}"`?

berth_reservations/logging.py
ahie: It might be safer to wrap everything in a try/catch here. If this middleware raises for any reason the client will never get a response back?

a-rmz: It will log an empty string.

berth_reservations/logging.py
a-rmz: I think with all the fields it will be pretty clear that the log comes from a request.

berth_reservations/logging.py
a-rmz: Most of the getters should be safe, but I wrapped in a `try/except` to be on the safe side.

berth_reservations/logging.py
```diff
@@ -0,0 +1,52 @@
...
+
+class LogRequestFilter(logging.Filter):
+    """Since the requests are being logged by the RequestLogger,
+    we filter the django.server logs to avoid having duplicate logs"""
+
+    def filter(self, record: logging.LogRecord):
+        return record.name != "django.server"
+
+
+class RequestLogger:
```
ahie, line 18: Other middleware seem to inherit from `MiddlewareMixin`, which adds async capability to the call. Might be useful here as well?

a-rmz: The `MiddlewareMixin` is to add backwards compatibility to the [pre-Django 1.10 middlewares](https://docs.djangoproject.com/en/3.1/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware).

The [example](https://docs.djangoproject.com/en/3.1/topics/http/middleware/#writing-your-own-middleware) on the django docs doesn't add it, and since we're not using the previous middleware style, we don't need it.
```python
class SimpleMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response
        # One-time configuration and initialization.

    def __call__(self, request):
        # Code to be executed for each request before
        # the view (and later middleware) are called.

        response = self.get_response(request)

        # Code to be executed for each request/response after
        # the view is called.

        return response
```

ahie: Oh, I see. The django >=3.1 way to add async support seems to be with the decorators, e.g. `@sync_and_async_middleware`. But I don't think there's much use to add that complexity here.

ahie (approved):
Lets give it a spin :shipit:

berth_reservations/logging.py
a-rmz: Yes! We're not using async for views, so not really necessary here. 👍

PR #434, opened 2021-01-07
Add filtering options to piers and berths

ahie:
Enables filtering berths by harbor and pier, and piers by harbor. This enables simplifying deeply nested queries on the FE.

### DIFF
--- a/resources/schema/types.py
+++ b/resources/schema/types.py
@@ -4,7 +4,7 @@ import graphql_geojson
 from django.utils.translation import gettext_lazy as _
 from graphene import relay
 from graphene_django.fields import DjangoConnectionField
-from graphene_django.filter import DjangoFilterConnectionField
+from graphene_django.filter import DjangoFilterConnectionField, GlobalIDFilter
 from graphene_django.types import DjangoObjectType
 
 from applications.models import BerthApplication, WinterStorageApplication
@@ -69,6 +69,7 @@ class PierNode(graphql_geojson.GeoJSONType):
             "gate",
             "lighting",
             "suitable_boat_types",
+            "harbor",
         ]
         geojson_field = "location"
         interfaces = (relay.Node,)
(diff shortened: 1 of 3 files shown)


a-rmz (commented):
Could you add some tests for this? 😇

resources/schema/types.py
```diff
@@ -83,6 +85,14 @@ class BerthNodeFilterSet(django_filters.FilterSet):
...
     )
     is_available = django_filters.BooleanFilter()
+    harbor = GlobalIDFilter(method="filter_by_harbor")
+    pier = GlobalIDFilter(method="filter_by_pier")
+
+    def filter_by_pier(self, queryset, name, value):
+        return queryset.filter(pier__id=from_global_id(value))
+
+    def filter_by_harbor(self, queryset, name, value):
+        return queryset.filter(pier__harbor__id=from_global_id(value))
```
a-rmz, line 95:
```suggestion
    harbor = GlobalIDFilter(field_name="pier__harbor", lookup_expr="id")
    pier = GlobalIDFilter(field_name="pier", lookup_expr="id")
```

a-rmz:
One last question: what happens if you pass both `pier/harbor` filters when the pier belongs to a different pier?

resources/schema/types.py
ahie: Done.

ahie:
> Could you add some tests for this? 😇

Done.

> One last question: what happens if you pass both pier/harbor filters when the pier belongs to a different pier?

It will just return zero results. Added test case for this.

a-rmz (approved):
Looks good now 👍 one last thing: could you add ass comment to the `BerthNode` description that if you use both `harbor` and `pier` filters with a pier that does not belong the harbour, it will not return anything.

ahie:
> Looks good now 👍 one last thing: could you add ass comment to the `BerthNode` description that if you use both `harbor` and `pier` filters with a pier that does not belong the harbour, it will not return anything.

Added description to the `berths` query.

a-rmz (approved):

PR #435, opened 2021-01-12
VEN-1069 | Cancelled and rejected order dates

a-rmz:
## Description :sparkles:
* Add a `cancelledAt` and `rejectedAt` props to the `Order` model and node
* Update the `paidAt` prop to a model annotation for better performance

### Closes :no_good_woman:
**[VEN-1069](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1069):** Add cancelled_at to order

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_models.py::test_order_paid_and_created
```

## Screenshots :camera_flash:
**Fields on django admin**
image

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -1,7 +1,7 @@
 import logging
-from datetime import date, datetime
+from datetime import date
 from decimal import Decimal
-from typing import Optional, Union
+from typing import Union
 
 from dateutil.utils import today
 from django.contrib.contenttypes.fields import GenericForeignKey
@@ -9,7 +9,8 @@ from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator
 from django.db import models
-from django.db.models import Q, UniqueConstraint
+from django.db.models import Max, OuterRef, Q, Subquery, UniqueConstraint, Value
+from django.db.models.functions import Coalesce
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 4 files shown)


payments/models.py
```diff
@@ -262,6 +263,33 @@ def update_expired(self) -> int:
 
         return too_old_waiting_orders.count()
 
+    def get_queryset(self):
```
haavikko, line 266: If paid_at/rejected_at/cancelled_at is not needed, is there any significant performance overhead added to the queries? I don't know if Django is able to optimize the query, and if it matters here at all. If there is, should there be an additional manager instead of overriding the default manager?

I think the changes can be merged as is, optimize later if there's need

payments/tests/test_payments_models.py
```diff
@@ -886,3 +887,33 @@ def test_order_set_status_no_application(berth):
...
+def test_order_paid_and_created(order, from_status, to_status):
+    order.status = from_status
+    order.save()
+
+    order.set_status(to_status)
+    assert order.status == to_status
+
+    order = Order.objects.get(id=order.id)
+    if to_status == OrderStatus.PAID:
+        assert order.paid_at == now()
```
haavikko, line 915: Maybe add a assertion that order.paid_at/rejected_at/cancelled_at == None on a order without any matching OrderLogEntries?

```suggestion
        assert order.paid_at == now()
        assert order.rejected_at == None
        assert order.cancelled_at == None        
```

haavikko (approved):
I think it can go in as is, but If you think it's useful, maybe add those extra assertions to tests?

payments/tests/test_payments_models.py
a-rmz: Updated

payments/models.py
a-rmz: It doesn't add that much overhead by executing it like this, but if several order instances need it, it gets quite heavy. Regarding the manager, it's better to keep it like that. It doesn't really add any complications or anything and does not modify the original manager. On top of that, Grapehene internally relies on the `.objects` manager to do the magic, so if we were to put this annotations on a different manager, we would have to manually add all the resolvers.

PR #439, opened 2021-01-14
Invalidate expired payments with nightly cronjob

haavikko:
### Closes :no_good_woman:
**[VEN-783](https://helsinkisolutionoffice.atlassian.net/browse/VEN-783):**


payments/management/commands/expire_too_old_unpaid_orders.py
```diff
@@ -1,18 +1,40 @@
...
+        num_of_updated_orders = Order.objects.expire_too_old_unpaid_orders(
+            settings.EXPIRE_WAITING_ORDERS_OLDER_THAN_DAYS
+        )
+        if dry_run:
+            logger.info(
+                "Dry run, do not save changes {} order(s) would be expired.".format(
+                    num_of_updated_orders
+                )
+            )
+            raise Exception("Initiate rollback")
```
a-rmz, line 38: Is this the best approach? Would it be safer to just get the "expirable" orders rather than change-and-rollabck?

payments/models.py
```diff
@@ -251,14 +252,26 @@ def _order_type_orders(self, product_ct: ContentType, lease_ct: ContentType):
...
         for order in too_old_waiting_orders:
+
+            if order._product_object_id and not order._get_product_from_object_id():
+                # _product_object_id contains an UUID to a product that no longer exists.
+                # set_status() will fail in this case, and instead of EXPIRED these
+                # should be set to ERROR.
+                sys.stdout.write(f"skip invalid order {order}")
+                continue
+
+            sys.stdout.write(f"{order}")
```
a-rmz, line 271:
```suggestion
            logger.info(order)
```

payments/models.py
```diff
@@ -251,14 +252,26 @@ def _order_type_orders(self, product_ct: ContentType, lease_ct: ContentType):
...
+            status=OrderStatus.WAITING, due_date__lt=expire_before_date,
         )
 
         for order in too_old_waiting_orders:
+
+            if order._product_object_id and not order._get_product_from_object_id():
+                # _product_object_id contains an UUID to a product that no longer exists.
+                # set_status() will fail in this case, and instead of EXPIRED these
+                # should be set to ERROR.
+                sys.stdout.write(f"skip invalid order {order}")
```
a-rmz, line 268:
```suggestion
                logger.info(f"skip invalid order: {order.id}")
```

payments/management/commands/expire_too_old_unpaid_orders.py
haavikko: The dry-run option is useful for manual testing in staging (don't need to rebuild test data anew each time). This would be unsafe if the code had side effects that aren't limited to database state (such as, in future, if an email was sent on state change).
But I'll refactor this to not use a rollback

haavikko: done

payments/models.py
```diff
@@ -251,17 +251,40 @@ def _order_type_orders(self, product_ct: ContentType, lease_ct: ContentType):
...
+                # _product_object_id contains an UUID to a product that no longer exists.
+                # set_status() will fail in this case, and instead of EXPIRED these
+                # should be set to ERROR.
+                logger.info(f"Product missing from order, skip invalid order {order}")
+                continue
+
+            if not dry_run:
+                order.set_status(
+                    OrderStatus.EXPIRED,
+                    comment=f"{_('Order expired at')} {expire_before_date}",
```
a-rmz, line 284: Should we log the current timestamp for better traceability?

a-rmz: Scratch that, with the `created_at` we can figure the exact time if needed.

a-rmz (approved):

PR #440, opened 2021-01-14
VEN-1063 | Add an email template for order cancellation

Nizar-Rahme:
## Description :sparkles:
- Add order cancellation email templates in all three languages 

### Closes :no_good_woman:
**[VEN-1063](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1063):** Order cancellation confirmation email


templates/email/messages/order_cancelled_en.html
```diff
@@ -0,0 +1,22 @@
+<p>Dear customer, we have received your cancellation. The invoice, application and agreement related to the berth are cancelled as of {{ order.rejectedAt }}.</p>
+
+<h3>Object</h3>
```
a-rmz, line 3: `Object` sounds a bit odd in English. Not sure what would  be the best but it sounds a bit strange.

templates/email/messages/order_cancelled_en.html
```diff
@@ -0,0 +1,22 @@
+<p>Dear customer, we have received your cancellation. The invoice, application and agreement related to the berth are cancelled as of {{ order.rejectedAt }}.</p>
```
a-rmz, line 1: The prop is on `snake_case`, same for the other templates. :)
```suggestion
<p>Dear customer, we have received your cancellation. The invoice, application and agreement related to the berth are cancelled as of {{ order.rejected_at }}.</p>
```

Nizar-Rahme: Fixed

templates/email/messages/order_cancelled_en.html
TatuArvela: Should this be Paikka / Plats / Berth?

Nizar-Rahme: It's `Kohde` in Finnish, there's `Paikka` a couple of lines below. What do you suggest @TatuArvela would be the right translation for `Kohde` in English? I also got translated `Område` in Swedish, do you think that's right?

TatuArvela: I think we're using "Paikka / Plats / Place" as the generic term when talking about both berths and winter storage places, so we could consider using that.

If using "Kohde", I'd say "Location" would be a good English translation. "Område" means "area", but can make sense in this context I think.

Nizar-Rahme: Fixed

a-rmz (approved):

PR #442, opened 2021-01-15
VEN-1007 | Individual berth performance

a-rmz:
## Description :sparkles:
* Upgrade (most) deps
* Add [`django-silk`](https://github.com/jazzband/django-silk) for profiling
* Add DataLoaders through a Graphene middleware
  * Add DataLoaders to resources
  * Add DataLoader for lease customers
* Update CountConnection to support DataLoaders
* Add a `HarborManager` to annotate some fields

### Closes :no_good_woman:
**[VEN-1007](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1007):** Improve Berths page loading time
[Partly closes it]

### Manual testing :construction_worker_man:
Test with either of the following queries:
```graphql
query NESTED_INDIVIDUAL_HARBOR($harborId: ID!) {
  harbor(id: $harborId) {
    id
    properties {
      name
      numberOfPlaces
      numberOfFreePlaces
      streetAddress
      zipCode
      municipality
      wwwUrl
      imageFile
      maps {
        id
        url
      }
      servicemapId
      maxWidth
      numberOfPlaces
      numberOfFreePlaces
      numberOfInactivePlaces
      piers {
        edges {
          node {
            id
            properties {
              identifier
              priceTier
              electricity
              wasteCollection
              lighting
              gate
              suitableBoatTypes {
                name
              }
              berths {
                edges {
                  node {
                    id
                    isActive
                    number
                    width
                    length
                    depth
                    mooringType
                    comment
                    leases {
                      edges {
                        node {
                          customer {
                            id
                          }
                          startDate
                          endDate
                          isActive
                        }
                      }
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
}

query INDIVIDUAL_HARBOR($harborId: ID!) {
  harbor(id: $harborId) {
    id
    properties {
      name
      numberOfPlaces
      numberOfFreePlaces
      streetAddress
      zipCode
      municipality
      wwwUrl
      imageFile
      maps {
        id
        url
      }
      servicemapId
      maxWidth
    }
  }
  piers(harbor: $harborId) {
    edges {
      node {
        id
        properties {
          identifier
          priceTier
          electricity
          wasteCollection
          water
          lighting
          gate
          suitableBoatTypes {
            name
          }
        }
      }
    }
  }
  berths(harbor: $harborId) {
    totalCount
    count
    edges {
      node {
        id
        isActive
        isAvailable
        number
        width
        length
        depth
        mooringType
        comment
        leases {
          edges {
[...]

PR #444, opened 2021-01-18
VEN-1007 | Revert dependency upgrades

a-rmz:
## Description :sparkles:
* Revert dependency upgrades that broke the JWT auth

### Related :handshake:
#442


haavikko (approved):
OK!

PR #446, opened 2021-01-21
HOTFIX | Harbor manager

a-rmz:
## Description :sparkles:
* Replace the `models.Manager` for parler's `TranslatableManager` since this breaks the Django Admin

### DIFF
--- a/.prod/on_deploy.sh
+++ b/.prod/on_deploy.sh
@@ -1,9 +1,4 @@
 #!/bin/sh
 
-python /app/manage.py makemigrations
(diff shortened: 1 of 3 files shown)


haavikko (approved):
OK go ahead!
Just a comment - the changes on_deploy.sh were added at the same time with django-silk and now reverted, does django-silk need those?

a-rmz:
Nope, those had to be only run one time. Now the migrations were run and the statics collected, so no neeed to do it again.

PR #450, opened 2021-01-28
Ven 1090/one time sms

mikkovihonen:
### Closes :no_good_woman:
**[VEN-1090](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1090):**


payments/management/commands/one_time_sms_message.py
```diff
@@ -0,0 +1,171 @@
...
+
+        profile_service = ProfileService(profile_token=profile_token)
+        sms_service = SMSNotificationService()
+        payment_provider = BamboraPayformProvider(
+            ui_return_url=settings.VENE_UI_RETURN_URL
+        )
+
+        MESSAGE = (
+            "Hyvä asiakas. Tässä maksulinkki venepaikan maksamiseen purjehduskaudelle 2021. "
+            "Maksathan laskun 31.1.2021 mennessä.\n\n"
```
haavikko, line 109: Kalle Järveläinen  2 hours ago
note: Need to change the date value for the SMS as we are a bit late with this. New date for the messages is 7.2.2021

haavikko (changes requested):

payments/management/commands/one_time_sms_message.py
```diff
@@ -0,0 +1,171 @@
...
+
+                # Save the phone to the order, in case it is required later
+                order.customer_phone = customer.phone
+                order.save()
+            else:
+                logger.debug(f"Phone found for: {order.customer.id}")
+
+            message = MESSAGE.format(
+                url=payment_provider.get_payment_email_url(order, "fi")
+            )
```
haavikko, line 146: Do we need two urls in the message, "fi" and "en"?

payments/management/commands/one_time_sms_message.py
```diff
@@ -0,0 +1,171 @@
...
+                Q(Q(customer_phone__isnull=True) | Q(customer_phone=""))
+            ).count()
+
+            self.stdout.write(f"Invoices to be sent: {invoices}")
+            self.stdout.write(f"Customers without phone: {without_phone}")
+
+            return
+
+        for order in orders:
+            # If the order doesn't have a customer phone number, fetch it from Profiili
```
haavikko, line 127: We should try do avoid errors inside the sms sending loop, so that the action doesn't stop halfway due to some error.
For example, is it possible that get_profile() fails, for example some customer doesn't have a profile? Maybe fetch the profiles outside of the loop?

payments/management/commands/one_time_sms_message.py
```diff
@@ -0,0 +1,171 @@
...
+            if not order.customer_phone:
+                logger.debug(f"Fetching profile for: {order.customer.id}")
+                customer = profile_service.get_profile(order.customer.id)
```
haavikko, line 130: put a try-catch block around get_profile call, as a defensive measure, and put to "no_customer_phone" list in case of failure?

payments/management/commands/one_time_sms_message.py
mikkovihonen: Changed

payments/management/commands/one_time_sms_message.py
mikkovihonen: I don't think so.

payments/management/commands/one_time_sms_message.py
mikkovihonen: Will do.

payments/management/commands/one_time_sms_message.py
mikkovihonen: Try catch block addresses this.

haavikko (approved):
OK great!

payments/management/commands/one_time_sms_message.py
```diff
@@ -0,0 +1,175 @@
...
+        sms_service = SMSNotificationService()
+        payment_provider = BamboraPayformProvider(
+            ui_return_url=settings.VENE_UI_RETURN_URL
+        )
+
+        MESSAGE = (
+            "Hyvä asiakas. Tässä maksulinkki venepaikan maksamiseen purjehduskaudelle 2021. "
+            "Maksathan laskun 7.2.2021 mennessä.\n\n"
+            "Dear Customer. Here is a payment link for paying for a berth for the sailing season 2021. "
+            "Please pay the invoice by February 7, 2021.\n\n"
```
a-rmz, line 111: IIRC the message should be in Swedish, not English. The ticket should have the latest message.

PR #455, opened 2021-02-10
VEN-1071 | Resend error invoices

a-rmz:
## Description :sparkles:
* When resending the invoice, it checks if it has `error` state and sets it to `waiting/offered` status in case it has been fixed
* Enable order status changes from `error -> waiting`
* Add a check to raise an error if the order doesn't have an email when sending the invoice

### Closes :no_good_woman:
**[VEN-1071](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1071):** Admins can't send fixed invoices

### Automated tests :gear:️
```shell
$ python payments/tests/test_payments_mutations_resend_order.py
```

### Manual testing :construction_worker_man:
1. Set an order and the lease to `ERROR` states
2. Exectue the the first mutation for that order

```graphql
mutation RESEND_ORDER_MUTATION($input: ResendOrderMutationInput!) {
    resendOrder(input: $input) {
        failedOrders {
            id
            error
        }
        sentOrders
    }
}
```
```json
{
  "input": {
    "orders": ["..."],
    "dueDate": "2021-02-15"
  }
}
```
3. Check the terminal for the sent email, and the statuses of the order and lease
```graphql
query Orders {
  order(id: "") {
    status
    lease {
      ... on BerthLeaseNode{
        status
      }
    }
  }
}
```

## Additional notes :spiral_notepad:
This only covers the case of manually resending one invoice, the case for batch invoicing will come in a separate PR.

PR #457, opened 2021-02-11
VEN-1071 | Resend batch error invoices

a-rmz:
## Description :sparkles:
* Before sending batch invoices, attempt re-sending the failed invoices
  * Clears the status from orders/leases
  * Updates the `order.customer_*`
  * Re-sends the invoice

### Closes :no_good_woman:
**[VEN-1071](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1071):** Admins can't send fixed invoices

### Related :handshake:
#455 

1. Set an order and the lease to `ERROR` states
2. Call the `sendExistingBerthInvoices` mutation
```graphql
mutation SendFailed($send: SendExistingBerthInvoicesMutationInput!) {
  sendExistingBerthInvoices(input: $input){
    ok
  }
}
```
```json
{
  "input": {
    "profileToken": "<profile token from Admin UI>"
  }
}
```
3. The email should be printed on the console
4. Check the new status of the order
```graphql
query Orders {
  order(id: "<order id>") {
    status
    lease {
      ... on BerthLeaseNode {
        status
      }
    }
  }
}
```

This PR depends on #455 to be merged

### DIFF
--- a/leases/services/invoice.py
+++ b/leases/services/invoice.py
@@ -3,15 +3,18 @@ from datetime import date
 from typing import Dict, List, Union
 from uuid import UUID
 
+import pytz
 from anymail.exceptions import AnymailError
 from dateutil.relativedelta import relativedelta
 from dateutil.utils import today
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group
 from django.core.exceptions import ValidationError
 from django.db import DataError, IntegrityError, transaction
 from django.db.models import QuerySet
 from django.http import HttpRequest
+from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _
 from django_ilmoitin.utils import send_notification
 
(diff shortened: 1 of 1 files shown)


leases/services/invoice.py
```diff
@@ -270,6 +287,32 @@ def send_invoices(self) -> None:  # noqa: C901
...
+        logger.debug("Resending failed invoices")
+        orders = self.get_failed_orders(self.season_start)
+
+        for order in orders:
+            order.set_status(
+                OrderStatus.WAITING,
+                comment=f"{get_ts()}: {_('Cleanup the invoice to attempt resending')}\n",
+            )
+
+            if not order.customer_email and not order.customer_phone:
```
haavikko, line 300: Should always update, in case email in profile has been updated and order.customer_email has not?

leases/services/invoice.py
```diff
@@ -285,3 +328,13 @@ def get_product(lease: BerthLease) -> BerthProduct:
     @staticmethod
     def get_valid_leases(season_start: date) -> QuerySet:
         return BerthLease.objects.get_renewable_leases(season_start=season_start)
+
+    @staticmethod
+    def get_failed_orders(season_start: date) -> QuerySet:
```
haavikko, line 333: Is it possible that a previous invoicing service run has left behind a Lease that is in error status, but no Order, in case Order creation fails ("No suitable product found")? Are the admins able to remove the Lease / how to resend invoice in that case?

a-rmz: We _shouldn't_ have any "no suitable product" errors since we moved to the newpricing schema. If this would be the case, we probably have bigger problems 😅

haavikko (approved):

PR #461, opened 2021-02-15
VEN-1083 | Change invoice status as paid

haavikko:
### Closes :no_good_woman:
**[VEN-1083](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1083):**


README.md
```diff
@@ -40,7 +40,7 @@ ALLOWED_HOSTS=*
 CORS_ORIGIN_ALLOW_ALL=1
 NOTIFICATIONS_ENABLED=1
 
-VENE_PAYMENTS_BAMBORA_API_URL=https://fake-bambora-api-url/api
+VENE_PAYMENTS_BAMBORA_API_URL=https://real-bambora-api-url/api
```
a-rmz, line 43: Any reason to change this?

payments/schema/mutations.py
```diff
@@ -413,7 +413,15 @@ def mutate_and_get_payload(cls, root, info, **input):
                 input["lease"] = lease
 
         try:
+            # handle case where input has both lease and status.
+            # set order status only after changing the lease, because setting order status
+            # usually triggers a change in lease status.
+            new_status = input.pop("status", None)
+            old_status = order.status
             update_object(order, input)
+            if new_status and new_status != old_status:
```
a-rmz, line 422: The `set_status` already checks that the old and new statuses are different, so we can skip that check.

README.md
haavikko: Some payment tests fail when running locally if this setting is used

haavikko: I think https://fake-bambora-api-url/api is meant to be used to trigger a simulated failure with bambora

a-rmz: AHHH right, good point.

a-rmz (approved):

PR #464, opened 2021-02-19
fix expired_at in cancellation email

haavikko:
### Closes :no_good_woman:
**[VEN-1063](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1063):**

### DIFF
--- a/berth_reservations/tests/conftest.py
+++ b/berth_reservations/tests/conftest.py
@@ -178,14 +178,14 @@ def notification_template_order_cancelled():
         notification = NotificationTemplate.objects.language("fi").create(
             type=value,
             subject="test order cancelled subject: {{ order.order_number }}!",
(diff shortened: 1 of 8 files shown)


a-rmz (commented):
If we're changing the context variable for `rejected`, we should also update the actual [templates](https://github.com/City-of-Helsinki/berth-reservations/blob/master/templates/email/messages/order_cancelled_en.html#L1) to have the same name.

Small nitpick: can we rename it to just `rejected_at`? The `order_` seems a bit redundant

a-rmz (approved):

a-rmz (changes requested):
@haavikko one last thing: can you also update the dummy context to render the proper date on the Preview?

a-rmz (approved):

PR #472, opened 2021-02-25
VEN-1177: add ApplicationStatus.REJECTED status, set application to r…

haavikko:
…ejected when customer rejects order

### Closes :no_good_woman:
**[VEN-1177](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1177):**


applications/migrations/0026_auto_20210225_0850.py
```diff
@@ -0,0 +1,57 @@
+# Generated by Django 3.1 on 2021-02-25 06:50
```
klemmari1, line 1: Would like to see the automatically generated migration filename to be renamed to something meaningful. For example `0026_add_rejected_application_status.py`.

a-rmz: Good catch @klemmari1 👍  Let's do it like that @haavikko

payments/models.py
```diff
@@ -775,12 +775,13 @@ def update_lease_and_application(self, new_status):
...
-                application = self.lease.application
+        if application := self.lease.application:
+            if new_status == OrderStatus.REJECTED:
+                application.status = ApplicationStatus.REJECTED
+                application.save(update_fields=["status"])
+            elif new_status in OrderStatus.get_paid_statuses():
                 application.status = ApplicationStatus.HANDLED
-                self.lease.application.save(update_fields=["status"])
-
+                application.save(update_fields=["status"])
```
a-rmz, line 784: At this point, would it make sense to extract the application logic to a util function like `get_lease_status`?

applications/migrations/0026_auto_20210225_0850.py
haavikko: fixed

payments/models.py
haavikko: added get_application_status. I also moved the logic for generating sticker numbers to its own function.

klemmari1:
This looks good to me! But perhaps I leave the approving for @a-rmz since I'm not super familiar with the project.

a-rmz (approved):

PR #473, opened 2021-02-25
VEN-1085 | Add boolean filter for application_code

klemmari1:
## Description :sparkles:

This filter will return:

- true: all applications with a non-empty `application_code`.
- false: all applications with an empty `applicaiton_code`.

### Closes :no_good_woman:
**[VEN-1085](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1085)** 

### Related :handshake:
**[VEN-1093](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1093)** 

### Automated tests :gear:️

```
pytest applications/tests/test_applications_new_schema_berth_queries.py::test_berth_applications_application_code_filter
pytest applications/tests/test_applications_new_schema_berth_queries.py::test_berth_applications_application_code_filter_empty_string
pytest applications/tests/test_applications_new_schema_berth_queries.py::test_berth_applications_application_code_filter_false
```

### DIFF
--- a/applications/new_schema/types.py
+++ b/applications/new_schema/types.py
@@ -61,6 +61,9 @@ class BerthApplicationFilter(django_filters.FilterSet):
     no_customer = django_filters.BooleanFilter(
         field_name="customer", lookup_expr="isnull"
     )
+    application_code = django_filters.BooleanFilter(
+        field_name="application_code", method="filter_application_code"
+    )
     order_by = django_filters.OrderingFilter(
         fields=("created_at",), label="Supports only `createdAt` and `-createdAt`.",
     )
@@ -74,6 +77,14 @@ class BerthApplicationFilter(django_filters.FilterSet):
     def filter_berth_switch(self, queryset, name, value):
         lookup = "__".join([name, "isnull"])
         return queryset.filter(**{lookup: not value})
+
+    def filter_application_code(self, queryset, name, value):
+        lookup = "__".join([name, "exact"])
+        return (
+            queryset.exclude(**{lookup: ""})
+            if value
+            else queryset.filter(**{lookup: ""})
+        )
 
 
 class BerthApplicationNode(DjangoObjectType):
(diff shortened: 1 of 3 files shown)


applications/new_schema/types.py
```diff
@@ -75,6 +78,10 @@ def filter_berth_switch(self, queryset, name, value):
         lookup = "__".join([name, "isnull"])
         return queryset.filter(**{lookup: not value})
 
+    def filter_application_code(self, queryset, name, value):
+        lookup = "__".join([name, "exact"])
+        return queryset.exclude(**{lookup: ""})
```
a-rmz, line 83: Would this also work if you pass `applicationCode: false`?

klemmari1: Hmm, atm it will probably do the same thing with `true` or `false` :thinking:

I guess this would be better:
```
    def filter_application_code(self, queryset, name, value):
        lookup = "__".join([name, "exact"])
        if value:
            return queryset.exclude(**{lookup: ""})
```

a-rmz:
```suggestion
    def filter_application_code(self, queryset, name, value):
        lookup = "__".join([name, "exact"])
        return queryset.exclude(**{lookup: ""}) if value else queryset.filter(**{lookup: ""})
```

klemmari1: Ah, so do we want to find all empty `application_code`s if `value==False`?

a-rmz: Wait no, the other way around 🤔  Basically: 
```
applicationCode: true --> applications WITH any code
applicationCode: false --> applications WITHOUT code
```

haavikko (approved):
Seems good to me!

applications/new_schema/types.py
klemmari1: Yea, that is what i meant :smile: `false` -> empty `application_code`s == appilcations without code.

PR #475, opened 2021-02-25
date_hierarchy in django admin for orders and leases

haavikko:
## Description :sparkles:

For debugging similar cases in the future, it would be useful to be able to filter orders by due_date and leases by start_date in the django admin.

### Closes :no_good_woman:
**[VEN-1177](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1177):**


klemmari1 (approved):
Looks good to me! :+1:

PR #476, opened 2021-02-26
VEN-1091 | Don't send invoice to non-billable customers

klemmari1:
## Description :sparkles:

Do not send invoices if an order customer is non-billable and set the order status as paid when accepting the order.

### Closes :no_good_woman:
**[VEN-1091](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1091):** 

### Automated tests :gear:️

```
pytest payments/tests/test_payments_utils.py::test_approve_order_with_non_billable_customer
```

### DIFF
--- a/payments/utils.py
+++ b/payments/utils.py
@@ -343,7 +343,12 @@ def approve_order(
             order.lease.application.status = ApplicationStatus.OFFER_SENT
             order.lease.application.save()
 
-    send_payment_notification(order, request, email, phone_number=order.customer_phone)
+    if order.customer.is_non_billable_customer():
+        order.set_status(OrderStatus.PAID_MANUALLY, "Non-billable customer.")
+    else:
+        send_payment_notification(
+            order, request, email, phone_number=order.customer_phone
+        )
 
 
 def send_cancellation_notice(order):
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -92,7 +92,9 @@ class CreateBerthLeaseMutation(graphene.ClientIDMutation):
             order = Order.objects.create(
                 customer=input["customer"], lease=lease, product=product
             )
-            get_contract_service().create_berth_contract(lease)
+            # Do not create a contract for non-billable customers.
+            if not application.customer.is_non_billable_customer():
+                get_contract_service().create_berth_contract(lease)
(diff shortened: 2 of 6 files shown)


a-rmz (changes requested):
### Add a check on the lease creation

On the lease creation, check if `customer.is_non_billable_customer()` and only generate the lease if it’s a billable customer.

* [leases/schema/mutations.py#L95](https://github.com/City-of-Helsinki/berth-reservations/blob/master/leases/schema/mutations.py#L95) [berth lease]
* [leases/schema/mutations.py#L260](https://github.com/City-of-Helsinki/berth-reservations/blob/master/leases/schema/mutations.py#L260) [WS lease]

### Add a check on the service invoice

On [leases/services/invoice.py#L92-L93](https://github.com/City-of-Helsinki/berth-reservations/blob/master/leases/services/invoice.py#L92-L93)

```python
if hasattr(lease, "contract") and lease.contract is not None:
     contract = lease.contract
```

add a also a check if `not lease.customer.is_non_billable()` to only create contracts for billable customers.

a-rmz (approved):
Good job!

payments/utils.py
```diff
@@ -343,7 +343,12 @@ def approve_order(
             order.lease.application.status = ApplicationStatus.OFFER_SENT
             order.lease.application.save()
 
-    send_payment_notification(order, request, email, phone_number=order.customer_phone)
+    if order.customer.is_non_billable_customer():
+        order.set_status(OrderStatus.PAID, "Non-billable customer.")
```
haavikko, line 347: Question to mostly Alex: In CreateBerthLeaseMutation and CreateWinterStorageLeaseMutation we use PAID_MANUALLY for non-billable customers, here it's PAID. We should use the same status in both cases, right? I remember discussion that PAID_MANUALLY is a better fit here, and PAID should be reserved for customers that have paid through the "normal" online channel. Anyway, the handling should be uniform.

a-rmz: Ahhh good point! I remember we had talked about `PAID` vs `PAID MANUALLY` somewhere else, then we should definitely go with the later. 👍

klemmari1: Good point! Fixed :+1:

PR #479, opened 2021-03-01
VEN-1024 | No berth confirmation email

a-rmz:
## Description :sparkles:
* Refactor `applications/notifications.py` as a separate sub-module
* Test that the `no berth` email has been sent
* Remove the unused `NO_SUITABLE_BERTH_NOTIFIED`, assuming that once it's been set to that status, the customer will ALWAYS be notified

### Closes :no_good_woman:
**[VEN-1024](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1024):**  Send No Berth email

### Automated tests :gear:️
```shell
$ pytest applications/tests/test_applications_new_schema_mutations.py::test_reject_berth_application
```

1. Send the `rejectBerthApplication` mutation
2. You should see on the console the rejection email for that application

The confirmation email was already being sent, the new assertions check for that


applications/notifications/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+from .dummy_context import load_dummy_context
+from .types import NotificationType
+
+for (value, label) in NotificationType.choices:
+    notifications.register(value, label)
+
+try:
+    load_dummy_context()
+except ProgrammingError as e:
+    if "django_content_type" in str(e):
```
haavikko, line 13: Please add comment why it's safe to ignore this (why and when the exception occurs)? Should load_dummy_context execute in production environment at all?

a-rmz: Sure I can add it there 👍 

The reason why we did that was because there were some race conditions when running the initial migrations. Since the `load_dummy_context` is executed on the module `__init__`, it's loaded even before the migrations are executed. So there were some conflicts when running the tests and callig the dummy context _before_ the content types were migrated. By adding this check, we just "dismiss" the loading of the dummy context so the application can be loaded correctly.

Note that this is not a critical issue or anything of that sort, since (1) the content types are already on any running system, that's why this was only failing on fresh test environments, (2) even if the dummy context were not loaded, on the next system reboot/deploy it would be loaded once the content types are loaded to the db.

klemmari1 (approved):
noice

PR #487, opened 2021-03-09
VEN-1191 | Change berth for a customer

klemmari1:
## Description :sparkles:

Add SwitchBerthMutation that allows to switch the berth from a berth lease to another berth that is not linked to a lease. The
berth lease must have a contract and the status must be paid in order to switch the berth for it.

In practice the old berth lease is terminated and a new one is created with the same information as the old one except with the berth changed to the new provided berth.

- Slight refactor of lease termination notification sending.

### Closes :no_good_woman:
**[VEN-1191](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1191)** 

### Automated tests :gear:️

```
pytest leases/tests/test_lease_mutations.py::test_switch_berth
pytest leases/tests/test_lease_mutations.py::test_switch_berth_with_switch_date
pytest leases/tests/test_lease_mutations.py::test_switch_berth_without_contract
pytest leases/tests/test_lease_mutations.py::test_switch_berth_with_berth_already_has_lease
pytest leases/tests/test_lease_mutations.py::test_switch_berth_non_paid_status
pytest leases/tests/test_lease_mutations.py::test_switch_berth_non_existing_berth_id
pytest leases/tests/test_lease_mutations.py::test_switch_berth_not_enough_permissions
```

### DIFF
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -1,7 +1,9 @@
 import threading
+from datetime import date, datetime
 
 import graphene
 from anymail.exceptions import AnymailError
+from dateutil.relativedelta import relativedelta
 from django.core.exceptions import ValidationError
 from django.db import transaction
 from django.utils.translation import gettext_lazy as _
@@ -12,6 +14,7 @@ from applications.new_schema import BerthApplicationNode, WinterStorageApplicati
 from berth_reservations.exceptions import VenepaikkaGraphQLError
 from contracts.services import get_contract_service
 from customers.models import CustomerProfile
+from leases.utils import exchange_berth_for_lease
 from payments.enums import OrderStatus
 from payments.models import BerthProduct, Order, WinterStorageProduct
 from resources.schema import BerthNode, WinterStoragePlaceNode, WinterStorageSectionNode
@@ -28,7 +31,7 @@ from ..enums import LeaseStatus
 from ..models import BerthLease, WinterStorageLease
 from ..services import BerthInvoicingService
 from ..stickers import get_next_sticker_number
-from ..utils import terminate_lease
+from ..utils import calculate_berth_lease_start_date, terminate_lease
 from .types import BerthLeaseNode, WinterStorageLeaseNode
 
 
@@ -213,6 +216,58 @@ class DeleteBerthLeaseMutation(graphene.ClientIDMutation):
         lease.delete()
 
         return DeleteBerthLeaseMutation()
+
+
+class SwitchBerthMutation(graphene.ClientIDMutation):
+    class Input:
+        old_lease_id = graphene.ID(required=True)
+        new_berth_id = graphene.ID(required=True)
+        switch_date = graphene.Date(
+            description="The date which will mark the end of the old lease and the start of the new lease."
+            "If none is provided, it will default to the day when the mutation is called"
+        )
+
+    old_berth_lease = graphene.Field(BerthLeaseNode)
+    new_berth_lease = graphene.Field(BerthLeaseNode)
+
+    @classmethod
+    def validate_switch_date(cls, switch_date):
+        if isinstance(switch_date, datetime):
+            switch_date = switch_date.date()
+        if switch_date < date.today() - relativedelta(months=6):
+            raise VenepaikkaGraphQLError(
+                _("Switch date is more than 6 months in the past")
+            )
+
+    @classmethod
+    def validate_old_lease(cls, old_lease):
+        if not old_lease.is_active:
+            raise VenepaikkaGraphQLError(_("Berth lease is not active"))
+
+    @classmethod
+    @add_permission_required(BerthLease)
+    @change_permission_required(BerthLease)
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, old_lease_id, new_berth_id, **input):
+        switch_date = input.get("switch_date") or calculate_berth_lease_start_date()
+        cls.validate_switch_date(switch_date)
+
+        old_lease = get_node_from_global_id(
+            info, old_lease_id, only_type=BerthLeaseNode, nullable=False,
+        )
+        cls.validate_old_lease(old_lease)
+
+        new_berth = get_node_from_global_id(
+            info, new_berth_id, only_type=BerthNode, nullable=False,
+        )
+
+        old_lease_comment = _("Lease terminated due to berth switch")
+        new_lease_comment = _("Lease created from a berth switch")
+        old_lease, new_lease = exchange_berth_for_lease(
+            old_lease, new_berth, switch_date, old_lease_comment, new_lease_comment,
+        )
+
+        return SwitchBerthMutation(old_berth_lease=old_lease, new_berth_lease=new_lease)
 
 
 class CreateWinterStorageLeaseMutation(graphene.ClientIDMutation):
@@ -575,6 +630,17 @@ class Mutation:
         "\n\nErrors:"
         "\n* A berth lease that is not `DRAFTED` anymore is passed"
         "\n* The passed lease ID doesn't exist"
+    )
+    switch_berth = SwitchBerthMutation.Field(
+        description="Switches the berth of a `BerthLease` object to the provided `Berth`."
+        "\n\n**Requires permissions** to edit and create leases."
+        "\n\nErrors:"
+        "\n* The passed lease or berth ID doesn't exist"
+        "\n* A berth lease that is not `PAID` is passed"
+        "\n* The the start date year does not match with the end date year"
+        "\n* A berth lease that has no contract is passed"
+        "\n* The passed berth already has a lease"
+        "\n* The provided `switch_date` is more than 6 months in the past"
     )
     terminate_berth_lease = TerminateBerthLeaseMutation.Field(
         description="Marks a `BerthLease` as terminated."
--- a/leases/utils.py
+++ b/leases/utils.py
@@ -11,12 +11,14 @@ from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 from django_ilmoitin.utils import send_notification
 
+from berth_reservations.exceptions import VenepaikkaGraphQLError
 from customers.services import ProfileService
 from leases.enums import LeaseStatus
 from utils.email import is_valid_email
 
 if TYPE_CHECKING:
-    from .models import BerthLease, WinterStorageLease
+    from leases.models import BerthLease, WinterStorageLease
+    from resources.models import Berth
 
 
 def calculate_season_start_date(lease_start: date = None) -> date:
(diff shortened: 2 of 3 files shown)


leases/schema/mutations.py
```diff
@@ -189,6 +189,69 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+    old_berth_lease = graphene.Field(BerthLeaseNode)
+    new_berth_lease = graphene.Field(BerthLeaseNode)
+
+    @classmethod
+    @add_permission_required(BerthLease)
+    @change_permission_required(BerthLease)
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, **input):
+        old_lease_id = input.pop("old_lease_id")
+        new_berth_id = input.pop("new_berth_id")
```
a-rmz, line 210:
```suggestion
    def mutate_and_get_payload(cls, root, info, old_lease_id, new_berth_id, **input):
```

leases/schema/mutations.py
```diff
@@ -189,6 +189,69 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+            "customer": old_lease.customer,
+            "end_date": old_lease.end_date,
+            "renew_automatically": old_lease.renew_automatically,
+            "status": old_lease.status,
+        }
+
+        old_lease = terminate_lease(
+            info, old_lease_id, BerthLeaseNode, end_date=switch_date,
+        )
+        old_lease.comment = "Lease is terminated due to berth switch"
```
a-rmz, line 234:
```suggestion
        old_lease.comment = f"\n{_('Lease is terminated due to berth switch')}"
```

leases/schema/mutations.py
```diff
@@ -189,6 +189,69 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        old_lease.save(update_fields=["comment"])
+
+        new_berth = get_node_from_global_id(
+            info, new_berth_id, only_type=BerthNode, nullable=False,
+        )
```
a-rmz, line 239: Can we get the berth before terminating the lease? in case it would fail or something, so the lease won't be terminated for nothing.

leases/schema/mutations.py
```diff
@@ -189,6 +189,69 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+        try:
+            kwargs.update({"berth": new_berth, "start_date": switch_date})
```
a-rmz, line 242: 1. Can you rename `kwargs` to (e.g.) `lease_args` to make it a bit clearer?
2. Can you define the whole `lease_args` here instead?
3. **OR** can you pass the params directly to the `create` instead of first creating a dict and then passing it?

a-rmz (commented):
One last thing, can you change `leases/schema/utils.py#46`, instead of
```python
    if profile_token:
        profile_service = ProfileService(profile_token=profile_token)
        profile = profile_service.get_profile(lease.customer.id)
        email = profile.email
    elif lease.application:
        email = lease.application.email
```
to
```python
    email = None

    if profile_token:
        profile_service = ProfileService(profile_token=profile_token)
        profile = profile_service.get_profile(lease.customer.id)
        email = profile.email
    
    if not email and lease.application:
        email = lease.application.email
```
So that, even if after calling profile service there's no email, it will still try to get the one from the application.

leases/schema/mutations.py
klemmari1: 1. Yep, let's rename.
2 / 3. I'm defining the kwargs before the termination to get the data of the old lease before the termination just in case. Now that I think about it, nothing is changed during the termination except for the status and comment fields so should be safe to pass them straight to the creation after the termination.

leases/schema/mutations.py
klemmari1: Yea. Now that I think about it, should we also put the whole process in a transaction?

leases/schema/mutations.py
klemmari1: So probably like this:
```
comment = _('Lease is terminated due to berth switch')
if len(old_lease.comment) > 0:
    old_lease.comment += f"\n{comment}"
else:
    old_lease.comment = comment
```

leases/schema/mutations.py
```diff
@@ -189,6 +189,71 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        old_lease = get_node_from_global_id(
+            info, old_lease_id, only_type=BerthLeaseNode, nullable=False,
+        )
+        new_berth = get_node_from_global_id(
+            info, new_berth_id, only_type=BerthNode, nullable=False,
+        )
+
+        end_date = old_lease.end_date
+        try:
+            contract = old_lease.contract
```
haavikko, line 220: What if it's a is_non_billable_customer who doesn't have a contract?

leases/schema/mutations.py
```diff
@@ -189,6 +189,71 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        try:
+            new_lease = BerthLease.objects.create(
+                berth=new_berth,
+                boat=old_lease.boat,
+                customer=old_lease.customer,
+                renew_automatically=old_lease.renew_automatically,
+                status=LeaseStatus.PAID,
+                start_date=switch_date,
+                end_date=end_date,
+                comment=_("Lease created from a berth switch"),
```
haavikko, line 244: Would be nice if comment contains the harbor name, pier id and number of the previous berth

leases/schema/mutations.py
```diff
@@ -189,6 +189,71 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+    old_berth_lease = graphene.Field(BerthLeaseNode)
+    new_berth_lease = graphene.Field(BerthLeaseNode)
+
+    @classmethod
+    @add_permission_required(BerthLease)
+    @change_permission_required(BerthLease)
+    @transaction.atomic
+    def mutate_and_get_payload(cls, root, info, old_lease_id, new_berth_id, **input):
+        switch_date = input.get("switch_date") or calculate_berth_lease_start_date()
```
haavikko, line 209: Should add some sanity checking for switch_date? I guess it should be possible to have switch_date in the past, in case admins just forgot to record a switch in the db(?). My guess is switch_date should be max 6 months in the past - maybe verify with Joonas?

leases/schema/mutations.py
```diff
@@ -189,6 +189,71 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+        old_lease = get_node_from_global_id(
+            info, old_lease_id, only_type=BerthLeaseNode, nullable=False,
+        )
```
haavikko, line 213: Should the old_lease be currently active? For example, prevent accidentally modifying leases from past seasons?
What about lease status? Is switching allowed for all lease statuses, including REFUSED/ERROR/TERMINATED/EXPIRED?

klemmari1: Hmm, yea, probably a good idea to add a check if the lease is active.

At the moment `leases.schema.utils.terminate_lease` checks that the lease status is PAID `leases/schema/utils.py:89`:
```
    if lease.status != LeaseStatus.PAID:
        raise VenepaikkaGraphQLError(_(f"Lease is not paid: {lease.status}"))
```

leases/schema/mutations.py
klemmari1: Good point! We have to address this case as well. So we should just ignore if the contract is missing?

leases/schema/mutations.py
klemmari1: Let's verify.

klemmari1: Verified that 6 months is a good threshold.

leases/schema/mutations.py
klemmari1: Added logic for leases without contracts.

leases/schema/mutations.py
klemmari1: Added check for active leases.

leases/schema/mutations.py
```diff
@@ -215,6 +217,58 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+    old_berth_lease = graphene.Field(BerthLeaseNode)
+    new_berth_lease = graphene.Field(BerthLeaseNode)
+
+    @classmethod
+    def validate_switch_date(cls, switch_date):
+        if isinstance(switch_date, datetime):
+            switch_date = switch_date.date()
+        if switch_date < date.today() - timedelta(weeks=4 * 6):
```
a-rmz, line 236:
```suggestion
        if switch_date < date.today() - relativedelta(months=6):
```

leases/utils.py
```diff
@@ -219,3 +221,60 @@ def terminate_lease(
...
+    new_lease_comment: str,
+) -> (BerthLease, BerthLease):
+    # Avoid circular imports
+    from leases.models import BerthLease
+
+    end_date = old_lease.end_date
+    contract = None
+    try:
+        contract = old_lease.contract
+    except old_lease.contract.DoesNotExist:
```
a-rmz, line 240: Will this work even when `contract` is `None`?

leases/utils.py
```diff
@@ -219,3 +221,60 @@ def terminate_lease(
...
+        pass
+
+    old_lease = terminate_lease(
+        lease=old_lease, end_date=switch_date, profile_token="", send_notice=False,
```
a-rmz, line 244: You can skip `profile_token` since it's a kwarg :)

```suggestion
        lease=old_lease, end_date=switch_date, send_notice=False,
```

leases/utils.py
klemmari1: There also a test for this ;) `leases.tests.test_lease_mutations.test_switch_berth_without_contract`

leases/utils.py
```diff
@@ -219,3 +221,60 @@ def terminate_lease(
...
+            status=LeaseStatus.PAID,
+            start_date=switch_date,
+            end_date=end_date,
+            comment=_(
+                f"{new_lease_comment}\n"
+                f"Previous berth info:\n"
+                f"Harbor name: {old_lease.berth.pier.harbor.name}\n"
+                f"Pier ID: {old_lease.berth.pier_id}\n"
+                f"Berth number: {old_lease.berth.number}\n"
+            ),
```
a-rmz, line 267: To translate only the relevant headers:
```suggestion
            comment=(
                f"{new_lease_comment}\n"
                f"{_('Previous berth info'):\n"
                f"{_('Harbor name')}: {old_lease.berth.pier.harbor.name}\n"
                f"{_('Pier ID')}: {old_lease.berth.pier.identifier}\n"
                f"{_('Berth number')}: {old_lease.berth.number}\n"
            ),
```

leases/utils.py
klemmari1: Although I should set the contract as `None` in that test..

a-rmz: Yeah, because if `old_lease.contract` is `None`, it would probably not have the `DoesNotExist` field 🤔

klemmari1: If `old_lease.contract` is `None`, and you try to access it `old_lease.contract`, it will throw an error:
```
leases.models.BerthLease.contract.RelatedObjectDoesNotExist: BerthLease has no contract.
```
which can be caught with `old_lease.contract.DoesNotExist`

klemmari1: Ah, I did not think this quite through. Yea, you are correct, it won't work if it is None.

Probably a better solution would be to replace this block with:
```
    if hasattr(old_lease, "contract"):
        contract = old_lease.contract
```

a-rmz: So you can't catch the exception if the variable does not have it :derp:

klemmari1: Yep. I was blinded by the passing test that was still working according to the old logic :grin: 

Okay, it's fixed now.

leases/utils.py
```diff
@@ -219,3 +221,58 @@ def terminate_lease(
...
+                f"{_('Harbor name')}: {old_lease.berth.pier.harbor.name}\n"
+                f"{_('Pier ID')}: {old_lease.berth.pier.identifier}\n"
+                f"{_('Berth number')}: {old_lease.berth.number}\n"
+            ),
+        )
+
+        if contract:
+            contract.lease = new_lease
+            contract.save()
+            old_lease.refresh_from_db()
```
a-rmz, line 271: _Maybe_ also refresh `new_lease`? So when you return it, it comes with the associated lease already.

klemmari1: Added.

a-rmz (approved):

PR #491, opened 2021-03-11
VEN-1202 | Base offer models

a-rmz:
## Description :sparkles:
* Add the base models for berth switch offers

### Related :handshake:
**[VEN-1202](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1202):** Mutation to send switch offer 

## Screenshots :camera_flash:
image


payments/models.py
```diff
@@ -957,3 +962,56 @@ class OrderToken(UUIDModel, TimeStampedModel):
...
+        self.customer_phone = profile.phone
+
+        self.save()
+
+
+class BerthSwitchOffer(AbstractOffer):
+    application = models.ForeignKey(
+        BerthApplication, related_name="switch_offers", on_delete=models.CASCADE
+    )
+    lease = models.ForeignKey(
```
haavikko, line 1003: Could add validation that this lease must be for the season that immediately precedes the season defined in the application?
Why linking to lease instead of Order?

klemmari1:
I was thinking, do we also need a date field for when the switch is happening?

payments/models.py
a-rmz: > _Could add validation that this lease must be for the season that immediately precedes the season defined in the application?_

Good catch 👍 

> _Why linking to lease instead of Order?_

It's a link to the _old_ lease, since the lease is kinda the "source of truth", meaning that even if the invoice is cancelled or anything, as long as the lease is not terminated, it's what determines whether the customer has the place or not.

a-rmz:
> _I was thinking, do we also need a date field for when the switch is happening?_

@klemmari1 I think we can use the same logic as we have currently. Since, in the end, the switch offers will be dealt before the season starts and they still have to be "confirmed" by paying the invoice, we can rely on the current date logic for it.

haavikko (approved):
If you have the time to make the suggested changes please do, but those can be done as part of other tickets too.

PR #493, opened 2021-03-15
VEN-1202 | Berth switch offers (django admin)

a-rmz:
## Description :sparkles:
*  Add `BerthSwitchOfferLogEntry` model for tracking status changes
* Fix switch offer factory to always have a paid lease
* Add berth switch offer models to admin

### Related :handshake:
#491 

## Screenshots :camera_flash:
image


customers/admin.py
```diff
@@ -63,6 +64,32 @@ class CustomerProfileAdmin(admin.ModelAdmin):
...
+    search_fields = ("id",)
+    list_display = ("id", "berth_leases", "winter_leases", "orders")
+
+    def get_queryset(self, request):
+        return (
+            super()
+            .get_queryset(request)
+            .annotate(
+                berth_lease_count=Count("berth_leases"),
+                winter_leases=Count("winter_storage_leases"),
```
klemmari1, line 76: `winter_lease_count`

customers/admin.py
```diff
@@ -63,6 +64,32 @@ class CustomerProfileAdmin(admin.ModelAdmin):
...
+                order_count=Count("orders"),
+            )
+        )
+
+    def berth_leases(self, obj):
+        return obj.berth_lease_count
+
+    def winter_leases(self, obj):
+        return obj.berth_lease_count
```
klemmari1, line 85: `return obj.winter_lease_count`

a-rmz: :derp:

payments/admin.py
```diff
@@ -274,11 +292,35 @@ def order_id(self, obj):
...
+        "application__id",
+        "berth__id",
+        "berth__number",
+        "berth__pier__harbor__translations__name",
+        "lease__id",
+        "customer_first_name",
+        "customer_last_name",
+    )
+    autocomplete_fields = ("customer", "application", "lease", "berth")
+
```
haavikko, line 316: This might be useful:
`    date_hierarchy = "due_date"
`

payments/models.py
```diff
@@ -1109,9 +1109,78 @@ def clean(self):
...
-        ) - relativedelta(years=1):
+        if (
+            self.lease.start_date.year
+            != (calculate_season_start_date(today()) - relativedelta(years=1)).year
+        ):
             raise ValidationError(
                 _("The exchanged lease has to be from the last season")
             )
+
+    def set_status(self, new_status: OfferStatus, comment: str = None) -> None:
```
haavikko, line 1120: +1 Actually I'll edit the offer expiration cronjob to use this instead of settings status directly

payments/tests/factories.py
```diff
@@ -197,6 +198,7 @@ class BerthSwitchOfferFactory(AbstractOfferFactory):
     lease = factory.SubFactory(
         BerthLeaseFactory,
         customer=factory.SelfAttribute("..customer"),
+        status=LeaseStatus.PAID,
```
haavikko, line 201: Was just fixing this same issue :+1:

payments/models.py
```diff
@@ -1109,9 +1109,78 @@ def clean(self):
             raise ValidationError(_("The application has to be a switch application"))
 
         # Validate that the lease can only be from the immediate last season
-        if self.lease.start_date.year != calculate_season_start_date(
-            today()
-        ) - relativedelta(years=1):
+        if (
+            self.lease.start_date.year
+            != (calculate_season_start_date(today()) - relativedelta(years=1)).year
```
haavikko, line 1114: I have the same fix in offer expiration branch :-) But I started to think more about this. What if the admin creates the switch offer in December? In that case, self.lease.start_date.year is the same as calculate_season_start_date(today()).year?

a-rmz: So you mean if the offer is created _after_ the summer season has ended?

klemmari1: I ran into this problem as well. I think if the offer is created after the season has ended, it will always return the next season.

haavikko: Yesterday I noticed I've been misunderstanding this. Actually, the customer's old lease and the new lease should be for the same season, right?

For example now in March 2021, a customer would need to have a paid lease for the summer season 2021 in order to be eligible for making a berth switch. So if I understand correctly, the process most commonly goes like this:
1. Customer creates a switch application
2. Customer receives invoice for the next summer season and pays it
3. Admin creates switch offer, linking it to the lease created in (2). Because customer has already paid for the season, no new invoice is created.
Steps 1 and 2 can actually be done in any order.

So the comment "Validate that the lease can only be from the immediate last season" should be changed to something like "Validate that the lease can only be from the upcoming/ongoing season" and the relativedelta(years=1) removed from the condition.

I think Joonas commented that switch offers might be created even during the season.

haavikko (changes requested):

payments/models.py
klemmari1: I was a bit confused about this logic as well as to why it has to be last season :D How I understood it, the lease should be for the next / ongoing season and should be paid. So my logic would be what @haavikko said but 1. and 2. the other way around since I think the lease needs to be paid in order to create a switch application?

a-rmz: Hmmm what I had understood is that switch applications are like normal applications but instead of them being for a new customer, they are for customers who already have a berth and want to switch it for next season. So they would fill the application and it would be handled before the new season. Let's figure this out asap.

haavikko: See comment from Joonas in https://helsinkicity.slack.com/archives/CBA1QHTHN/p1615898900047800

How I understand it: At the time the customer creates a switch application, the ustomer needs to have a berth lease that is paid. Customer is able to create a switch application at any time of year. If the customer creates the application in, let's say November 2021, then the customer must have a paid berth lease for the 2021 season, but need not yet have paid the invoice for 2022 season.

But in the example above, the admins would only create a berth switch offer for the customer only after the customer has paid their invoice for season 2022.

haavikko (approved):

PR #497, opened 2021-03-18
Tech | Add a mapping between old and new harbors

a-rmz:
## Description :sparkles:
The old harbors app was the first version of the whole system. It had a
simple information architecture which covered the MVP needs on the
moment, necessary only to release the applications.

When applications was released and the development started to shift
towards the admin UI, a new, more organised, clearly divided
categorisation came with the resources app. The "harbors" on the old
harbors app were now split into Piers and Harbors.

Some of the old harbors corresponded directly to a new harbor on a 1:1
mapping, some didn't but they mapped to a new pier.

Adding this mapping allows to connect the old harbors (used for
applications) with the new ones (used everywhere else), so some
functionality can be unblocked, like using the berth switch
applications.

There are three old harbors which don't have a direct mapping.
```
Pajalahden venesatama (rantamuuri)
Merihaan venesatama (Hakaniemenranta 30) [already deleted from the “new” harbours, so probably won't harm]
Hietalahdenallas (Jätkäsaaren puoli) / Venesatama
````
Joonas is already aware and will check with the admins.

### CODE REVIEW

File: resources/models.py [hunk 1]
```diff
@@ -495,6 +495,13 @@ class Pier(AbstractAreaSection):
         verbose_name=_("price tier"),
         default=PriceTier.TIER_1,
     )
+    harbors_harbor = models.ForeignKey(
```
haavikko, line 498: Too bad the data doesn't allow to make this too an one-to-one mapping :-(

harbors/management/commands/harbors_connect_to_resources.py
```diff
@@ -0,0 +1,119 @@
...
+            harbor.resources_harbor = None
+            harbor.save()
+            cleaned_harbors += 1
+        self.stdout.write("Harbors cleaned: " + str(cleaned_harbors))
+
+        for pier in Pier.objects.filter(harbors_harbor__isnull=False):
+            pier.harbors_harbor = None
+            pier.save()
+            cleaned_piers += 1
+        self.stdout.write("Piers cleaned: " + str(cleaned_piers))
```
haavikko, line 88: +1 for cleanup!

haavikko (approved):

klemmari1 (approved):
LGTM!

harbors/management/commands/harbors_connect_to_resources.py
a-rmz: Only a two harbors are mapped to several piers, otherwise it could be a 1:1
image

resources/models.py
a-rmz: Only a two harbors are mapped to several piers, otherwise it could be a 1:1
image

a-rmz:
Tested already on the review env
image

PR #507, opened 2021-03-23
VEN-1202 | Create berth switch mutation

a-rmz:
## Description :sparkles:
- Update `BerthSwitchOffer` models
  - Rename `OfferStatus.SENT -> OFFERED` for consistency with other statuses
- Make django admin customer fields use select2
- Add BerthSwitchOffer GQL types and queries
- Add `createBerthSwitchMutation`

### Closes :no_good_woman:
**[VEN-1202](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1202):** Mutation to create switch offer

### Related :handshake:
#491 
#493 

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_queries_berth_switch_offers.py
$ pytest payments/tests/test_payments_mutations_create_berth_switch_offer.py
payments/tests/test_payments_models.py::test_offer_without_due_date_not_drafted
```

**Application with switch**
1. Create an application with a berth switch, the switch info should be
```
Harbor: Eläintarhanlahden venesatama (id: 4)
Pier: -
Berth number: 1
```
2. Handle the application (connect it to a customer)
3. Create a lease for the current season for the same customer and same harbor (`2cf50e2f-e101-48bd-8dfd-cf0d7501b654`), pier (`28b22309-7e06-42c9-8464-ce4ff8b88cc4`) and berth (`acdca8a7-7875-4a8a-be56-d1c6d206b4b2`)
3. Mark that lease as paid
4. Run the create mutation
```graphql
mutation CREATE_BERTH_SWITCH_OFFER_MUTATION($input: CreateBerthSwitchOfferMutationInput!) {
  createBerthSwitchOffer(input: $input) {
    berthSwitchOffer {
      id
      status
      dueDate
      customer {
        id
      }
      application {
        id
      }
      lease {
        id
      }
      berth {
        id
      }
    }
  }
}
```

```json
{
  "input": {
    "applicationId": "SWITCH APPLICATION ID",
    "newBerthId": "ID OF THE NEW BERTH"
  }
}
```

**Manually specifying the lease**
Same process as the previous, but instead of (3) creating a lease for the exact same berth as the switch, change it for another berth.

On the input, add the `oldLeaseId`:
```json
{
  "input": {
    "applicationId": "SWITCH APPLICATION ID",
    "newBerthId": "ID OF THE NEW BERTH",
    "oldLeaseId": "ID OF THE OLD LEASE"
  }
}
```

## Screenshots :camera_flash:
image

## Additional notes 📝 
Before doing any manual tests, you have to run the following commands:
```shell
python manage.py migrate
python manage.py loaddata helsinki-harbor-resources.json
python manage.py loaddata helsinki-harbors.json
python manage.py harbors_connect_to_resources
```


payments/schema/mutations.py
```diff
@@ -741,6 +747,118 @@ def mutate_and_get_payload(cls, root, info, order_id, **input):
...
+
+        old_berth = Berth.objects.get(berth_filters)
+
+        # Find the lease on the current season
+        return BerthLease.objects.get(
+            customer=application.customer,
+            berth=old_berth,
+            status=LeaseStatus.PAID,
+            start_date=calculate_season_start_date(),
+            end_date=calculate_season_end_date(),
```
haavikko, line 812: Thinking of a special case. Are berth switch offers ever created in the time range from September 14th to December 31th?
Let's say, for some reason, that the admin wants to create a switch offer sometime in December 2021.
In that case, the 2021 season has already ended and in order to proceed with the switch, the customer needs to have a paid lease for the 2022 season. but calculate_season_(start|end)_date would return dates from 2021. So we'd need here functions like "calculate_current_or_upcoming_season_(start|end)_date"?

Actually I'm not entirely sure if this case would ever occur and if I understand the process correctly, but maybe the rules should be:
* If switch offer is created while season is active, the user gets the new berth after accepting the offer. Customer must have paid lease for the ongoing season.
* If switch offer is created at any other time, then the user gets the new berth when the next season starts. Customer must have paid lease for the next season.

haavikko: Joonas and Kalle commented in slack that it's very unlikely that switch offers would be created in days between season's end and the start of the next year, so this question is kind of moot. Maybe just prevent creating the switch offer if the lease.end_date < today()?

payments/schema/mutations.py
```diff
@@ -741,6 +747,118 @@ def mutate_and_get_payload(cls, root, info, order_id, **input):
...
+            piers = switch.harbor.resources_pier.all()
+
+            # Try to match the one with the identifier provided by the customer
+            pier = piers.filter(identifier__iexact=switch.pier).first()
+
+            # If no pier is found but the harbor only has one pier associated,
+            # we can assume that the old berth has to be on that one pier
+            if not pier and piers.count() == 1:
+                pier = piers.first()
+
```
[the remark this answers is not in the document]
a-rmz: It will be handled/raised when you try to find a berth that has no pier, so it will raise the `NO_LEASE` error and the frontend will then explicitly pass the lease id>

payments/schema/mutations.py
a-rmz: > but calculate_season_(start|end)date would return dates from 2021. So we'd need here functions like "calculate_current_or_upcoming_season(start|end)_date"?

The calculate season start/end check if the season has already ended (e.g. between September and December). If so, they return the dates for the next season.
```python
if today >= date(day=14, month=9, year=today.year):
    return default.replace(year=today.year + 1)
```

klemmari1:
The description has:
>    - Make due_date optional
Is this included in this PR?

payments/schema/mutations.py
haavikko: OK! I got confused when I tried:
`
In [5]: calculate_season_start_date(datetime.date(2020,11,1))
Out[5]: datetime.date(2020, 6, 10)
`
But now I see that calling the function without argument indeed fixes it.

a-rmz:
>Is that included in this PR?

Good point, that was solved on a previous PR and it's merged already

payments/schema/mutations.py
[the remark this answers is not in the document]
haavikko: ok

haavikko (approved):
Looks good to me!

klemmari1 (approved):
LGTM!

noice2

PR #508, opened 2021-03-25
VEN-1251: do not filter by berth length when querying berths for an application

haavikko:
Admins want to see also berths that are too short. Highlighting will be done in UI.

### Closes :no_good_woman:
**[VEN-1251](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1251):**

### DIFF
--- a/resources/schema/utils.py
+++ b/resources/schema/utils.py
@@ -11,8 +11,8 @@ from ..models import Berth, BerthType, Pier
 
 
 def resolve_piers(info, **kwargs):
-    min_width = kwargs.get("min_berth_width", 0)
-    min_length = kwargs.get("min_berth_length", 0)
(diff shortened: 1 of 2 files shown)


resources/schema/utils.py
```diff
@@ -42,7 +42,9 @@ def resolve_piers(info, **kwargs):
             )
 
             min_width = application.boat_width
-            min_length = application.boat_length
+
+            # VEN-1251: berths that are too short need to be shown in the UI
+            min_length = None
```
a-rmz, line 47: Is it necessary to explicitly set it as `None` or would the declaration `min_length = kwargs.get("min_berth_length")` be enough?

haavikko: hmm, that would work too. fixed.

a-rmz (approved):

PR #511, opened 2021-03-29
VEN-1229 | Remove old Harbors app

a-rmz:
## Description :sparkles:
- Remove all occurrences of `harbors`
- Move all the API contents from `/graphql_v2` to `/graphql`
- Move the submit application queries to the new schema
- Make sure the applications use the resources models instead of harbors
- Remove applications.schema and replace it with applications.new_schema
- Move all the queries mutation from berth_reservations.schema to the unified schema
  - payments
  - applications
  - contracts
- Remove the `old_api_client` from tests
- The `HarborChoiceType`/`WinterStorageAreaChoiceType` had the harbor/area fields PLUS a harborName/areaName and an extra `winterStorageSectionIds`. Both have been removed since now they're accessible through the `properties` field
- Change the `BerthSwitch` model to require a `berth` instead of a `harbor` with text pier id/berth number
  - Adapt the queries and mutations for it

### Closes :no_good_woman:
**[VEN-1229](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1299):** Remove harbors app

### Related 🤝 
https://github.com/City-of-Helsinki/berth-federation-gateway/pull/31

### Automated tests :gear:️
```shell
$ pytest
```

To remove everything, first remove all the harbor-related tables and migrations, then run the migrations
```
$ python manage.py shell < ./scripts/delete_incomplete_switch_applications.py
$ python manage.py remove_harbor_models
$ python manage.py migrate
```

The commands will be run manually in staging/prod to make sure that they are executed correctly.

Final note: apologies for the single commit, but there's too many changes and very little return on investing more time on splitting into smaller commits.


applications/migrations/0029_replace_harborchoice_harbors_with_resources.py
```diff
@@ -0,0 +1,141 @@
+# Generated by Django 3.1 on 2021-03-25 08:10
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def copy_harbor_choices(apps, schema_editor):
```
klemmari1, line 7: I'm not super familiar with the M2M relationships using a through model. Can you quickly explain why do we need this tmp model to copy and restore the data? It is not possible to do something like:

```
    for choice in HarborChoice.objects.all():
        choice.harbor=Harbor.objects.filter(servicemap_id=choice.harbor.servicemap_id).first()
        choice.save(update_fields=["harbor"])
```

applications/migrations/0029_replace_harborchoice_harbors_with_resources.py
```diff
@@ -0,0 +1,141 @@
...
+    TmpHarborChoice = apps.get_model("applications", "TmpHarborChoice")
+    HarborChoice = apps.get_model("applications", "HarborChoice")
+
+    for choice in HarborChoice.objects.all():
+        choice.harbor = TmpHarborChoice.objects.get(id=choice.id).harbor
+        choice.save()
+
+
+def undo_restore_harbor_choices(apps, schema_editor):
+    TmpHarborChoice = apps.get_model("applications", "TmpHarborChoice")
```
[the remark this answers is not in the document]
a-rmz: Right, this is a leftover from some previous testing, it should be gone 🔥

applications/migrations/0029_replace_harborchoice_harbors_with_resources.py
a-rmz: Because this is done before we have changed the `harbor_id` to the new harbors. So they have conflicting types, the "current" `harbor_id: int` and the "new" `harbor_id: UUID`.

By adding this temporary model, we point the choices to the correct `resources.Harbor`, then we change the actual model accordingly now that the data has been backed up in the temp model, so the table will have a type `uuid` instead of `int`. Then, when the model has been corrected, we migrate the data back and get rid of the old model.

klemmari1: Ah yep, thanks! Makes sense now that I look at it again :grin:

klemmari1:
References to `helsinki-harbors.json` and `helsinki-winter-areas.json` should be removed.

haavikko (approved):
From old slack post:

the data migration is quite complex and it doesn't seem reversible(?), how do we prepare for prod deployment? For example, arrange something like this with devops:

In advance, announce downtime for maintenance
Juho takes admin UI and customer UI offline (inaccessible to admins/customers)
Juho takes backup of the production database
We deploy the migrations and monitor for errors
Juho brings admin UI up, Joonas logs in and tests admin UI
If something seems wrong, database is restored from backup

applications/migrations/0033_replace_berthswitch_harbor_for_berth.py
```diff
@@ -0,0 +1,59 @@
...
+                # Assign the correct berth to the switch
+                berth = Berth.objects.get(
+                    harbor=switch.harbor,
+                    harbor__pier__identifier__iexact=switch.pier,
+                    number=switch.berth_number,
+                )
+                switch.berth = berth
+                switch.save()
+            except Berth.DoesNotExist as e:
+                print(f"{switch.harbor.id},{switch.pier},{switch.berth_number}")
```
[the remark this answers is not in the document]
a-rmz: I think we can see them from the Grafana logs

PR #512, opened 2021-03-29
 VEN-1201 | Accept/reject switch offer

klemmari1:
## Description :sparkles:

Add a mutation for accepting & rejecting offers. The accepting uses the logic of berth switching from #487.

Rejecting will mark the offer as REJECTED and nothing else is changed.

The mutation will return the offer and the new lease.

### Closes :no_good_woman:
**[VEN-1201](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1201)** 

### Automated tests :gear:️

```
pytest test_payments_mutations_accept_berth_switch_offer.py
```

### TODO:
- Restore the permission check for `BerthSwitchOfferNode` when the customer login is sorted out.

### DIFF
--- a/payments/schema/mutations.py
+++ b/payments/schema/mutations.py
@@ -22,7 +22,11 @@ from customers.services import ProfileService
 from leases.enums import LeaseStatus
 from leases.models import BerthLease, WinterStorageLease
 from leases.schema import BerthLeaseNode, WinterStorageLeaseNode
-from leases.utils import calculate_season_end_date, calculate_season_start_date
+from leases.utils import (
+    calculate_season_end_date,
+    calculate_season_start_date,
+    exchange_berth_for_lease,
+)
 from resources.models import Berth
 from resources.schema import BerthNode, WinterStorageAreaNode
 from users.decorators import (
@@ -34,7 +38,7 @@ from users.decorators import (
 from utils.relay import get_node_from_global_id
 from utils.schema import update_object
 
-from ..enums import OrderStatus, OrderType, ProductServiceType
+from ..enums import OfferStatus, OrderStatus, OrderType, ProductServiceType
 from ..exceptions import VenepaikkaPaymentError
 from ..models import (
     AdditionalProduct,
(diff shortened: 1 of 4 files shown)


payments/schema/mutations.py
```diff
@@ -859,6 +863,43 @@ def mutate_and_get_payload(cls, root, info, application_id, new_berth_id, **inpu
         return CreateBerthSwitchOfferMutation(berth_switch_offer=offer)
 
 
+class AcceptBerthSwitchOfferMutation(graphene.ClientIDMutation):
+    class Input:
+        offer_id = graphene.ID(required=True)
+        is_accepted = graphene.Boolean(required=True)
+
+    offer = graphene.Field(BerthSwitchOfferNode)
```
a-rmz, line 871: Not 100% about this 🤔 since the mutation is being executed by the customer UI, it wouldn't expect any response. And if it did, I think it should also return the new lease 🤔 🤔 🤔 🤔

a-rmz: Ok, since now we'll have "both" APIs living under tha same roof, let's keep the offer and also add the new lease, and then let the UI decide what information they need back.

klemmari1: Yep, sounds good to me.

klemmari1: New lease added to response.

haavikko: I checked Lease and Offer, to me it seems they don't have any fields that the customer isn't allowed to see (it's their lease and if the customer made a GDPR request, the city would need to provide the data anyway I think).

payments/tests/test_payments_mutations_accept_berth_switch_offer.py
```diff
@@ -0,0 +1,185 @@
...
+            }
+            status
+            dueDate
+            customerFirstName
+            customerLastName
+            customerEmail
+            customerPhone
+        }
+        newLease {
+            id
```
haavikko, line 54: validate the status of newLease too?

payments/tests/test_payments_mutations_accept_berth_switch_offer.py
```diff
@@ -0,0 +1,185 @@
...
+    assert executed["data"]["acceptBerthSwitchOffer"]["newLease"] == {
+        "id": to_global_id(
+            BerthLeaseNode,
+            BerthLease.objects.exclude(id=berth_switch_offer.lease_id).first().id,
+        ),
+    }
+
+
+@pytest.mark.parametrize(
+    "api_client", ["berth_services"], indirect=True,
```
haavikko, line 114: Customers would not have berth_services permissions in a real case? Are the permission issues in scope of this ticket?

payments/tests/test_payments_mutations_accept_berth_switch_offer.py
klemmari1: Ah yep, forgot. Added.

payments/tests/test_payments_mutations_accept_berth_switch_offer.py
klemmari1: The permissions are not super clear to me what is used for what. Are they documented somewhere?

a-rmz: Since it's a mutation that will be executed by customers which have not necesarily logged in, it should be a public query, so the base api client should work.

payments/schema/types.py
```diff
@@ -269,11 +269,6 @@ class Meta:
         interfaces = (graphene.relay.Node,)
         connection_class = CountConnection
 
-    @classmethod
-    @view_permission_required(BerthSwitchOffer)
-    def get_queryset(cls, queryset, info):
-        return super().get_queryset(queryset, info)
-
```
a-rmz, line 276: Now that we don't return the offer, we should add this back

klemmari1: It still needs to be removed so that the customers can access (`payments/schema/mutations.py:874`):
```
        offer = get_node_from_global_id(
            info, offer_id, only_type=BerthSwitchOfferNode, nullable=False,
        )
 ```

a-rmz: Does `get_node_from_global_id` check for permissions as well? I think it doesn't

klemmari1: At least when testing (for example `payments.tests.test_payments_mutations_accept_berth_switch_offer.test_accept_offer`) it gives the error "You do not have permission to perform this action" when getting the instance in `utils/relay.py:20`. I think this is because it uses `BerthSwitchOfferNode.get_node` which uses `get_queryset` to fetch the model instance and that restricts the access.

a-rmz: Hmmmm makes sense... let's keep a TODO that we have to enable permission checking for the node once we get the customer login sorted out.

klemmari1: Sounds good. I'll add the TODO to BerthSwitchOfferNode.

a-rmz (approved):

PR #516, opened 2021-04-01
VEN-1216 | Update user profile query permissions and add offers to profile query

klemmari1:
## Description :sparkles:

- Allow customers to view their own Orders and Offers.
- Add BerthSwitchOffer to be queried through ProfileNode.

### Closes :no_good_woman:
**[VEN-1216](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1216): Query for the customer personal information** 

### Automated tests :gear:️

```
pytest customers/tests/test_customers_queries.py::test_query_berth_profile_self_user
```

### DIFF
--- a/customers/schema/types.py
+++ b/customers/schema/types.py
@@ -9,8 +9,9 @@ from graphql_jwt.decorators import login_required
 from applications.models import BerthApplication
 from applications.new_schema import BerthApplicationNode, WinterStorageApplicationNode
 from applications.new_schema.types import WinterStorageApplicationFilter
-from leases.models import BerthLease
+from leases.models import BerthLease, WinterStorageLease
 from leases.schema import BerthLeaseNode, WinterStorageLeaseNode
+from payments.models import BerthSwitchOffer, Order
 from utils.relay import get_node_from_global_id, return_node_if_user_has_permissions
 from utils.schema import CountConnection
 
@@ -100,6 +101,7 @@ class ProfileNode(DjangoObjectType):
             "winter_storage_applications",
             "winter_storage_leases",
             "orders",
+            "offers",
         )
         interfaces = (relay.Node,)
         connection_class = CountConnection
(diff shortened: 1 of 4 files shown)


customers/schema/types.py
```diff
@@ -134,6 +136,7 @@ class Meta:
     )
     winter_storage_leases = DjangoConnectionField(WinterStorageLeaseNode)
     orders = DjangoConnectionField("payments.schema.OrderNode")
+    offers = DjangoFilterConnectionField("payments.schema.BerthSwitchOfferNode")
```
a-rmz, line 139: Do we actually need the filter connection? 🤔

klemmari1: Whoops, that slipped there :grin:

customers/schema/types.py
```diff
@@ -145,13 +148,25 @@ def resolve_winter_storage_applications(self, info, **kwargs):
...
         profile = get_node_from_global_id(info, self.id, only_type=ProfileNode)
         return return_node_if_user_has_permissions(
-            profile, info.context.user, CustomerProfile, BerthApplication, BerthLease
+            profile,
+            info.context.user,
+            CustomerProfile,
+            BerthApplication,
+            BerthLease,
+            Order,
+            BerthSwitchOffer,
```
a-rmz, line 157: We have to add winter leases, since they can also be queried through the node.

a-rmz (commented):
We should add (still figuring out how) an `api_client` for a customer, so we can test specifically that the customer can _only_ access their own stuff and nothing else.

customers/schema/types.py
klemmari1: and I think boats as well

klemmari1: Added

payments/schema/types.py
```diff
@@ -225,9 +228,12 @@ class Meta:
...
 
     @classmethod
-    @view_permission_required(Order)
-    def get_queryset(cls, queryset, info):
-        return super().get_queryset(queryset, info)
+    @login_required
+    def get_node(cls, info, id):
+        node = super().get_node(info, id)
+        return return_node_if_user_has_permissions(
+            node, info.context.user, Order, CustomerProfile
```
haavikko, line 235: Do we need to check lease permissions too?

klemmari1: Added

a-rmz (approved):

PR #519, opened 2021-04-07
VEN-1234 | Differentiate between sent and unsent invoices

klemmari1:
## Description :sparkles:

- Divided the waiting status of invoices into two separate statuses: WAITING -> (DRAFTED, OFFERED)
- Modify Order due_date to be optional and make sure that it is set when sending the order (invoice) to a customer (setting the order status as OFFERED)

### Closes :no_good_woman:
**[VEN-1234](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1234): Differentiate between sent/unsent invoices**

### Automated tests :gear:️

Testing the due_date change:
```
pytest payments.tests.test_payments_models.py::test_order_set_due_date_after_status_change
pytest payments.tests.test_payments_models.py::test_order_due_date_has_to_be_set_for_order_in_offered_status
```


payments/models.py
```diff
@@ -587,6 +585,9 @@ def clean(self):
             # if due_date change is allowed
             self._check_due_date(old_instance)
 
+        if self.status == OrderStatus.OFFERED and not self.due_date:
```
haavikko, line 588: Do we want to allow Orders to not have due_date if they are in ERROR? The only time an order is set to error state is when send the invoice fails, so it should have a due date at the time?

payments/models.py
```diff
@@ -720,7 +721,13 @@ def set_status(self, new_status: OrderStatus, comment: str = None) -> None:
             return
 
         valid_status_changes = {
-            OrderStatus.WAITING: (
+            OrderStatus.DRAFTED: (
+                OrderStatus.OFFERED,
+                OrderStatus.PAID_MANUALLY,
+                OrderStatus.ERROR,
```
haavikko, line 727: Is the transition DRAFTED -> ERROR ever used? If it's not needed it would be nice to remove, then we'd not need to worry about orders in ERROR state that don't have a due_date

payments/models.py
```diff
@@ -720,7 +721,13 @@ def set_status(self, new_status: OrderStatus, comment: str = None) -> None:
...
+                OrderStatus.CANCELLED,
```
haavikko, line 728: Also, do we need this transition actually? If it's not used in the application flow let's remove it, allowing it later is easy if needed. If the order has not yet been sent to the customer, maybe the correct thing to do would be to remove it completely from the database?

payments/models.py
klemmari1: True. I left it here in case it would be needed at some point but now that I think about it, it's best to be left out.

payments/models.py
klemmari1: Good point! I'll remove this as well.

payments/models.py
klemmari1: Now that I think about it, this is probably needed in the case when the email is missing for example. So:
`leases/services/invoice.py:192`

payments/models.py
klemmari1: I don't think the Orders would need a due_date in the case of DRAFTED -> ERROR state, but they would get the due_date when the order is set from ERROR -> OFFERED.

payments/models.py
klemmari1: I think I will change the logic so that when sending the invoices in `leases/services/invoice.py:286` the order would be in DRAFTED state and it would be transitioned to OFFERED in `payments.utils.approve_order`

a-rmz: The lease is already being set to offered
```python
# leases/services/invoice.py:345
if order.lease and order.order_type == OrderType.LEASE_ORDER:
    # Update lease status
    order.lease.status = LeaseStatus.OFFERED
    order.lease.save()
```

klemmari1: But don't we want to set the Order as OFFERED there as well?

a-rmz: AHHHH, I was looking for the place where we were setting the order to offered, but then realised that this is what this PR is for 🤦 

Yes, let's do it here 👍

payments/models.py
```diff
@@ -731,11 +737,11 @@ def set_status(self, new_status: OrderStatus, comment: str = None) -> None:
             OrderStatus.PAID: (OrderStatus.REFUNDED,),
             OrderStatus.PAID_MANUALLY: (OrderStatus.REFUNDED,),
             OrderStatus.ERROR: (
```
a-rmz, line 739: Let's add also `DRAFTED` here, following the comment on `leases/services/invoice.py`

payments/models.py
```diff
@@ -807,6 +816,10 @@ def update_sticker_number_if_needed(self):
             self.lease.sticker_number = sticker_number
             self.lease.save(update_fields=["sticker_number"])
 
+    def set_due_date(self):
```
a-rmz, line 819: Maybe?
```suggestion
    def set_default_due_date(self):
```

payments/notifications/dummy_context.py
```diff
@@ -128,7 +128,7 @@ def _get_cancelled_order_context(subject: str = "Order cancelled"):
                 dict,
                 FACTORY_CLASS=OrderFactory,
                 customer=customer,
-                status=OrderStatus.WAITING,
+                status=OrderStatus.CANCELLED,
```
a-rmz, line 131: Nice catch 😅

payments/schema/mutations.py
```diff
@@ -487,7 +487,7 @@ def mutate_and_get_payload(cls, root, info, **input):
         order_number = input.get("order_number", None)
         try:
             order = Order.objects.get(order_number=order_number)
-            if order.status != OrderStatus.WAITING:
+            if order.status not in OrderStatus.get_waiting_statuses():
```
a-rmz, line 490: Should we only allow cancelling `OFFERED` orders? Wouldn't make much sense to cancel only drafted ones.

payments/utils.py
```diff
@@ -414,12 +417,12 @@ def send_refund_notice(order):
...
     if order.status == OrderStatus.ERROR or order.lease.status == LeaseStatus.ERROR:
         order.set_status(
-            OrderStatus.WAITING,
+            OrderStatus.OFFERED,
             comment=f"{today()}: {_('Cleanup the invoice to attempt resending')}\n",
         )
     elif order.status == OrderStatus.CANCELLED:
         order.set_status(
-            OrderStatus.WAITING,
+            OrderStatus.OFFERED,
```
a-rmz, line 425: Same as the comment on `leases/services/invoice.py`, let's set them to `DRAFTED` (and maybe even remove the due date?) and let the actual sending handle the status change to `OFFERED`.

a-rmz (changes requested):

payments/models.py
klemmari1: Alrighty. So do I remove ` and not self.due_date` so that it gets a new due date even though it has one?

klemmari1: Or perhaps the best solution would be to clean the due date in `leases.services.invoice.BaseInvoicingService.fail_order`

a-rmz: Sounds good 👍

leases/services/invoice.py
```diff
@@ -308,7 +308,7 @@ def resend_failed_invoices(self, profiles: dict) -> None:
 
         for order in orders:
             order.set_status(
-                OrderStatus.WAITING,
+                OrderStatus.OFFERED,
```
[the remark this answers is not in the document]
klemmari1: The lease has to be in OFFERED status for the resending though `payments/utils.py:431` and `payments/schema/mutations.py:673`:
```
    if order.lease.status != LeaseStatus.OFFERED:
        raise ValueError(_("Cannot resend an order for a lease not in status OFFERED"))
```
and setting the order status to OFFERED also updates the lease status so perhaps it is best to set it to OFFERED here and in `payments.utils.py`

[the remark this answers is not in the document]
a-rmz: Hmmmm good point. I think it would be redundant to set them to `OFFERED` here and then later on the same process mark it as `OFFERED` again. It makes sense that you can't re-send something that hasn't been sent.

Thinking...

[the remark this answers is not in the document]
klemmari1: If we mark the status as OFFERED here, we don't have to do the status change in `resend_offer`.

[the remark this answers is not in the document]
a-rmz: Fair enough, it would make sense that we should not move the status on resend, we should only resend `offered` offers, so it wouldn't make sense. Let's go with setting them as `OFFERED` 👍

payments/migrations/0028_update_order_statuses.py
```diff
@@ -0,0 +1,87 @@
...
+                    ("expired", "Expired"),
+                    ("error", "Error"),
+                    ("paid", "Paid"),
+                    ("paid_man", "Paid manually"),
+                    ("refunded", "Refunded"),
+                ],
+                max_length=9,
+            ),
+        ),
+        migrations.RunPython(change_order_statuses, reverse_code=revert_order_statuses),
```
a-rmz, line 86: Does this work once you've removed `waiting` from the options?

klemmari1: Tested locally and this works both ways. I think it works because of the `update()` instead of `save()`.

a-rmz: So I guess the database does not check and enforce that you only have one of the choices, so I guess it would work?

klemmari1: Yes, exactly.

a-rmz (approved):
## Sweaty palms, good one!

PR #520, opened 2021-04-08
VEN-1216 | Update group permissions on deploy

klemmari1:
## Description :sparkles:

In case of new models or updated permissions, run the `set_group_model_permissions` script on deploy so that groups get the updated permissions.

### Closes :no_good_woman:
**[VEN-1216](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1216): Query for the customer personal information**

### DIFF
--- a/resources/tests/factories.py
+++ b/resources/tests/factories.py
@@ -1,3 +1,5 @@
+import string
+
 import factory
 from django.contrib.gis.geos import Point
 from factory.random import randgen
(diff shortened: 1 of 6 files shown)


resources/tests/factories.py
```diff
@@ -62,7 +63,7 @@ class Meta:
 
 
 class AbstractAreaSectionFactory(factory.django.DjangoModelFactory):
-    identifier = factory.Faker("random_letter")
+    identifier = factory.fuzzy.FuzzyText(length=30)
```
a-rmz, line 66: The fuzzy stuff [is deprecated](https://factoryboy.readthedocs.io/en/stable/fuzzy.html). What's the issue that this was causing?

If we need a unique value, should use [sequences](https://factoryboy.readthedocs.io/en/stable/reference.html?highlight=sequence#sequence)

klemmari1: Ah, so it seems. I'll swap it to a sequence.

The issue was [this error](https://github.com/City-of-Helsinki/berth-reservations/pull/519/checks?check_run_id=2295013788#step:6:1575) in another PR. There were two piers with the same identifier in `payments/schema/mutations.py:799` which caused it to choose the wrong one and finding no berths in `payments/schema/mutations.py:808`.

haavikko (approved):
Looks good to me!

PR #523, opened 2021-04-09
VEN-1268 | Add berth information to order details

klemmari1:
## Description :sparkles:

Add berth information to OrderDetailsType. The 3 added fields include:

- harbor name
- pier identifier
- berth number

### Closes :no_good_woman:
**[VEN-1268](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1268): Add berth information to order details** 

### Automated tests :gear:️

```
pytest payments/tests/test_payments_queries.py::test_get_order_status
```

### DIFF
--- a/payments/schema/types.py
+++ b/payments/schema/types.py
@@ -296,6 +296,9 @@ class BerthSwitchOfferNode(DjangoObjectType, AbstractOfferNode):
 class OrderDetailsType(graphene.ObjectType):
     order_type = OrderTypeEnum(required=True)
     status = OrderStatusEnum(required=True)
+    harbor = graphene.String(required=True)
+    pier = graphene.String(required=True)
+    berth = graphene.String(required=True)
 
 
 class FailedOrderType(graphene.ObjectType):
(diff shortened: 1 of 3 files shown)


a-rmz (changes requested):
Sorry to change this so late on, but I just noticed the design was a bit different than I had originally thought.

Since the frontend needs to display the information on separate lines, we have to return them as separate fields:
image

So instead of `berth_information`, can you replace it for three fields `berth/pier/harbor`?
 
I updated the ticket with that description as well.

klemmari1:
> Sorry to change this so late on, but I just noticed the design was a bit different than I had originally thought.
> 
> Since the frontend needs to display the information on separate lines, we have to return them as separate fields:
> image
> 
> So instead of `berth_information`, can you replace it for three fields `berth/pier/harbor`?
> 
> I updated the ticket with that description as well.

Alrighty, no problem. Not a big change.

haavikko (approved):
Looks good!

payments/schema/types.py
```diff
@@ -296,6 +296,9 @@ class Meta:
 class OrderDetailsType(graphene.ObjectType):
     order_type = OrderTypeEnum(required=True)
     status = OrderStatusEnum(required=True)
+    harbor = graphene.String()
+    pier = graphene.String()
+    berth = graphene.String()
```
a-rmz, line 301: Can you mark this as required as well? 😇

a-rmz (approved):

PR #527, opened 2021-04-12
VEN-1208 | Proper status transition for expiring offers

a-rmz:
## Description :sparkles:
* Use `set_status` for expiring orders instead of manually setting the status


klemmari1 (approved):
LGTM!

haavikko (approved):
OK! :-) I thought I had changed this code to use set_status before ... wonder if I lost it in some rebase operation at the same time with the settings.py config

PR #528, opened 2021-04-12
VEN-1277 | Get the correct Talpa ID on Order admin

a-rmz:
## Description :sparkles:
* Pass the correct property to the `get_area` so the Talpa ID is shown properly


klemmari1 (approved):
Good stuff!

PR #529, opened 2021-04-12
VEN-1277 | Fix test for additional product storage on ice payload

a-rmz:
## Description :sparkles:
* Some tests which should not have storage on ice were getting the value because of the randomness, this makes sure that we don't get it.

### Automated testing ⚙️ 
```shell
$ pytest payments/tests/test_bambora_payform.py::test_payload_additional_product_order
```


klemmari1 (approved):
Looks good to me!

PR #532, opened 2021-04-13
VEN-1211: Mutation to cancel switch offer

haavikko:
## Description :sparkles:

Added a UpdateBerthSwitchOfferMutation, which currently supports just changing the offer status and due_date (it's based on the corresponding UpdateOrderMutation so it was easiest to do this way).

### Closes :no_good_woman:
**[VEN-1211](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1211):**

### DIFF
--- a/payments/models.py
+++ b/payments/models.py
@@ -1147,7 +1147,11 @@ class BerthSwitchOffer(AbstractOffer):
             raise ValidationError(_("The application has to be a switch application"))
 
         # Validate that once the offer has been sent, it will have a due date
-        if self.status != OfferStatus.DRAFTED and not self.due_date:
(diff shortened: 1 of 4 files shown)


a-rmz (commented):
So to cancel the offer, the mutation would have to set manually the status to cancelled, right?

haavikko:
> So to cancel the offer, the mutation would have to set manually the status to cancelled, right?

In order to cancel the offer, the UI needs to send to UpdateBerthSwitchOfferMutation parameters
* id=<offer_global_id>
* status=cancelled

a-rmz (approved):

PR #533, opened 2021-04-13
VEN-1262 | Create boat when a lease is created

a-rmz:
## Description :sparkles:
* Create boat when a lease is created
* Extract `lookup_boat` to a separate util to lookup or create the necessary boat

When a lease is being created for an application and no boat is passed, it creates a boat based on the information filled on the application. This is handled during the lease creation process, that way we ensure that we don’t end up creating unnecessary boats that will just clutter the system. If necessary, this can be refactored to create boats on other steps of the process.

### Closes :no_good_woman:
**[VEN-1262](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1262):** Create boat during application processing

### Automated tests :gear:️
```shell
```

Create a lease for an existing application, check that the results are the expected:
```graphql
mutation CreateLease {
  createBerthLease(input: {applicationId: "", berthId: "QmVydGhOb2RlOjBhNjEzYTQzLTAxMjktNDQxNS1iNjlhLTZjMmI4ZmJkYzJkMQ=="}) {
    berthLease {
      customer {
        id
      }
      boat {
        id
        name
        model
        registrationNumber
        width
        length
        owner {
          id
        }
      }
    }
  }
}
```

### DIFF
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -33,6 +33,7 @@ from ..services import BerthInvoicingService
 from ..stickers import get_next_sticker_number
 from ..utils import calculate_berth_lease_start_date, terminate_lease
 from .types import BerthLeaseNode, WinterStorageLeaseNode
+from .utils import lookup_or_create_boat
 
 
 class AbstractLeaseInput:
@@ -62,7 +63,7 @@ class CreateBerthLeaseMutation(graphene.ClientIDMutation):
                         "Can not specify both application and customer when creating a new berth lease"
                     )
                 )
-            application = get_node_from_global_id(
+            application: BerthApplication = get_node_from_global_id(
                 info, application_id, only_type=BerthApplicationNode, nullable=False,
             )
             if not application.customer:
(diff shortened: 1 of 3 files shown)


leases/schema/mutations.py
```diff
@@ -84,29 +85,14 @@ def lookup_application_and_customer(cls, info, input):
                 )
             )
 
-    @classmethod
-    def lookup_boat(cls, info, input):
```
haavikko, line 88: good refactoring :-)

leases/schema/utils.py
```diff
@@ -10,3 +17,41 @@ def parse_to_dict(object):
         return {"id": to_global_id(node_type, id), "error": error}
 
     return parse_to_dict
+
+
+def lookup_or_create_boat(info, input: dict) -> Optional[Boat]:
```
haavikko, line 22: The related case VEN-1271 (script to create boats for processed applications) describes this matching logic::

> If the customer already has a boat with the same registration number, update the fields of the existing boat. The information in the application is assumed to be more accurate.

The same logic would apply here?

haavikko (changes requested):

leases/schema/utils.py
a-rmz: Good point, this is definitely a simplified version, but maybe it's worth investing a bit on making it a bit more robust. 👍

a-rmz: Added

haavikko (approved):
OK great!

PR #534, opened 2021-04-14
VEN-1282 | Mark a berth as reserved when there's an open offer

a-rmz:
## Description :sparkles:
* Mark a berth as reserved when there's an offer for the berth either `DRAFTED` or `OFFERED`
  * If the offer is accepted, the lease will be the one reserving the berth rather than the offer
  * In any other case (rejected, cancelled, expired), the berth will be released

### Closes :no_good_woman:
**[VEN-1282](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1282):** Sending an offer should reserve the berth

### Automated tests :gear:️
```shell
$ pytest resources/tests/test_resources_models.py::test_berth_is_available_inactive_offer
$ pytest resources/tests/test_resources_models.py::test_berth_is_not_available_active_offer
```

```python
# 1. Create an offer
from payments.tests.factories import BerthSwitchOfferFactory
from payments.enums import OfferStatus
offer = BerthSwitchOfferFactory(status=OfferStatus.DRAFTED)

# 2. Check the status of the berth (not available)
from resources.models import Berth
assert not Berth.objects.get(id=offer.berth.id).is_available

# 3. Move the order to accepted or rejected
offer.set_status(OfferStatus.REJECTED)

# 4. Check the status of the berth (available)
from resources.models import Berth
Berth.objects.get(id=offer.berth.id).is_available
```

### DIFF
--- a/resources/models.py
+++ b/resources/models.py
@@ -3,6 +3,7 @@ from django.conf import settings
 from django.contrib.gis.db import models
 from django.core.files.storage import FileSystemStorage
 from django.db.models import (
+    BooleanField,
     Count,
     DecimalField,
     Exists,
@@ -16,7 +17,7 @@ from django.db.models import (
     UniqueConstraint,
     Value,
 )
-from django.db.models.expressions import RawSQL
+from django.db.models.expressions import ExpressionWrapper, RawSQL
 from django.db.models.functions import Coalesce
 from django.utils.translation import gettext_lazy as _
 from munigeo.models import Municipality
@@ -30,7 +31,7 @@ from leases.utils import (
     calculate_berth_lease_start_date,
     calculate_winter_storage_lease_end_date,
 )
-from payments.enums import PriceTier
+from payments.enums import OfferStatus, PriceTier
 from utils.models import TimeStampedModel, UUIDModel
 
 from .enums import AreaRegion, BerthMooringType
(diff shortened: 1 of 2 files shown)


resources/models.py
```diff
@@ -663,21 +665,37 @@ def get_queryset(self):
...
+        # (this requires separate Exists clauses
+        active_offers = ~Exists(
+            BerthSwitchOffer.objects.filter(
+                status__in=(OfferStatus.DRAFTED, OfferStatus.OFFERED),
+                berth=OuterRef("pk"),
+            ).values("pk")
+        )
+
+        # Need to explicitly mark the result of the AND as a BooleanField
+        is_available = ExpressionWrapper(
```
haavikko, line 696: Haven't used ExpressionWrapper myself, nice to learn something new :-)

resources/models.py
```diff
@@ -663,21 +665,37 @@ def get_queryset(self):
...
+            Q(active_leases & active_offers), output_field=BooleanField()
+        )
```
haavikko, line 698: Does the query change have any performance effect? Maybe add a note to testing instructions to test in staging with a large harbor

haavikko (approved):
See the comment about testing, but other than that, great!

resources/models.py
a-rmz: I don't think I had either, definitely a good way to learn on the go 🏃

resources/models.py
a-rmz: It _shouldn't_, the only thing that's being added to the generated SQL is this:
```sql
SELECT
    U0."id" 
FROM
    "payments_berthswitchoffer" U0 
WHERE (
    U0."berth_id" = "resources_berth"."id" 
    AND U0."status" IN (drafted, offered)
)
```

And since we're using `EXISTS`, the evaluation will stop as soon as it finds the opposite case.

But maybe worth testing with Puotila 🤷  just to be safe.

PR #536, opened 2021-04-15
VEN-1283: modify UpdateOrderMutation to allow multiple Orders at a time

haavikko:
## Description :sparkles:
* UpdateOrderMutation is now more like ResendOrderMutation and ApproveOrderMutation, allowing multiple orders in input

### Closes :no_good_woman:
**[VEN-1283(https://helsinkisolutionoffice.atlassian.net/browse/VEN-1283):**

### DIFF
--- a/payments/schema/mutations.py
+++ b/payments/schema/mutations.py
@@ -40,7 +40,7 @@ from utils.relay import get_node_from_global_id
 from utils.schema import update_object
 
 from ..enums import OfferStatus, OrderStatus, OrderType, ProductServiceType
-from ..exceptions import VenepaikkaPaymentError
+from ..exceptions import OrderStatusTransitionError, VenepaikkaPaymentError
 from ..models import (
     AdditionalProduct,
     BerthProduct,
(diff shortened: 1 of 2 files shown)


payments/schema/mutations.py
```diff
@@ -394,52 +394,67 @@ def mutate_and_get_payload(cls, root, info, **input):
         return CreateAdditionalProductOrderMutation(order=order)
 
 
+class UpdateOrderInput(OrderInput, graphene.InputObjectType):
+    id = graphene.ID(required=True)
+
+
 class UpdateOrderMutation(graphene.ClientIDMutation):
```
a-rmz, line 401: Should this be renamed to `UpdateOrdersMutation`?

Or can we even keep them as separate mutations?

haavikko: hmm yes UpdateOrder`s`Mutation makes more sense

haavikko: UpdateOrdersMutation has all functionality that the old UpdateOrderMutation had, so I don't think it's worth keeping

a-rmz: Fair enough, I guess if we need the individual one, we can add it later.

payments/schema/mutations.py
```diff
@@ -394,52 +394,67 @@ def mutate_and_get_payload(cls, root, info, **input):
...
-        return UpdateOrderMutation(order=order)
+        for order_input in orders:
+            order_id = order_input.pop("id")
+            try:
+                with transaction.atomic():
+                    order = get_node_from_global_id(
+                        info, order_id, only_type=OrderNode, nullable=False
+                    )
+                    lease_id = order_input.pop("lease_id", None)
+                    if lease_id:
```
a-rmz, line 424: This can now be simplified toÑ
```suggestion
                    if lease_id := order_input.pop("lease_id", None):
```

payments/schema/mutations.py
```diff
@@ -396,52 +396,66 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+
+                    # handle case where order_input has both lease and status.
+                    # set order status only after changing the lease, because setting order status
+                    # usually triggers a change in lease status.
+                    new_status = order_input.pop("status", None)
+                    update_object(order, order_input)
+                    if new_status:
+                        order.set_status(new_status, _("Manually updated by admin"))
+
+            except (ValidationError, IntegrityError, VenepaikkaGraphQLError) as e:
```
a-rmz, line 452: We should also catch here for `OrderStatusTransitionError`. If you try to move one order to an invalid state, the whole request will fail because the exception is not caught.

a-rmz (changes requested):

payments/schema/mutations.py
haavikko: good catch, fixed

a-rmz: Great! Now it did catch the error

```json
{
  "data": {
    "updateOrders": {
      "failedOrders": [
        {
          "id": "T3JkZXJOb2RlOjE2NmNlZTdiLWYyMDYtNDZmYy05NjVmLWU1NmUwZjM0NmIwMg==",
          "error": "Cannot set order co73zgdhkwcqa state to \"refunded\", it is in an invalid state \"drafted\"."
        }
      ],
      "successfulOrders": [
        {
          "id": "T3JkZXJOb2RlOjUwNGQ5NzQ4LWUyMDQtNGZlNy05OWFiLTQ5NWJmZDBlZDcxNg==",
          "comment": "Another new comment!",
          "price": "100.00",
          "taxPercentage": "24.00",
          "dueDate": "2021-11-30",
          "status": "DRAFTED",
          "customer": {
            "id": "UHJvZmlsZU5vZGU6YzRhYzljNjUtYzdjYy00OTkyLWFkYzEtNjZkMGM0YmRlZjA3"
          },
          "product": null,
          "lease": null
        }
      ]
    }
  }
}
```

a-rmz (approved):
## ✨ Great job ✨

PR #540, opened 2021-04-16
VEN-1245 | Add mutation for creating CustomerProfile for the current user

klemmari1:
## Description :sparkles:

Add `CreateMyBerthProfileMutation` that calls the open-city-profile backend with the profile_token of the current user to fetch the `Profile` UUID of the user.
Then it creates a corresponding `CustomerProfile` with the fetched UUID.

- Add `get_my_profile` query to `ProfileService`.

### Closes :no_good_woman:
**[VEN-1245](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1245): Add authorisation for customers** 

### Automated tests :gear:️

```
pytest customers/tests/test_customers_mutations.py::test_create_my_berth_profile
pytest customers/tests/test_customers_mutations.py::test_create_my_berth_profile_does_not_exist
pytest customers/tests/test_customers_mutations.py::test_create_my_berth_profile_already_exists

pytest customers/tests/test_profile_service.py::test_get_my_profile
```


customers/services/profile.py
```diff
@@ -127,6 +127,35 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
+                }}
+            }}
+        """
+
+        response = self.query(query)
+        my_profile = response.pop("myProfile")
+        if not my_profile:
+            return None
+        user = self.parse_user(my_profile)
+        return user
```
a-rmz, line 157:
```suggestion
        return self.parse_user(my_profile) if my_profile else None
```

a-rmz, line 157:
```suggestion
        return self.parse_user(my_profile) if my_profile else None
```

customers/services/profile.py
```diff
@@ -127,6 +127,35 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
+                        address
+                        postal_code: postalCode
+                        city
+                    }}
+                }}
+            }}
+        """
+
+        response = self.query(query)
+        my_profile = response.pop("myProfile")
```
a-rmz, line 153: We should do here a `response.get`, on the very unlikely case that the response didn't contain the `myProfile` prop, it could fail unexpectedly.

customers/tests/conftest.py
```diff
@@ -80,6 +82,18 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def mocked_response_my_profile(data="", *args, **kwargs):
+    def wrapper(*args, **kwargs):
+        if data != "":
```
a-rmz, line 87: ?
```suggestion
        if data:
```

customers/schema/mutations.py
```diff
@@ -218,6 +220,38 @@ def mutate_and_get_payload(cls, root, info, **input):
...
+        if profile_exists:
+            return None
+
+        profile_service = ProfileService(profile_token=profile_token)
+        my_profile = profile_service.get_my_profile()
+        if not my_profile:
+            raise VenepaikkaGraphQLError("Open city profile not found")
+
+        try:
+            profile = CustomerProfile.objects.create(id=my_profile.id, user=user)
```
a-rmz, line 246: Does the login/sign in process actually create a user on the system? Since we're not using `user` for anything, idk how valuable is it to keep store it.

customers/tests/conftest.py
klemmari1: I left the default as `""` because if the customer profile is not created, the data will be `None` and I wanted it so that I would be able to test that case.

customers/schema/mutations.py
klemmari1: `GraphQLApiTokenAuthentication` will create an `User` instance if it does not exist using `helusers.oidc.ApiTokenAuthentication.authenticate` -> `helusers.oidc.resolve_user` -> `helusers.user_utils.get_or_create_user`

The main reason I added it here so that we can check if the `CustomerProfile` exists for the user without querying the profile backend:
```
profile_exists = CustomerProfile.objects.filter(user=user).exists()
```

a-rmz: Can't we replace the `user resolver` with our own that would fetch the `CustomerProfile` based on the profile id on the JWT claims? I think start using `user` now would add another layer of unnecessary complexity.

a-rmz: On second though... since now the customers become _users_ of our system, it would kinda make sense that we add them as users.

I was thinking that, if we replace the resolver, then the admins who don't have a customer profile won't be able to access the system anymore (since we'd be looking for CP, not for Users).

a-rmz: We _mabye_ should add another permission `Group` customer in case it'd be necessary and assign it to this users by default..? Not very sure about this one.

haavikko: I agree that the concept of `User` makes good sense for customers.
Adding a "customers" `Group` sounds good to me.

users/utils.py
```diff
@@ -37,3 +42,47 @@ def user_has_change_permission(user, *models):
...
+        user.groups.count() == 1
+        and user.groups.first().name == settings.CUSTOMER_GROUP_NAME
+    ):
+        return True
+    else:
+        return False
+
+
+def get_berth_customers_group():
+    return Group.objects.get(name=settings.CUSTOMER_GROUP_NAME)
```
a-rmz, line 60: Can we somehow cache the result of this? Since the value won't probably change often (or ever)

users/utils.py
```diff
@@ -37,3 +42,47 @@ def user_has_change_permission(user, *models):
...
+
+def user_is_linked_to_node(user, node):
+    if user is None:
+        return False
+    return user == get_node_user(node)
+
+
+def validate_user_is_linked_to_node(user, node):
+    if not user_is_linked_to_node(user, node):
+        raise PermissionDenied(_("You do not have permission to perform this action."))
```
a-rmz, line 88: Where are we gonna use this?

users/utils.py
haavikko: Yes, the thought crossed my mind too.

users/utils.py
haavikko: It would be useful for mutations where customer wants to edit an object. For example, in VEN-1222 it would be used to check if the current user is the owner of the application that's being edited.

customers/tests/conftest.py
haavikko: Added comment in code to clarify

users/utils.py
haavikko: fixed

customers/services/profile.py
```diff
@@ -133,6 +133,32 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
         user = self.parse_user(response.pop("profile"))
         return user
 
+    def get_my_profile(self) -> Optional[HelsinkiProfileUser]:
+        query = """
+            query MyProfile {{
+                myProfile {{
```
a-rmz, line 139: Small nitpick:
```suggestion
                my_profile: myProfile {{
```

customers/services/profile.py
```diff
@@ -133,6 +133,32 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
+                        address
+                        postal_code: postalCode
+                        city
+                    }}
+                }}
+            }}
+        """
+
+        response = self.query(query)
+        my_profile = response.get("myProfile")
```
a-rmz, line 159: Nitpick pt 2:
```suggestion
        my_profile = response.get("my_profile")
```

customers/tests/conftest.py
```diff
@@ -43,23 +43,25 @@ def hki_profile_address() -> dict:
...
+    faker = Faker()
+    profile = CustomerProfileFactory()
+    return {
+        "id": to_global_id(ProfileNode, profile.id),
+        "first_name": profile.user.first_name,
+        "last_name": profile.user.last_name,
+        "primary_email": {"email": profile.user.email},
+        "primary_phone": {"phone": faker.phone_number()},
+        "primary_address": MOCK_HKI_PROFILE_ADDRESS,
+    }
```
a-rmz, line 56: I had added this on a previous PR since I needed the functionality, just double-checking that we won't have repeated utils.

users/utils.py
```diff
@@ -37,3 +42,53 @@ def user_has_change_permission(user, *models):
...
+
+
+_berth_customers_group = None
+
+
+def get_berth_customers_group():
+    global _berth_customers_group
+    if not _berth_customers_group:
+        _berth_customers_group = Group.objects.get(name=settings.CUSTOMER_GROUP_NAME)
+    return _berth_customers_group
```
a-rmz, line 66: Can we use the power of Python for this one? E.g. [lru_cache](https://docs.python.org/dev/library/functools.html#functools.lru_cache):

```suggestion
@lru_cache(maxsize=32)
def get_berth_customers_group():
    return Group.objects.get(name=settings.CUSTOMER_GROUP_NAME)
```
(`maxsize` is an arbitrary number, but since we don't expect the value to change often, we can keep it as a small number)

users/utils.py
```diff
@@ -37,3 +42,53 @@ def user_has_change_permission(user, *models):
...
+        node_user = node.user
+    elif hasattr(node, "customer") and hasattr(node.customer, "user"):
+        node_user = node.customer.user
+    return node_user
+
+
+def user_is_linked_to_node(user, node):
+    if user is None:
+        return False
+    return user == get_node_user(node)
```
a-rmz, line 89: Shouldn't it be the other way around tho? I'm probably missing something, but our nodes would be connected to the customer rather than the user, wouldn't they?

I'm thinking, for example, of the berth lease case. They have a `customer` which is a `CustomerProfile`, not a user. But maybe I'm missing something on how graphene/relay handle all this?

a-rmz:
Also, thinking a bit more about this, we'd need to create users for all existing customers for the permission checking to work.

haavikko (approved):
Looks good to me!

users/utils.py
a-rmz: Resolved: the customers are connected to users, and the authorization is done based on the user, so we check that the logged user is the same as the customer's user.

PR #544, opened 2021-04-21
VEN-1258 | Add GDPR API

klemmari1:
## Description :sparkles:

Add serializers so that the customer can query their personal data.

I've tried to keep the duplicate information to a mininum so that for example Boats are listed only once in the customer profile and when they are linked to Leases, for example, we only show the boat id in those instances.

There is a missing feature in the helsinki-profile-gdpr-api that does not automatically serialize forward facing relations (eg ForeignKeys) so, for now, those have to be serialized "manually". For example: `"name": "berth", "accessor": lambda x: x.serialize()`

### Closes :no_good_woman:
**[VEN-1258](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1258): Add the GDPR API** 

### Automated tests :gear:️

```
pytest customers/tests/test_gdpr_api.py
```


customers/tests/test_gdpr_api.py
```diff
@@ -0,0 +1,713 @@
...
+                    ],
+                }
+            ],
+        },
+        winter_storage_leases_dict,
+    )
+
+
+@override_settings(GDPR_API_DELETE_SCOPE="berths.gdprdelete")
+def test_delete_profile(rest_api_client, requests_mock, settings):
```
charn, line 700: I would expect there probably are situations where deleting the `CustomerProfile` is not allowed? Maybe there the existing `on_delete` work perfectly, but maybe adding a test for that kind of situation might be useful.

charn (approved):
As far as I can tell this looks quite good. :+1:  It seems the separate gdpr-api module works almost perfectly although there's a missing feature for serializing `ForeignKey` objects which implement the `SerializableMixin`.

If it's feasible I would probably still add a test for a case where deleting the `CustomerProfile` is not allowed, I would assume there's situation where this might be the case.

a-rmz (approved):

PR #546, opened 2021-04-21
Get refund email from application

klemmari1:
## Description :sparkles:

If refund `customer_email` does not exist, try to fetch the email from the related application `email` field.

### Closes :no_good_woman:
-

### Automated tests :gear:️

```
pytest payments/tests/test_bambora_payform_refund.py::test_initiate_refund_no_order_email
pytest payments/tests/test_bambora_payform_refund.py::test_initiate_refund_no_order_email_or_application
```

### DIFF
--- a/payments/tests/test_bambora_payform_refund.py
+++ b/payments/tests/test_bambora_payform_refund.py
@@ -47,7 +47,7 @@ def test_initiate_refund_success(provider_base_config: dict, order: Order):
         order=order, token="98765", valid_until=now() - relativedelta(hours=1)
     )
     valid_token = OrderToken.objects.create(
-        order=order, token="12345", valid_until=now() + relativedelta(days=7)
+        order=order, token="12345", valid_until=now() - relativedelta(days=7)
     )
 
     payment_provider = create_bambora_provider(provider_base_config, request)
@@ -82,7 +82,7 @@ def test_initiate_refund_no_order_email(provider_base_config: dict, order: Order
     order.save()
 
     OrderToken.objects.create(
-        order=order, token="12345", valid_until=now() + relativedelta(days=7)
+        order=order, token="12345", valid_until=now() - relativedelta(days=7)
     )
 
     payment_provider = create_bambora_provider(provider_base_config, request)
@@ -111,7 +111,7 @@ def test_initiate_refund_no_order_email_or_application(
     order.save()
 
     OrderToken.objects.create(
-        order=order, token="12345", valid_until=now() + relativedelta(days=7)
+        order=order, token="12345", valid_until=now() - relativedelta(days=7)
     )
 
     payment_provider = create_bambora_provider(provider_base_config, request)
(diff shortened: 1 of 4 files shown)


payments/tests/test_bambora_payform_refund.py
```diff
@@ -69,6 +69,58 @@ def test_initiate_refund_success(provider_base_config: dict, order: Order):
...
+def test_initiate_refund_no_order_email(provider_base_config: dict, order: Order):
+    """Test the request creator constructs the payload base and returns a url that contains a token"""
+    request = RequestFactory().request()
+    order.status = OrderStatus.PAID
+    order.email = None
+    order.lease.status = LeaseStatus.PAID
+    order.lease.save()
+    order.save()
+
+    OrderToken.objects.create(
```
haavikko, line 84: This token is no longer valid? It doesn't have effect on test result though

payments/tests/test_bambora_payform_refund.py
```diff
@@ -69,6 +69,58 @@ def test_initiate_refund_success(provider_base_config: dict, order: Order):
...
+    request = RequestFactory().request()
+    order.status = OrderStatus.PAID
+    order.customer_email = None
+    order.lease.status = LeaseStatus.PAID
+    order.lease.application = None
+    order.lease.save()
+    order.save()
+
+    OrderToken.objects.create(
+        order=order, token="98765", valid_until=now() - relativedelta(hours=1)
```
haavikko, line 114: Here too, token valid_until is in the past, doesn't change the result though

haavikko (changes requested):
Commented on the minor issue of token expiry in tests.
I can take up the branch if you are already busy with the new project.

payments/tests/test_bambora_payform_refund.py
klemmari1: Ah yep those slipped in there :sweat_smile: Good spot!

klemmari1:
I also pushed a fix for the test that failed randomly here.

payments/tests/test_bambora_payform_refund.py
a-rmz: It doesn't. We just use the token to ensure that it was generated through Bambora. Since the payment could've been issued long time ago, it doesn't really matter if it's valid or not, as long as it exists.

a-rmz: Actually, if there are valid tokens, that means that the payment has not been made/settled, since we expire all the tokens when we receive the confirmation from VismaPay. It's not very likely to happen anyway, but I just noticed that.

a-rmz: Ok, solved, I pushed a commit that also covers this case.

a-rmz (approved):

haavikko (approved):
OK, great, thanks for the clarification!

PR #547, opened 2021-04-21
VEN-1300: disallow setting status of paid_manually invoices to refunded

haavikko:
## Description :sparkles:
There's actually another check in initiate_refund that comes to play before set_status is called, but let's fix it anyway.

### Closes :no_good_woman:
**[VEN-1300](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1300):**


klemmari1 (approved):
LGTM!

PR #550, opened 2021-04-22
VEN-877 | Import leases from leases csv

a-rmz:
## Description :sparkles:
* Add a command to create a profile (on Profiili and vene)
* Add the command to import the winter leases from the leases csv

### Closes :no_good_woman:
**[VEN-877](https://helsinkisolutionoffice.atlassian.net/browse/VEN-877):** find a way to import marked WS data from latest Timmi data file and import it to production

1. Get the `profile-token` from the Admin UI
```js
JSON.parse(localStorage.apiTokens)["https://api.hel.fi/auth/helsinkiprofile"]
```
2. Download the attached CSV ([test_winter_invoices.csv](https://github.com/City-of-Helsinki/berth-reservations/files/6357930/test_winter_invoices.csv))
3. Call the create command
```shell
python manage.py import_winter_leases_from_csv \
  --profile-token "" \
  --lease-file-path "/path/to/test_winter_invoices.csv"
```

* The CSV should have the following format:
```
Section ID;Place number;Customer name;Customer email;Customer phone;Width;Length;Boat registration;Comments
```
The required fields are `section id`, `place number`, and `customer name` (`{last name} {first name}`)

### DIFF
--- a/customers/services/profile.py
+++ b/customers/services/profile.py
@@ -1,9 +1,15 @@
 from dataclasses import dataclass
-from typing import Dict
+from typing import Dict, List, Optional, Union
 from uuid import UUID
 
 import requests
+from django.db import transaction
 
+from customers.exceptions import (
+    MultipleProfilesException,
+    NoProfilesException,
+    ProfileServiceException,
+)
 from utils.relay import from_global_id, to_global_id
 
 PROFILE_API_URL = "PROFILE_API_URL"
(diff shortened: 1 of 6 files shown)


customers/services/profile.py
```diff
@@ -127,13 +133,117 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
+        phone: str = "",
+        force_only_one: bool = True,
+    ) -> Union[List[HelsinkiProfileUser], HelsinkiProfileUser]:
+        """
+        Find the profile based on the passed criteria, the missing parameters
+        are replaced by empty values.
+
+        force_only_one: bool -> If more than one profile is found, it will raise an error
+        """
+        query = f"""
```
haavikko, line 150: +1 Using a f-string makes the query more readable :-)

customers/services/profile.py
```diff
@@ -127,13 +133,117 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
+                    id
+                }
+            }
+        }
+        """
+        response = self.query(query=mutation, variables=variables)
+        profile = None
+        if global_id := response.get("createProfile", {}).get("profile", {}).get("id"):
+            profile_id = from_global_id(global_id)
+            profile = CustomerProfile.objects.create(id=profile_id)
```
haavikko, line 231: Should there be an else: branch that logs some error info from response if global_id is not found?

customers/services/profile.py
```diff
@@ -127,13 +133,117 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
 
     def parse_user(self, profile: Dict[str, dict]) -> HelsinkiProfileUser:
         user_id = from_global_id(profile.pop("id"))
-        email = (profile.pop("primary_email") or {}).get("email")
-        phone = (profile.pop("primary_phone") or {}).get("phone")
+        email = None
+        phone = None
+        if "primary_email" in profile:
+            primary_email = profile.pop("primary_email")
+            email = primary_email is not None and primary_email.get("email")
```
haavikko, line 243: if primary_email is None, then email == False, is that intended?

customers/services/profile.py
```diff
@@ -127,13 +133,117 @@ def get_profile(self, id: UUID) -> HelsinkiProfileUser:
...
+        if "primary_phone" in profile:
+            primary_phone = profile.pop("primary_phone")
+            phone = primary_phone is not None and primary_phone.get("phone")
```
haavikko, line 246: if primary_phone is None, then phone == False, is that intended?

haavikko (commented):
Kids woke up, I'll continue reviewing later :-)

leases/management/commands/import_winter_leases_from_csv.py
```diff
@@ -0,0 +1,171 @@
...
+            profile = CustomerProfile.objects.get(id=helsinki_profile.id)
+        except (NoProfilesException, CustomerProfile.DoesNotExist):
+            logger.debug(f"Creating profile for {last_name}, {first_name} ({email})")
+            profile = self.profile_service.create_profile(
+                first_name, last_name, email, phone
+            )
+        return profile
+
+    def handle(
+        self, *args, profile_token=None, lease_file_path=None, dry_run=False, **options
```
haavikko, line 79: dry_run argument not used?

leases/management/commands/import_winter_leases_from_csv.py
```diff
@@ -0,0 +1,171 @@
...
+            "--lease-file-path",
+            nargs="?",
+            type=str,
+            help="[Required] The file to parse",
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            default=False,
+            help="Only show a list of the invoices that can be sent "
```
haavikko, line 55: help text needs updating?

leases/management/commands/import_winter_leases_from_csv.py
```diff
@@ -0,0 +1,171 @@
...
+            )
+            profile = CustomerProfile.objects.get(id=helsinki_profile.id)
+        except (NoProfilesException, CustomerProfile.DoesNotExist):
+            logger.debug(f"Creating profile for {last_name}, {first_name} ({email})")
+            profile = self.profile_service.create_profile(
+                first_name, last_name, email, phone
+            )
+        return profile
+
+    def handle(
```
haavikko, line 78: How should transactions be managed?
Changes to the profile service (profiles created) can not be rolled back.
Maybe it's still better to run in a transaction: profiles created during previous (failed) runs will be found in subsequent runs.

leases/management/commands/import_winter_leases_from_csv.py
```diff
@@ -0,0 +1,171 @@
...
+
+                try:
+                    place = WinterStoragePlace.objects.get(
+                        winter_storage_section_id=lease_input.section_id,
+                        number=lease_input.place_number,
+                    )
+
+                    boat = None
+                    if lease_input.boat_width and lease_input.boat_length:
+                        boat, _created = Boat.objects.get_or_create(
```
haavikko, line 117: What if admins already have created a Boat for the user? Maybe skip get_or_create step if the customer already has a boat with the same registration number?

leases/management/commands/import_winter_leases_from_csv.py
```diff
@@ -0,0 +1,171 @@
...
+                            str(e),
+                        )
+                    )
+
+        with open("./successful_leases.csv", "w+", encoding="utf-8") as successful_file:
+            writer = csv.writer(successful_file, delimiter=",")
+            writer.writerow(["Lease id", "Section id", "Place number", "Customer name"])
+            writer.writerows(successful)
+
+        with open("./multiple_profiles.csv", "w+", encoding="utf-8") as multiple_file:
```
haavikko, line 161: What's the plan for processing multiple_profiles.csv and failed_leases.csv?
Should the admins go through them manually? If so, should these files have more information about the failed lease? Or line number in original file, in case admins want to look up information from there?

customers/services/profile.py
a-rmz: Hmmm. I _think_ the only case where the global id would be `None` is if there's some error on the Profile service, and that's being caught now by the `query` method, so probably we don't even need this condition.

```python
r = requests.post(url=self.api_url, json=body, headers=headers)
r.raise_for_status()
response = r.json()
if errors := response.get("errors"):
    raise ProfileServiceException(str(errors))
```

customers/services/profile.py
a-rmz: Good catch.

leases/management/commands/import_winter_leases_from_csv.py
a-rmz: Good point :derp: I think in this case the `dry-run` would be a bit complex to code for something that will only be run once, so I'd rather just remove it, unless you'd have an objection?

leases/management/commands/import_winter_leases_from_csv.py
a-rmz: > Changes to the profile service (profiles created) can not be rolled back
But with the `get_or_create`, we make sure that if the profile was created on a previous run, it will still be found.


> How should transactions be managed?
I'll add the transaction for each processed row 👍

leases/management/commands/import_winter_leases_from_csv.py
a-rmz: Added first a check if the customer has a matching boat :thu

leases/management/commands/import_winter_leases_from_csv.py
a-rmz: Not on the first place. Ideally, we'd get those files and see if there's anything that we can fix manually. If there's anything regarding the customer information, the we hand it to the admins.

leases/management/commands/import_winter_leases_from_csv.py
haavikko: ok by me!

customers/services/profile.py
haavikko: ok

haavikko (approved):
"pre-approved" - feel free to merge after removing --dry-run parameter :-)

PR #556, opened 2021-04-27
Tech | Optimise inlines on CustomerProfile admin

a-rmz:
## Description :sparkles:
* Replace the default `select` for the inline lease forms with `select2` to avoid fetching all the winter places and sections for every instance of the lease


haavikko (approved):
+1

PR #557, opened 2021-04-28
VEN-1304: update application status when switch offer status is updated

haavikko:
## Description :sparkles:

Set application status at these points in offer life cycle:
* offer rejected -> application rejected
* offer expired -> application expired
* After the Offer is generated -> the application is set to OFFER_GENERATED status.
* Offer never needs to enter ApplicationStatus.HANDLED state (which is rendered as "invoice paid" in the admin UI)

### Closes :no_good_woman:
**[VEN-1304](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1304):**

### DIFF
--- a/applications/models.py
+++ b/applications/models.py
@@ -239,7 +239,7 @@ class BerthApplication(BaseApplication):
         if self.berth_switch:
             allowed_statuses_for_switch = [
                 ApplicationStatus.PENDING,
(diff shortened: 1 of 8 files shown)


a-rmz (approved):
## Good job! 🔥 
I tested it locally and it worked 😸

PR #559, opened 2021-04-28
Tech | Remove deprecated harbor/winter area map

a-rmz:
## Description :sparkles:
* `HarborMap`/`WinterStorageAreaMap`s are not used (and won't be used). Since we don't support file upload either, let's just get rid of it to avoid extra confusion and unnecessary code


haavikko:
There are now two HarborMaps in staging db which are obsolete. I didn't notice a migration in this PR to drop the table?

haavikko:
review env deployment seems to fail, it's probably due to the missing migrations to drop the models?

haavikko (changes requested):
Need migrations to drop the tables?

a-rmz:
Whoopsies 😅 quite an important thing to miss... fixed now 👍

haavikko (approved):
Lookin' good!

PR #563, opened 2021-04-30
Fix | Connect the switch application with the new lease

a-rmz:
## Description :sparkles:
* Connect the switch application with the new lease

When the `switch offer` is accepted, the new lease is generated but it's not connected to the application. 

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_payments_mutations_accept_berth_switch_offer.py::test_accept_offer
```


haavikko (approved):
Looks good to me :-)

PR #565, opened 2021-05-03
VEN-1319 | Refactor refunding to replace amount for products

a-rmz:
## Description :sparkles:
* Refactor refunding to replace amount for products

As per VismaPay support:
> _We checked some of your test refunds and noticed, that you are refunding with custom amount. With Helsingin kaupunki Liikuntapalvelu (931) merchant account you have to refund products, otherwise the accounting data will break._

### Closes :no_good_woman:
**[VEN-1319](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1319):** Refund VismaPay by products

### Automated tests :gear:️
```shell
$ pytest payments/tests/test_bambora_payform_refund.py
$ pytest payments/tests/test_payments_mutations_refunds.py
```

TBD

### DIFF
--- a/payments/providers/bambora_payform.py
+++ b/payments/providers/bambora_payform.py
@@ -1,8 +1,9 @@
 import hashlib
 import hmac
 import logging
+from dataclasses import dataclass
 from datetime import datetime
-from typing import Iterable
+from typing import Iterable, List, Optional
 
 import requests
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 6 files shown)


payments/providers/bambora_payform.py
```diff
@@ -288,21 +398,33 @@ def initiate_refund(self, order: Order) -> OrderRefund:
...
+
+        # The refund requires the product IDs that Bambora assigns them
+        payment_details = self.get_payment_details(order_number)
+
+        products = []
+        refund_amount = 0
+
+        for product in payment_details.payment_products:
+            products.append({"id": product.product_id, "count": 1})
+            refund_amount = price_from_fractional_int(product.price)
```
haavikko, line 419: Would it make sense to validate that refund_amount calculated from products matches order.total_price?
Not necessarily fail, but log a warning if the amounts do not match?

If, for any reason (bug in Bambora or whatever), the refund_amount is something other than order.total_price, then terminate_lease() will not be called later, and it would be good to have some trace about what happened.

haavikko (approved):
Please check comment about logging, but looks good to me otherwise!

payments/providers/bambora_payform.py
a-rmz: Hmmm right. I thought about it but I didn't come to a conclusion on how would we handle that case. Do we (devs) have to manually check and fix the order? If it has already been registered on VismaPay, what do we do about it?

Probably we should raise an error since we don't expect it to happen, and if it does, then that means trouble.

haavikko: My reasoning for allowing the refund to proceed with a different amount was that Visma Pay is more likely than us correct about the amount that the customer has actually paid, and using the same amount for refund will more likely lead to the result that customer&admins wanted.

But since this is not expected to happen, raising an error is a good default. If this happens often, we definitely want to know about it so we can investigate.

PR #566, opened 2021-05-03
VEN-1319 | Cast bambora product price to int

a-rmz:
## Description :sparkles:
* VismaPay sends the product price as string (not int), so cast it to int to operate properly


haavikko (approved):
Approved :-) I wonder why they'd use strings for these purposes, as the api docs state the values are integers. Some issue with converting from xml-formatted messages maybe.

PR #586, opened 2021-05-28
VEN-1293 unmarked winter storage fix

haavikko:
## Description :sparkles:
Fix error when retrieving OrderDetails for unmarked winter storage
Add tests for the bug.

### Closes :no_good_woman:
**[VEN-1293](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1293):**

### DIFF
--- a/payments/schema/queries.py
+++ b/payments/schema/queries.py
@@ -153,9 +153,15 @@ class Query:
             section_identifier = order.lease.berth.pier.identifier
             area_name = order.lease.berth.pier.harbor.name
         elif isinstance(order.lease, WinterStorageLease):
-            place_number = str(order.lease.place.number)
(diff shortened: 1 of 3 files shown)


resources/models.py
```diff
@@ -86,7 +86,6 @@ class AbstractArea(TimeStampedModel, UUIDModel):
         help_text=_("ID in the Servicemap system"),
         blank=True,
         null=True,
-        unique=True,
```
Nizar-Rahme, line 89: Why do we have to remove `unique` here?

Nizar-Rahme (approved):
✔️

resources/models.py
haavikko: good catch. This change should be in a different PR!

PR #589, opened 2021-06-02
VEN-1222 | Mutation for deleting application

a-rmz:
## Description :sparkles:
* Add a decorator that allows checking for permissions for admins in case the user is an admin, or checking that the given nodes belong to the customer, in case the user is a customer
* Set the `DeleteBerthApplicationMutation` to use that check to allow customers deleting their own stuff

### Closes :no_good_woman:
**[VEN-1222](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1222):** Mutation for canceling the application

### Automated tests :gear:️
```shell
$ pytest applications/tests/test_applications_mutations.py
```

### DIFF
--- a/applications/schema/mutations.py
+++ b/applications/schema/mutations.py
@@ -7,9 +7,11 @@ from customers.models import CustomerProfile
 from leases.models import BerthLease, WinterStorageLease
 from users.decorators import (
     change_permission_required,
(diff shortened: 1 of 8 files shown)


haavikko (approved):
🤩 looks good to me

PR #596, opened 2021-06-15
VEN-1219 | Update own application

a-rmz:
## Description :sparkles:
* Refactor the `UpdateXXXApplicationMutation` mutations to support a customer updating the information on their own application

### Closes :no_good_woman:
**[VEN-1219](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1219):** Mutation for the customer to update own applications

### Automated tests :gear:️
```shell
$ pytest applications/tests/test_applications_mutations.py
```

### DIFF
--- a/applications/schema/inputs.py
+++ b/applications/schema/inputs.py
@@ -54,7 +54,7 @@ class BerthApplicationInput(BaseApplicationInput):
     choices = graphene.List(graphene.NonNull(HarborChoiceInput), required=True)
 
 
-class UpdateBerthApplicationInput:
+class BaseUpdateApplicationInput:
     id = graphene.ID(required=True)
     customer_id = graphene.ID()
     language = graphene.String()
@@ -80,6 +80,22 @@ class UpdateBerthApplicationInput:
     accept_other_culture_news = graphene.Boolean()
     boat_draught = graphene.Decimal()
     boat_weight = graphene.Decimal()
+
+
+class WinterStorageAreaChoiceInput(graphene.InputObjectType):
+    winter_area_id = graphene.ID(required=True)
+    priority = graphene.Int(required=True)
+
+
+class WinterStorageApplicationInput(BaseApplicationInput):
+    storage_method = WinterStorageMethodEnum(required=True)
+    trailer_registration_number = graphene.String()
+    chosen_areas = graphene.List(
+        graphene.NonNull(WinterStorageAreaChoiceInput), required=True
+    )
+
+
+class UpdateBerthApplicationInput(BaseUpdateApplicationInput):
     accessibility_required = graphene.Boolean()
     boat_propulsion = graphene.String()
     boat_hull_material = graphene.String()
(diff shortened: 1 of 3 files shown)


haavikko (approved):
Looks good! add_choices/remove_choices is an interesting way of implementing the functionality but should work

applications/schema/inputs.py
```diff
@@ -65,3 +92,43 @@ class WinterStorageApplicationInput(BaseApplicationInput):
...
+    accessibility_required = graphene.Boolean()
+    boat_propulsion = graphene.String()
+    boat_hull_material = graphene.String()
+    boat_intended_use = graphene.String()
+    renting_period = graphene.String()
+    rent_from = graphene.String()
+    rent_till = graphene.String()
+    boat_is_inspected = graphene.Boolean()
+    boat_is_insured = graphene.Boolean()
+    agree_to_terms = graphene.Boolean()
```
nikomakela, line 109: I first wrote "shouldn't this be required", but found out that it's not required on model level either. What happens if it is set false? :) In my opinion the "Agree to terms" could have been an UI only feature. Since it's been not mandatory earlier either, of course it's ok like this.

applications/tests/test_applications_mutations.py
```diff
@@ -702,3 +717,315 @@ def test_delete_winter_storage_application_inexistent_application(
...
+                    "municipality": "Helsinki",
+                    "boatWidth": "2.00",
+                    "boatLength": "3.00",
+                    "informationAccuracyConfirmed": True,
+                    "acceptFitnessNews": False,
+                    "acceptLibraryNews": False,
+                    "acceptOtherCultureNews": False,
+                    "acceptBoatingNewsletter": True,
+                    "harborChoices": [{"harbor": {"id": harbor_node_id}}],
+                }
```
nikomakela, line 815: Trivial thing, but this could be a merge of the "variables" -variable and the changed entries so it would be easy to see which values are changed https://stackoverflow.com/a/26853961/784642

applications/tests/test_applications_mutations.py
```diff
@@ -702,3 +717,315 @@ def test_delete_winter_storage_application_inexistent_application(
...
+                    "boatWidth": "2.00",
+                    "boatLength": "3.00",
+                    "informationAccuracyConfirmed": True,
+                    "acceptFitnessNews": False,
+                    "acceptLibraryNews": False,
+                    "acceptOtherCultureNews": False,
+                    "acceptBoatingNewsletter": True,
+                    "winterStorageAreaChoices": [
+                        {"winterStorageArea": {"id": area_node_id}}
+                    ],
```
nikomakela, line 973: Trivial thing, but this could be a merge of the "variables" -variable and the changed entries so it would be easy to see which values are changed https://stackoverflow.com/a/26853961/784642

nikomakela (approved):

applications/schema/inputs.py
a-rmz: Yeah, I'm actually not sure where is this required 😅 but since this is the `update` action, it's not mandatory, but then it probably has to be removed from the update because in theory it was accepted already.

applications/tests/test_applications_mutations.py
a-rmz: Hmmm that's a good idea, I think this one is a bit more explicit, and since we're already doing the assertions like this we'll keep it for consistency 😁

PR #600, opened 2021-06-22
VEN-1360 | Application changelog

a-rmz:
## Description :sparkles:
* Add an `ApplicationChange` model that holds entries of changes to the values of applications
* Create an entry whenever an application is saved
* Create an entry whenever the `Update` mutation is called and the harbor/area choices change
* Small refactor to the logic of how `remove_choices` work, instead of passing the choice id, pass the priority to make it clearer

### Closes :no_good_woman:
**[VEN-1360](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1360):** Add changelog to application

### Automated tests :gear:️
```shell
$ pytest applications/tests/test_applications_mutations.py::test_update_winter_storage_application_by_owner
$ pytest applications/tests/test_applications_mutations.py::test_update_berth_application_by_owner
```

1. Change any value on an application, either through the API or the Django admin
2. Check that the change entry was created, either through the API (`berthApplication.changes`) or the Django Admin

## Screenshots :camera_flash:
**Django Admin:**
image

**GraphQL API:**
image


haavikko (approved):
Looks good!
BTW in YJDH we're using django-simple-history to track model changes

a-rmz:
Ohhhh `django-simple-history` looks like this but done right 😅 might be worth checking before merging this, thanks for the heads up!

a-rmz:
Gave it a try and it was a bit overkill for what we need. It's definitely a useful library, but also doesn't cover the case of changing related models (as we need here for the harbor choices). Maybe worth exploring later if we ever get more time for this. 😁

PR #598, opened 2021-06-17
VEN-1245 | Check owner permissions

a-rmz:
## Description :sparkles:
* Add a util function to only return the resources that belong to the customer making the request
* Add the check for ownership to missing nodes and querysets

### Closes :no_good_woman:
**[VEN-1245](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1245):** Add authorisation for customers

### Related :handshake:
#540 

### Automated tests :gear:️
```shell
$ pytest
```

PR #602, opened 2021-06-23
VEN-1245 | Remove additional permission checks

a-rmz:
## Description :sparkles:
* Remove additional permission checks for some nodes

Some nodes had additional `view` permission checking, which was not allowing for customers to query their own information. This permissions are already being checked on the `Node` level on the `get_queryset`method, so this extra checks were redundant.

### Related :handshake:
**[VEN-1245](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1245):** Add authorisation for customers
#540 
#598 

1. Login as a non-admin user
2. Execute one of the fixed queries (leases, applications), you should see only the nodes that belong to the logged in user
```graphql
query BerthLeases {
  berthLeases {
    count
    totalCount
    edges {
      node {
        id
      }
    }
  }
}
```

### DIFF
--- a/applications/schema/queries.py
+++ b/applications/schema/queries.py
@@ -1,10 +1,6 @@
 import graphene
 from graphene_django.filter import DjangoFilterConnectionField
 
-from customers.models import CustomerProfile
-from leases.models import BerthLease, WinterStorageLease
-from users.decorators import view_permission_required
-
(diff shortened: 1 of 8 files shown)


haavikko (approved):
The permission checks do have tests, right? Looks good to me!

a-rmz:
Huh, I thought I had added explicit tests but apparently I forgot 😅 I'll add a few for customer-owned resources

PR #615, opened 2021-10-22
Upgrade all Python package dependencies

nikomakela:
Upgraded the Python dependencies with pip-compile --upgrade. Providing_args in Signals is now deprecated, so they are removed, since they were there only for documentation. Some fixes added to failing number fakers after update, since the old implementation always returned the max value and raised integrity errors (duplicate rows). Also had to fix the emails, since the faker now faked example.com addresses, which are invalid in app / tests. Added a Tunnistamo amr-fix to oidc implementation.

VEN-1379 VEN-1380 VEN-1381 VEN-1382


nikomakela:
> I guess that all the style changes are because of the upgrade of `black` version?

Yeah, almost all of them are in this commit that upgrades the black: https://github.com/City-of-Helsinki/berth-reservations/pull/615/commits/5f3c393f22a4f25890ad0bda655b18c16c5fd748

a-rmz:
Sweet! Well, I guess if all the tests are passing, that's a good sign 👍

a-rmz (approved):

a-rmz:
Upgraded locally and seems to be working, let's do this

PR #616, opened 2021-10-28
VEN-1394 | Fix end date calculation for terminated leases

a-rmz:
### Closes :no_good_woman:
**[VEN-1394](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1394):** In admin UI it is not possible to terminate lease while season is over

### Automated tests :gear:️
```shell
```

### DIFF
--- a/leases/models.py
+++ b/leases/models.py
@@ -61,6 +61,18 @@ class AbstractLease(TimeStampedModel, UUIDModel):
     def order(self):
         return self.orders.filter(order_type=OrderType.LEASE_ORDER).first()
 
+    @property
+    def has_started(self):
+        return today().date() >= self.start_date
+
+    @property
+    def has_ended(self):
+        return today().date() > self.end_date
+
+    @property
+    def is_ongoing(self):
+        return self.has_started and not self.has_ended
+
     def clean(self):
         if self.boat and self.boat.owner != self.customer:
             raise ValidationError(
--- a/leases/utils.py
+++ b/leases/utils.py
@@ -159,6 +159,36 @@ def calculate_winter_storage_lease_end_date() -> date:
     return calculate_winter_season_end_date()
 
 
+def calculate_lease_termination_date(
+    lease: Union[BerthLease, WinterStorageLease], end_date: date = None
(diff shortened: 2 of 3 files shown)


leases/models.py
```diff
@@ -61,6 +61,18 @@ class Meta:
     def order(self):
         return self.orders.filter(order_type=OrderType.LEASE_ORDER).first()
 
+    @property
+    def has_started(self):
+        return today().date() >= self.start_date
+
+    @property
+    def has_ended(self):
+        return today().date() > self.end_date
```
tuomas777, line 70: Do these also the current timezone into account?

leases/utils.py
```diff
@@ -159,6 +159,34 @@ def calculate_winter_storage_lease_end_date() -> date:
...
+
+    # If the lease is terminated after its season has ended, the end date will be set
+    # to the date when the season ends.
+    elif lease.has_ended:
+        if isinstance(lease, BerthLease):
+            return calculate_season_end_date(lease.start_date)
+        else:  # WinterStorageLease
+            return calculate_winter_season_end_date(lease.start_date)
+
+
```
tuomas777, line 189: What if none of the `if`s matches, should this then return `None` as it seems to do now? Then at least the return type hint is slightly wrong.

a-rmz: Good point, this last case returns the default end date for the lease, so it should return it in the last case.

leases/models.py
a-rmz: Hmmm 🤔 well this calculations are done only by and for the system, so it should use the timezone set by the app? I'm not 100% sure but I guess yes.

tuomas777: For some reason I wasn't sure about that dateutil's today(), but it seems to work correctly, so you are most probably absolutely correct 🙂

tuomas777 (approved):
LGTM 👍 ⛵

leases/models.py
a-rmz: It has worked so far but I never really thought about it 🙈

PR #619, opened 2021-11-04
VEN-1397 | Require customers to use strong authentication

tuomas777:
## Description :sparkles:

Require that all queries and mutations executed by customers are carried out under strong authentication by checking the `loa` claim of the API token.

 🚧 **WIP:** It is still a bit unclear whether this is the way we want to address the need.

### Closes :no_good_woman:
**[VEN-1397](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1397): Only allow users with strong authentication**


tuomas777:
It was decided that this functionality will not be added, at least for now.

PR #620, opened 2021-11-08
VEN-1406: connect application to an existing boat

haavikko:
### Closes :no_good_woman:
**[VEN-1406](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1406):**


a-rmz:
After discussing with @haavikko and @tuomas777, this implementation will be rejected and a more clean implementation will be used.

PR #621, opened 2021-11-08
VEN-1372 | Allow creating winter storage lease without application

tuomas777:
## Description :sparkles:

Added `customerId` field to `CreateWinterStorageLeaseMutation` for providing the customer when a winter storage lease is created without an application. A place is mandatory in such leases.

### Closes :no_good_woman:
**[VEN-1372](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1372): Create marked WS lease without application** 

The implementation is based on the corresponding implementation of berth leases without an application.

### DIFF
--- a/leases/schema/mutations.py
+++ b/leases/schema/mutations.py
@@ -43,6 +43,45 @@ class AbstractLeaseInput:
     comment = graphene.String()
 
 
+def _lookup_application_and_customer(info, input, application_node_type):
+    from customers.schema import ProfileNode  # import here avoid circular import
+
+    if application_id := input.pop("application_id", None):
+        if "customer_id" in input:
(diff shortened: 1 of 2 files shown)


leases/schema/mutations.py
```diff
@@ -43,6 +43,45 @@ class AbstractLeaseInput:
     comment = graphene.String()
 
 
+def _lookup_application_and_customer(info, input, target_application_node):
```
a-rmz, line 46: Nitpick:

```suggestion
def _lookup_application_and_customer(info, input, application_node_type):
```

a-rmz (approved):
One small nitpick but otherwise looks good 😻

PR #622, opened 2021-11-09
VEN-1416 | Fix createMyBerthProfile mutation

tuomas777:
## Description :sparkles:

Fixed two issues related to `createMyBerthProfile` mutation:

* `myProfile` query to HKI profile which the mutation executes was broken.
* the mutation returned an error if the user already had a profile. Changed it to always return a valid `ProfileNode` and a boolean field `created` that tells if a new profile was created or not.

### Closes :no_good_woman:
**[VEN-1416](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1416): createMyBerthProfile mutation call fails with unclear reason**

### DIFF
--- a/customers/schema/mutations.py
+++ b/customers/schema/mutations.py
@@ -229,6 +229,7 @@ class CreateMyBerthProfileMutation(graphene.ClientIDMutation):
         )
 
     profile = graphene.Field(ProfileNode)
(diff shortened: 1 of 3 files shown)


haavikko (approved):
Looks good to me!

PR #624, opened 2021-11-17
VEN-1420 | Resolve annotated props on single node calls

a-rmz:
## Description :sparkles:
* Add an explicit resolver for Harbor/Area properties that are only annotated by the manager, so they have to be explicitly queried

### Closes :no_good_woman:
**[VEN-1420](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1420):** Querying harbours through berthApplication results in an error

```graphql
query Applications {
  berthApplications {
    edges {
      node {
        id
        harborChoices {
          harbor {
            id
            properties {
              name
              electricity
              gate
              lighting
              wasteCollection
              water
            }
          }
        }
      }
    }
  }
}
```
Should then include the values for the selected harbor.
```json
{
    "properties": {
        "name": "Airorannan venesatama",
        "electricity": false,
        "gate": false,
        "lighting": false,
        "wasteCollection": false,
        "water": false
    }
}
```

### DIFF
--- a/utils/dataloaders.py
+++ b/utils/dataloaders.py
@@ -1,0 +1,19 @@
+from collections import defaultdict
+from typing import Type
+
+from django.db.models import Model
+from promise import Promise
+from promise.dataloader import DataLoader
+
+
+def model_loader(model: Type[Model]):
+    class ModelLoader(DataLoader):
+        def batch_load_fn(self, object_ids):
+            objects = defaultdict(model)
+
+            for object in model.objects.filter(id__in=object_ids).iterator():
+                objects[object.id] = object
+
+            return Promise.resolve([objects.get(object_id) for object_id in object_ids])
+
+    return ModelLoader
--- a/applications/schema/types.py
+++ b/applications/schema/types.py
@@ -43,6 +43,9 @@ class HarborChoiceType(DjangoObjectType):
         model = HarborChoice
         exclude = ("id", "application")
 
+    def resolve_harbor(self, info, **kwargs):
+        return info.context.harbor_loader.load(self.harbor_id)
+
 
 class BerthSwitchReasonType(DjangoObjectType):
     title = graphene.String()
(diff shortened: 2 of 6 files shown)


tuomas777 (commented):
Hmm, wouldn't this cause a separate DB query for every field?

a-rmz:
That's a good point :derp: but hmmm not sure how else to resolve this fields then 🤔

tuomas777 (approved):
TBH I know almost nothing about those dataloaders, but this looks very good as far as I can tell 🙂 👍 ⛵

a-rmz:
Ohhhh they're a very useful tool! They reduced our query times from around 2 minutes down to 2 seconds and only a handful of queries. It "loads" all the queries on the same SQL query so instead of fetching for each Harbor individually, it adds all the ids to the loader and it gets all the results in a single go, and then it knows which call needs which harbor based on the ID.

It wouldn't be _critical_ for this case, but if we can make it better, why not 😄 

https://github.com/City-of-Helsinki/berth-reservations/pull/442
https://docs.graphene-python.org/en/latest/execution/dataloader/#dataloader
https://github.com/City-of-Helsinki/open-city-profile/pull/250

utils/dataloaders.py
```diff
@@ -0,0 +1,19 @@
...
+
+from django.db.models import Model
+from promise import Promise
+from promise.dataloader import DataLoader
+
+
+def model_loader(model: Type[Model]):
+    class ModelLoader(DataLoader):
+        def batch_load_fn(self, object_ids):
+            objects = defaultdict(model)
```
haavikko, line 12: Seems there's something I don't understand about DataLoaders...
What would happen if some object_ids are not found?
objects.get(object_id) would return a new instance of model, what's the use case for that? If graphene asks for objects that are not found, can we just raise exception?

haavikko (approved):
Looks good to me! Left a question about DataLoaders, but that's probably something I don't get about the loaders.

utils/dataloaders.py
a-rmz: If the object is not found, it would return `None`, since we're filtering the objects, not explicitly `get`-ing. 

`objects.get(object_id)` would return the value that was stored on this `objects` dict, so in the end, the method calling the dataloader (e.g. the `resolve_harbor`) would be the one which would have to deal with the `None`, same way as it would with a plain `Harbor.objects.get()`.

PR #626, opened 2021-11-18
VEN-1414: add TalpaProductAccounting mapping

haavikko:
## Description :sparkles:

@a-rmz please take a look especially at get_product_accounting_for_order, and the way STORAGE_ON_ICE orders are identified there

### Closes :no_good_woman:
**[VEN-1414](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1414):** 

### Automated tests :gear:️
payments/tests/test_payments_talpa_product_accounting.py


payments/models.py
```diff
@@ -296,6 +294,70 @@ def name(self):
...
+            )
+        if product_type is None:
+            raise TalpaProductAccountingNotFoundError(
+                _("Talpa product type for order could not be determined")
+            )
+        return self.get(region=region, product_type=product_type)
+
+
+class TalpaProductAccounting(UUIDModel):
+    objects = TalpaProductAccountingManager()
```
a-rmz, line 348: Nitpick: the `objects` (or any manager) should be the last field to be defined 😄

payments/models.py
```diff
@@ -296,6 +294,70 @@ def name(self):
...
+    region = models.CharField(
+        choices=AreaRegion.choices, verbose_name=_("area region"), max_length=32,
+    )
+    talpa_ecom_product_id = models.UUIDField()
+    talpa_ecom_accounting_id = models.UUIDField()
```
a-rmz, line 353: Nitpick: can you also add verbose names for this ones? 🙏

payments/enums.py
```diff
@@ -125,3 +125,8 @@ class PricingCategory(IntegerChoices):
     DINGHY = 1, _("Dinghy places")
     TRAILER = 2, _("Trailer places")
     VASIKKASAARI = 3, _("Vasikkasaari berths")
+
+
+class TalpaProductType(TextChoices):
+    BERTH = "west", _("West")
+    WINTER = "east", _("East")
```
a-rmz, line 132: This is probably a typo 😅 but you have `BERTH = "west"` and `WINTER = "east"`

a-rmz (changes requested):
There's the small mixup with `TalpaProductType`, it's a bit complicated to follow which one is which 😅 

Also, there are a lot of formatting changes. Did you update your `black/isort` versions? We bumped them a few weeks ago.

```
pip-sync requirements.txt requirements-dev.txt
```

payments/enums.py
haavikko: yes, fixed!

payments/models.py
haavikko: ok, fixed

payments/models.py
haavikko: ok, added

haavikko:
Fixed the formatting changes now. I had to re-do "pre-commit install", as the pre-commit hook was using a wrong black version from an older virtualenv.

payments/models.py
```diff
@@ -296,6 +298,74 @@ def name(self):
...
+        """
+
+        if order.order_type == OrderType.ADDITIONAL_PRODUCT_ORDER:
+            if (
+                order.order_lines.count() == 1
+                and order.order_lines.all().first().product.service
+                == ProductServiceType.STORAGE_ON_ICE
+            ):
+                product_type = TalpaProductType.WINTER
+                region = order.lease.berth.pier.harbor.region
```
a-rmz, line 316: Storage on ice woks only for berth leases, right? Should we somehow catch the case where the order doesn't have a lease?

payments/models.py
```diff
@@ -296,6 +298,74 @@ def name(self):
...
+                product_type = TalpaProductType.BERTH
+            else:
+                raise TalpaProductAccountingNotFoundError(
+                    _("TalpaProductAccounting parameters for order")
+                )
+
+        if region is None:
+            raise TalpaProductAccountingNotFoundError(
+                _("Region for order could not be determined")
+            )
```
a-rmz, line 343: As a small improvement, instead of explicitly looking the region for each type, we could use the `resolve_area` util and just get it on a single point, like

```suggestion
        region = getattr(resolve_area(order), "region", None)
        if region is None:
            raise TalpaProductAccountingNotFoundError(
                _("Region for order could not be determined")
            )
```

payments/models.py
```diff
@@ -296,6 +298,74 @@ def name(self):
...
+            elif order.lease_order_type in (
+                LeaseOrderType.BERTH_SWITCH_ORDER,
+                LeaseOrderType.RENEW_BERTH_ORDER,
+                LeaseOrderType.NEW_BERTH_ORDER,
+            ):
+                region = order.lease.berth.pier.harbor.region
+                product_type = TalpaProductType.BERTH
+            else:
+                raise TalpaProductAccountingNotFoundError(
+                    _("TalpaProductAccounting parameters for order")
```
a-rmz, line 337: Do you mean _missing_ parameters for order?

payments/models.py
```diff
@@ -296,6 +298,74 @@ def name(self):
...
+                _("Region for order could not be determined")
+            )
+        if product_type is None:
+            raise TalpaProductAccountingNotFoundError(
+                _("Talpa product type for order could not be determined")
+            )
+        return self.get(region=region, product_type=product_type)
+
+
+class TalpaProductAccounting(UUIDModel):
```
a-rmz, line 351: We should add a unique constraint for `region` + `product_type` to ensure we don't have multiple results when fetching the accounting info for an order.

payments/tests/factories.py
```diff
@@ -122,6 +125,16 @@ class Meta:
         model = AdditionalProduct
 
 
+class TalpaProductAccountingFactory(factory.django.DjangoModelFactory):
+    class Meta:
+        model = TalpaProductAccounting
```
a-rmz, line 130: To avoid having duplicates on the testing db:
```suggestion
    class Meta:
        model = TalpaProductAccounting
        django_get_or_create = ("region", "product_type",)
```

payments/tests/test_payments_talpa_product_accounting.py
```diff
@@ -0,0 +1,88 @@
...
+from resources.enums import AreaRegion
+
+
+@pytest.mark.parametrize(
+    "order_with_products",
+    ["berth_order", "winter_storage_order"],
+    indirect=True,
+)
+@pytest.mark.parametrize("region", [AreaRegion.WEST, AreaRegion.EAST])
+def test_berth_product_order(
```
tuomas777, line 21: To be more consistent with other tests' naming, I guess the name could include also "winter storage"?

payments/models.py
```diff
@@ -296,6 +299,80 @@ def name(self):
     )
 
 
+class TalpaProductAccountingManager(models.Manager):
```
tuomas777, line 302: FWIW personally I use custom `Queryset`s instead of `Manager`s ~always because they allow method chaining without any actual downsides (that I know of at least :D)

a-rmz: Welp, we've been using custom managers throughout the project, so I guess it won't harm to keep this one also for consistency 😅

payments/tests/test_payments_talpa_product_accounting.py
a-rmz: Or just plain `product` order :derp: but nice idea

a-rmz (approved):

PR #627, opened 2021-11-19
VEN-1406 | Connect applications to existing boats

tuomas777:
## Description :sparkles:

From now on, all `Application`s are connected to a `Boat` by a foreign key relation. Also, all `boat_XXX` fields are removed from `Application`s, they exist only on `Boat`s now. There is a data migration that will take care of creating `Boat`s for all already existing `Application`s. If an `Application` has a customer, the related `Boat` must have the same customer as the owner.

For the API side, there is new field `boatId` in application create and update mutation inputs, which can be used to set a `Boat` for an `Application`. The old `boatXXX` fields are also still fully functional in the mutations, and using those will create/update the related `Boat` instance. The API changes **should** be backwards compatible, but that hasn't been properly tested.

The management command `create_boat_for_existing_applications` is removed because there is no longer need for it. 

### Closes :no_good_woman:
**[VEN-1406](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1406): Connect an application to an existing boat** 

Needed. A lot.

* Changes in `boat_XXX` fields are not saved in `ApplicationChange` entries anymore, because there didn't seem to be any trivial ways to do that with the new model structure, and this PR is quite big already. If that is needed, it should be implemented separately later.

* It is no longer possible to change the customer of an application if it already has one. The reason is that the application's customer would no longer match the related boat's owner, and there is no trivial solution how to handle that.


applications/schema/mutations.py
```diff
@@ -42,6 +43,108 @@
...
 
+def _validate_boat_fields_common(application_data):
+    if "boat_id" in application_data:
+        if any(
+            field.startswith("boat_") and field != "boat_id"
+            for field in application_data
+        ):
+            raise VenepaikkaGraphQLError(
+                _(
+                    'Cannot use both "boatId" and other "boatXXX" field(s) at the same time."'
```
a-rmz, line 54: Small nitpick: for consistency with the rest of our error messages, could you rename all this `boatXXX` to snake case `boat_xxx`? 🙏

applications/schema/mutations.py
```diff
@@ -42,6 +43,108 @@
...
+        boat = application.boat
+        boat_data = _get_boat_data(application_data)
+
+    if application_data.get("customer") and not boat.owner:
+        # an anonymous application is begin assigned to a customer,
+        # assign the boat as well
+        boat_data["owner"] = application_data["customer"]
+
+    if boat_data:
+        update_object(application.boat, boat_data)
```
a-rmz, line 113: Shouldn't this be:
```suggestion
        update_object(boat, boat_data)
```

a-rmz (approved):
Only a few small comments but looks good! Nice job!!! 🤩

applications/schema/mutations.py
tuomas777: Indeed it should. Well spotted!

utils/relay.py
```diff
@@ -149,7 +149,12 @@ def return_queryset_if_user_has_permissions(
     if is_customer(user):
         if customer_queryset is not None:
             return customer_queryset
-        return queryset.filter(customer__user=user)
+        if hasattr(queryset.model, "customer"):
+            return queryset.filter(customer__user=user)
+        elif hasattr(queryset.model, "owner"):
+            return queryset.filter(owner__user=user)
+        else:
+            raise Exception(f"Unsupported model {queryset.model}.")
```
a-rmz, line 157: Maybe a `ValueError` or something a biiiit more specific? But up to you, lgtm otherwise 👍

tuomas777: I guess it should be `TypeError` then. Changing to that.

PR #629, opened 2021-11-23
VEN-1414 | Update TalpaProductAccounting fields

a-rmz:
## Description :sparkles:
* Remove the accounting id field that won't be used
* Make the ID field editable to allow adding the mappings after they've been created in Talpa


haavikko (approved):
thumbs up!

PR #635, opened 2021-12-02
Fix migration that populates boat for applications

tuomas777:
## Description :sparkles:

Now we use an existing boat only if the application has a related `CustomerProfile` instance and a `boat_registration_number`, and there is exactly one `Boat` that matches those.

Also, applications are now handled in creation order.


haavikko (approved):
Looks good to me!

PR #636, opened 2021-12-09
Talpa ecom fixes

tuomas777:


payments/providers/talpa_ecom.py
```diff
@@ -428,7 +428,7 @@ def get_payment_details(self, order: Order) -> TalpaEComPaymentDetails:
         headers = {"user": get_customer_hash(order.customer)}
 
         r = requests.get(
-            f"{self.url_payment_experience_api}/{order.talpa_ecom_id}",
+            f"{self.url_payment_experience_api.rstrip('/')}/{order.talpa_ecom_id}",
```
haavikko, line 431: Could also use something like urllib.parse.urljoin but for a case this simple not necessary

haavikko (approved):
Looks good to me!

payments/providers/talpa_ecom.py
tuomas777: I actually initially used `urljoin()` but then realised it has a quite amount of [weirdness](https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin) which makes it not so great in this use case 🙂

PR #642, opened 2022-02-01
Fix docker environment related conflicts

charn:
## Description :sparkles:

With the current default settings it was impossible to run berth-reservations along with tunnistamo and open-city-profile.


nikomakela (approved):
👍

PR #647, opened 2022-02-11
VEN-1434 VEN-1444 | Admin can export a customer list to an excel

charn:
## Description :sparkles:

This PR adds initial excel export for customers, but also some general classes for adding more exporters later.

## Issues :bug:
**[VEN-1444](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1444):** BE: API implements different Excel export views in their own endpoints
**[VEN-1434](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1434):** As an admin I can export a customer list to excel

Fetching user information from Helsinki profile backend using `ProfileService` coming later.

### DIFF
--- a/exports/tests/test_customer_export.py
+++ b/exports/tests/test_customer_export.py
@@ -1,0 +1,93 @@
+import pytest
+import xlrd
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
+from django.urls import reverse
+from rest_framework import status
+
+from berth_reservations.tests.factories import CustomerProfileFactory
+from customers.enums import InvoicingType
+from customers.models import CustomerProfile
+from customers.schema import ProfileNode
+from exports.xlsx_writer import CustomerXlsx
+from utils.relay import to_global_id
+
+
+def to_global_profile_node_ids(ids):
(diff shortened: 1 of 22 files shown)


exports/tests/test_customer_export.py
```diff
@@ -0,0 +1,93 @@
...
+    CustomerProfileFactory.create_batch(2, user=None)
+    ids = CustomerProfile.objects.all().values_list("id", flat=True)
+    global_ids = to_global_profile_node_ids(ids)
+
+    with django_assert_max_num_queries(2):
+        response = superuser_api_client.post(
+            reverse("customer_xlsx"), data={"ids": global_ids}
+        )
+
+    assert response.status_code == status.HTTP_200_OK
```
nikomakela, line 51: This test is included in the test right under it.

nikomakela (approved):
Looking too damn good. Ship it so we can easier test it! 🚀

exports/tests/test_customer_export.py
charn: It is, though here it's mostly making sure that we have gotten a valid response and thus `django_assert_max_num_queries` is testing a valid situation.

PR #648, opened 2022-02-14
Using Case-When to order the customers by profile query set

nikomakela:
VEN-1469. Fixing the issue with in_bulk implementation by changing it to case-when-preserved order-by -query


charn (approved):
LGTM! :+1: :shipit:

PR #650, opened 2022-02-15
Exclude Helsinki-customer filter

nikomakela:
VEN-1469. The open city profile's profiles-query does not support the excluding of cities - it supports only filtering them - so the excluding of the city of Helsinki postal districts needs to be done where the id's are resolved.


charn (approved):
LGTM! :+1: :shipit:

nikomakela:
Closing this -  #662 should be used instead

PR #653, opened 2022-02-16
VEN-1465 | Customer excel export fetches information from Helsinki profile

charn:
This PR adds Helsinki profile related information to the customer excel export.

Customer excel exporter view now requires a new `profile_token` (API key for the Helsinki profile GraphQL API) argument, which is used to fetch customer related information from Helsinki profile into the exported excel.

Additionally `xlrd` is changed to `openpyxl` since new versions of xlrd are only intended for reading old `.xls` files.

### Closes :no_good_woman:
**[VEN-1465](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1465):** BE: backend resolves the names of the customer profile

### DIFF
--- a/applications/tests/test_applications_utils.py
+++ b/applications/tests/test_applications_utils.py
@@ -1,7 +1,9 @@
+import io
+
 import pytest
-import xlrd
 from django.conf import settings
 from freezegun import freeze_time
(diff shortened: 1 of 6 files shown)


nikomakela (approved):
Good!

PR #654, opened 2022-02-16
Recursively fetch all the ids from profile

nikomakela:
VEN-1669


nikomakela:
@charn @Sartonon you can merge this to hopefully the fix current implementation while I'm having my holiday

charn (approved):
LGTM! :+1:

PR #658, opened 2022-02-21
Add missing curly bracket to query

charn:


nikomakela (approved):

nikomakela:
Works in my local, with a new curly and without :S But yeah, seems like there should be one.

PR #659, opened 2022-02-21
Find profile filters first by berth data and then by profile data

nikomakela:
VEN-1469. Filter first by berth filters, then by profile filters and then return the berth query in the right order.

### DIFF
--- a/customers/schema/queries.py
+++ b/customers/schema/queries.py
@@ -1,3 +1,6 @@
+from typing import List
+from uuid import UUID
+
 import graphene
 import graphene_django_optimizer as gql_optimizer
 from django.db.models import Case, Count, Q, When
@@ -126,7 +129,9 @@ def _general_filters(params, qs):
     return qs
 
 
-def _get_ids_from_profile_service(kwargs, profile_token):
+def _get_ids_from_profile_service(
+    kwargs: dict, profile_token: str, helsinki_profile_query_ids: List[UUID]
+):
     from customers.services import ProfileService
     from customers.services.profile import BATCH_SIZE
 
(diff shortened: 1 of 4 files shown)


customers/services/profile.py
```diff
@@ -367,7 +374,7 @@ def create_profile(
         response = self.query(query=mutation, variables=variables)
         global_id = response.get("createProfile", {}).get("profile", {}).get("id")
         profile_id = UUID(from_global_id(global_id))
-        profile = CustomerProfile.objects.create(id=profile_id)
+        profile = CustomerProfile.objects.get(id=profile_id)
```
charn, line 377: What's this change?

nikomakela: Good catch! It's an accident! I used the part of the script to import local data

customers/schema/queries.py
```diff
@@ -202,56 +209,62 @@ class Query:
...
-        marked_winter_storage_areas = kwargs.pop("marked_winter_storage_areas", [])
-        marked_winter_storage_places = kwargs.pop("marked_winter_storage_places", [])
-
-        unmarked_winter_storage_lease_customer = kwargs.pop(
-            "unmarked_winter_storage_lease_customer", False
-        )
-        unmarked_winter_storage_areas = kwargs.pop("unmarked_winter_storage_areas", [])
+        harbors = [h for h in kwargs.pop("harbors", []) if h is not None]
+        piers = [p for p in kwargs.pop("piers", []) if p is not None]
+        berths = [b for b in kwargs.pop("berths", []) if b is not None]
```
charn, line 214: Added a comment related to this on the other PR https://github.com/City-of-Helsinki/berth-reservations/pull/660#pullrequestreview-889545260

charn (approved):
LGTM! :+1: :shipit: That one remaining comment I had is not really a show stopper.

customers/schema/queries.py
nikomakela: Trivial / minor in my opinion. It would need some changes to UI also. Lets do it later if decided

PR #660, opened 2022-02-21
Removed the filter trigger parameters from berthProfiles query

nikomakela:
VEN-1469 VEN-1477.

### DIFF
--- a/customers/schema/queries.py
+++ b/customers/schema/queries.py
@@ -168,14 +168,11 @@ class Query:
         boat_types=graphene.List(graphene.ID),
         boat_registration_number=graphene.String(),
         lease_count=graphene.Boolean(),
-        berth_lease_customer=graphene.Boolean(),
         harbors=graphene.List(graphene.String),
         piers=graphene.List(graphene.String),
         berths=graphene.List(graphene.String),
-        marked_winter_storage_lease_customer=graphene.Boolean(),
         marked_winter_storage_areas=graphene.List(graphene.String),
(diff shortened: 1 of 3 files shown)


Jokinen (approved):

customers/schema/queries.py
```diff
@@ -211,21 +209,18 @@ class Query:
...
-        marked_winter_storage_areas = kwargs.pop("marked_winter_storage_areas", [])
-        marked_winter_storage_places = kwargs.pop("marked_winter_storage_places", [])
-
-        unmarked_winter_storage_lease_customer = kwargs.pop(
-            "unmarked_winter_storage_lease_customer", False
-        )
-        unmarked_winter_storage_areas = kwargs.pop("unmarked_winter_storage_areas", [])
+        harbors = [h for h in kwargs.pop("harbors", []) if h is not None]
+        piers = [p for p in kwargs.pop("piers", []) if p is not None]
+        berths = [b for b in kwargs.pop("berths", []) if b is not None]
```
charn, line 214: I guess one way to prevent this would be to define the argument like `berths=graphene.List(graphene.NonNull(graphene.String))`. Though, would that break something. :thinking:

nikomakela: Could be possible like that, but it would be useless without UI changes. UI should not send [None] either, but a None instead. Harbors, piers and berths would be checked the same way anyway. Quite trivial change in my opinion. The thing is that do we want the filters to be always included in the request.

PR #661, opened 2022-02-21
Fix to berthProfiles berth leases filter

nikomakela:
VEN-1469. Instead of summing berth leases and winter storages together, check whether or not there are multiple of one of them.

"Enemmän kuin yksi vene- tai talvisäilytyspaikka"


nikomakela:
Asked about this from designer and this pull request is a valid one and the change should be done.

Jokinen (approved):

PR #662, opened 2022-02-21
Refactor exclude Helsinki customer filter using postal codes

nikomakela:
VEN-1469. The open city profile's profiles-query does not support the excluding of cities - it supports only filtering them - so the excluding of the city of Helsinki postal districts needs to be done where the id's are resolved.

Alternative for #650 - using postal code instead

### DIFF
--- a/customers/schema/queries.py
+++ b/customers/schema/queries.py
@@ -21,7 +21,14 @@ from ..models import CustomerProfile
 from ..utils import from_global_ids
 from .types import CustomerGroupEnum, InvoicingTypeEnum, ProfileFilterSet, ProfileNode
 
-HELSINKI_PROFILES_FILTERS = ["first_name", "last_name", "email", "address", "sort_by"]
+HELSINKI_PROFILES_FILTERS = [
+    "first_name",
+    "last_name",
+    "email",
+    "address",
+    "non_helsinki_citizen",
+    "sort_by",
+]
 
 
 def _filter_winter_storage_leases(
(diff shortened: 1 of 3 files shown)


customers/schema/queries.py
```diff
@@ -126,6 +135,12 @@ def _general_filters(params, qs):
     return qs
 
 
+def _exclude_helsinki_citizen(
+    users: List[HelsinkiProfileUser],
+) -> List[HelsinkiProfileUser]:
+    return [user for user in users if user.postal_code.startswith("00")]
```
nikomakela, line 141: it should be "not startswith"

Jokinen:
Added a WIP PR that contains the control for the UI for this new filter in the `berthProfiles` query:

https://github.com/City-of-Helsinki/berth-reservations-admin/pull/472

Types may need to be updated after the backend change has been merged.

customers/tests/test_customers_queries.py
```diff
@@ -1431,3 +1431,48 @@ def test_filter_by_hki_profile_order_by(mock_find_profile, names, superuser_api_
             edge["node"] for edge in executed["data"]["berthProfiles"]["edges"]
         ]
     ]
+
+
+@patch("customers.services.profile.ProfileService.find_profile")
+def test_filter_non_helsinki_citizen(mock_find_profile, superuser_api_client):
+    profile_1, profile_2, profile_3 = CustomerProfileFactory.create_batch(3)
+    mock_find_profile.return_value = [
+        HelsinkiProfileUser(
```
charn, line 1440: I guess, we need to add the postal_codes to these test fixtures.

PR #665, opened 2022-03-29
Remove ids filter from berth profiles query

nikomakela:
VEN-1481. The ids filter was denied by Open City Profile API, so it can be removed from Berth API. When resolving the berth profiles, first fetch all the profiles with given filters, then query from berth db. 

The Open city profile now responds with` 413 Request Entity Too Large` when there are ~3500 ids. This differs from what we have investigated when we implemented the current version of customer query.

### DIFF
--- a/customers/schema/queries.py
+++ b/customers/schema/queries.py
@@ -1,6 +1,3 @@
-from typing import List
-from uuid import UUID
-
 import graphene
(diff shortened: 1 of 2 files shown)


charn (approved):
Tested locally and this change does improve the customers admin view query quite a bit. :100: LGTM! :+1: :shipit:

PR #671, opened 2022-04-25
Consistent templates and cancel feature to new berth offer and berth invoice templates

nikomakela:
VEN-1430 VEN-1460. Consistent templates and add a cancellation link to the berth renewal handling template.

(An improvement for #670).

Tried to create a base template which could have been extended by the invoice templates, but the sandboxed Jinja does not support that. Tried with include -block also without luck. Since the lead text and the heading differs in templates, I needed to duplicate the long HTML under them.


scripts/load_notification_templates.py
```diff
@@ -37,12 +37,12 @@ def load_email_templates():
             "plain": None,
         },
         PaymentNotificationType.NEW_BERTH_ORDER_APPROVED: {
-            "html": "berth_offer_new_{lang}.html",
-            "plain": None,
+            "html": "berth_offer_new_{lang}.html",  # Extending base_berth_offer_invoice base-html-template.
```
charn, line 40: I guess the base template does not yet exist?

nikomakela: wops.. Actually, the base tempalte did not work, so the comment is now outdated. Instead of base template, the code is now duplicated. Extending a base or including a snippet did not work in this case when we are using a sandboxed jinja environment and when the templates are handled like they now are. It would need a huge refactor.

charn (approved):
LGTM! :+1: :shipit:

PR #672, opened 2022-04-26
Add previous season lease resolver to the berths node

nikomakela:
VEN-1375 VEN-1500.

### DIFF
--- a/leases/utils.py
+++ b/leases/utils.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from datetime import date
-from typing import TYPE_CHECKING, Union
+from typing import Tuple, TYPE_CHECKING, Union
 
 from babel.dates import format_date
 from dateutil.relativedelta import relativedelta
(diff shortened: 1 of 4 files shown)


melniiv (approved):

leases/utils.py
```diff
@@ -66,6 +66,16 @@ def calculate_season_end_date(lease_end: date = None) -> date:
     return default
 
 
+def calculate_prev_season_start_date() -> date:
+    today = date.today()
+    return date(day=10, month=6, year=today.year - 1)
```
charn, line 71: I might put these constant into settings (e.g. `BERTH_SEASON_START_MONTH` and `BERTH_SEASON_END_MONTH`). But maybe that's not totally necessary.

charn (approved):
LGTM! :+1: :shipit:

leases/utils.py
nikomakela: I thought that too, but I think it should need a greater refactor by then. Also, I started to think what we could really achieve with them and what would happen to the history data if the constant would be changed. Anyway, I created a new ticket of this: VEN-1507 (https://helsinkisolutionoffice.atlassian.net/browse/VEN-1507).

PR #674, opened 2022-04-27
PT-1436 PT-533 | Patch graphene.Decimal

charn:
## Description :sparkles:

This PR patches `graphene.Decimal.parse_value` with improved float handling. Also we replace all explicit Float types in mutation inputs.

## Issues :bug:
**[VEN-1436](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1436):** BE: Unmarked winter storage notification has floats instead of decimals
**[VEN-533](https://helsinkisolutionoffice.atlassian.net/browse/VEN-533):** Backend: passing floats for DecimalFields does not always pass validation

### DIFF
--- a/applications/schema/inputs.py
+++ b/applications/schema/inputs.py
@@ -40,8 +40,8 @@ class BaseApplicationInput(graphene.InputObjectType):
 
 
 class BerthApplicationInput(BaseApplicationInput):
-    boat_draught = graphene.Float()
-    boat_weight = graphene.Float()
+    boat_draught = graphene.Decimal()
+    boat_weight = graphene.Decimal()
     accessibility_required = graphene.Boolean()
     boat_propulsion = graphene.String()
     boat_hull_material = graphene.String()
--- a/berth_reservations/settings.py
+++ b/berth_reservations/settings.py
@@ -6,6 +6,8 @@ import sentry_sdk
 from django.utils.translation import gettext_lazy as _
 from sentry_sdk.integrations.django import DjangoIntegration
 
+from utils.patch import monkeypatch_graphene_decimal
+
 checkout_dir = environ.Path(__file__) - 2
 assert os.path.exists(checkout_dir("manage.py"))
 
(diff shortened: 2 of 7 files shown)


nikomakela (approved):
Nice catch! It's a bit sad though that the `Decimals` needs to be found and replaced with the new `DecimalScalar`. I tested that after the type change, the UIs' (graphql) type generator seems to change the number types to any-types: 
image

Another approach could be using the patches or meta-class changes or whatever they are called in Python. In Groovy (language) any methods could be added or overrode with ExpandoMetaClasses https://blog.mrhaki.com/2009/12/groovy-goodness-adding-or-overriding.html, which means that you could permanently overrode the methods of the 3rd party classes. I wonder how this could be achieved in Django.

applications/schema/inputs.py
```diff
@@ -40,8 +40,8 @@ class BaseApplicationInput(graphene.InputObjectType):
 
 
 class BerthApplicationInput(BaseApplicationInput):
-    boat_draught = graphene.Float()
-    boat_weight = graphene.Float()
+    boat_draught = graphene.Decimal()
```
nikomakela, line 43: FYI: Tested this with berth-reservation-admin (UI) `yarn type:generator`. The numbers are changed to now with the any-type. Floats were typed as numbers.

berth_reservations/settings.py
```diff
@@ -342,3 +344,6 @@
     exec(code, globals(), locals())
 
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
+
+# Monkey-patching
```
nikomakela, line 348: Could be highlighted even more! And should be mentioned clearly in the change logs.

utils/patch.py
```diff
@@ -0,0 +1,20 @@
+from decimal import Decimal as _Decimal
+
+import graphene
+from graphene.types import Decimal
+
+
+def monkeypatch_graphene_decimal():
+    if graphene.__version__ != "2.1.9":
+        raise Exception(
+            "Graphene version changed. Check if graphene.Decimal still needs patching."
```
nikomakela, line 10: Nice!

nikomakela (approved):

PR #677, opened 2022-04-29
Reject payment notification sending when order has paper invoice customer

nikomakela:
VEN-1175. Raise the InvoicingRejectedForPaperInvoiceCustomersError when payment notification is being sent to a paper customer. Catch the exception on order approving so the changes to the order are still persisted. Fixed the tests for paper invoice customer testing.

When the customer has selected paper invoicing, all the order handling should be done as when the invoicing type would be something else (only digital invoicing currently available), but the payment notification should not be sent, because a paper document should be used instead.

Investigation of the payment notification process is done in the Jira ticket comments: https://helsinkisolutionoffice.atlassian.net/browse/VEN-1175?focusedCommentId=61250


charn (approved):
LGTM! :+1: :shipit:

PR #680, opened 2022-05-04
Added a comment-field to the winter storage place

nikomakela:
VEN-1513 VEN-1514 VEN-1516: Added the same kind of comment field to the winter storage place as is in use with berth (input).


melniiv (approved):

charn (approved):
LGTM! :+1: :shipit:

PR #681, opened 2022-05-05
VEN-1511 | Add management command for fetching customer email

charn:
This PR adds a management command for fetching emails for a given list of profile ids from Helsinki profile. Also this fixes a bug where `ProfileService.get_all_profiles` ended up into an infinite query loop when fetching profiles without the `profile_ids` argument.

## Issues :bug:
- **[VEN-1511](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1511)** 
## Usage
    ./manage.py fetch_customer_emails --profile-token helsinki-profile-jwt-token-here --input-file customer_ids.txt --output-file customer_emails.csv

### DIFF
--- a/customers/management/commands/fetch_customer_emails.py
+++ b/customers/management/commands/fetch_customer_emails.py
@@ -1,0 +1,55 @@
+import csv
+import logging
+
+from django.core.management import BaseCommand
+
+from customers.services import ProfileService
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


nikomakela (approved):
Nice!

nikomakela:
a minor / trivial improvement could be that an error could be raised when the api-connection does not work. I had a wrong PROFILE_API_URL set to my env, and I still get the "Exported Processed 8 profiles" -message, but the file was empty. Because this script will be ran always from the pod that already has the right env-variables, I think it's very ok to leave it like this!

PR #682, opened 2022-05-06
Added an possibility to postpone the invoicing of non-digital customers

nikomakela:
VEN-1170. Added an option to exclude paper invoices from unpaid orders handling. Added a result set  option to the expire too old unpaid orders command.


utils/base_expiration_command.py
```diff
@@ -27,6 +27,11 @@ def add_arguments(self, parser):
             action="store_true",
             help=f"Run the command even if {self.feature_flag_name} is not set",
         )
+        parser.add_argument(
```
charn, line 30: So this new argument will not yet be used anywhere e.g. in the cronjobs?

charn (approved):
LGTM! :+1: :shipit:

utils/base_expiration_command.py
nikomakela: nope. Not used yet, but can be used if needed. I just wanted to add it there already, but didn't want to play with the cron jobs, so set the default to be "excluded".

PR #688, opened 2022-05-27
VEN-1504 | Expire old offers after 3 days

charn:
This PR changes the following settings:
- `EXPIRE_WAITING_ORDERS_OLDER_THAN_DAYS` and
- `EXPIRE_WAITING_OFFERS_OLDER_THAN_DAYS` 

These are used by the following management commands which get run as a cron jobs once per day:
- `expire_too_old_unpaid_orders` (orders) and
- `expire_too_old_offers` (berth switch offers)

## Ticket :bug:
**[VEN-1504](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1504):** BE: Expire old offer 3 days after due date


nikomakela (approved):

nikomakela:
I also understood that its the both attributes, orders and offers, but the ticket mentioned only offers. Should it be rechecked?

charn:
> I also understood that its the both attributes, orders and offers, but the ticket mentioned only offers. Should it be rechecked?

I'll try to confirm this before merging.

PR #699, opened 2022-11-25
VEN-1538 Add sticker_season filter for customer query

kristianjokela:
## Description :sparkles:
This adds a season filter for customers queries, which is needed in order to filter customers by sticker number AND season the sticker was issued on.

Sticker season is just computed property, and relies on months to decide the assigned season. But since this is just for querying the season, the users are supposed to know the correct value - the filter simply uses `lease.start_date` and `lease.end_date`, ie. there's no magic conversion before/after 15.9 etc. happening.

### Closes :no_good_woman:
**[VEN-1538](https://helsinkisolutionoffice.atlassian.net/browse/VEN-1538):** 

### Related :handshake:
https://helsinkisolutionoffice.atlassian.net/browse/VEN-983


nikomakela (approved):
The changes looks good to me! The new variation could be tested with a unit test also.

kristianjokela:
> The changes looks good to me! The new variation could be tested with a unit test also.

Good idea, I added tests!

PR #783, opened 2025-01-31
Upgrade Docker image, Python, Django and dependencies

nikomakela:
VEN-1584.

NOTE: pipeline changes are in https://dev.azure.com/City-of-Helsinki/venepaikat/_git/venepaikat-pipelines/pullrequest/10680.

- Upgrade Docker-image, Python, Django and environment requirements.
- Migrate deprecated and removed code on Django upgrade.
- Bring backward compatibility for Django's Choices enum.
- Fix tests after upgrade.


terovirtanen:
**venepaikat-api branch is deployed to platta: https://venepaikat-api-pr783.dev.hel.ninja** :rocket::rocket::rocket:

karisal-anders (approved):
"psycopg vs psycopg-binary" @ https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary suggest using the source distribution for production use. So not switching to psycopg[binary] would be probably better, but if psycopg[binary] works also in production, then I don't see this as a blocker.

nikomakela:
> "psycopg vs psycopg-binary" @ https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary suggest using the source distribution for production use. So not switching to psycopg[binary] would be probably better, but if psycopg[binary] works also in production, then I don't see this as a blocker.
> 
> Please take a look at the failing stages/tests in CI/CD pipeline.

Yea, you are right. Only your documentation was for wrong version, but comment is still valid. I just didn't know how to use it, but now I do.

Version 3 docs: https://www.psycopg.org/psycopg3/docs/basic/install.html.

terovirtanen:
**venepaikat-api branch is deployed to platta: https://venepaikat-api-pr783.dev.hel.ninja** :rocket::rocket::rocket:

nikomakela:
The tests were about 10 seconds too slow to execute so they were paused. I won't fix that issue now. The logs shows that all passed, but the cancel process was launched already, so it is reported as cancelled.

PR #787, opened 2025-02-20
fix: increase uwsgi buffer-size

nikomakela:
VEN-1584.

Keycloak will include so many ad-groups in the JWT that the request buffers-size needs to be increased. Set a consistent value copied from Kukkuu.


terovirtanen:
**venepaikat-api branch is deployed to platta: https://venepaikat-api-pr787.dev.hel.ninja** :rocket::rocket::rocket:

